PR #13, opened 2015-01-26
fix pom and clean deprecated class  and add bat file to windows

Fameing:


andsel:
Hi Igor,
thank you for your patch. To accept the pull request (beeing Moquette in the middle of transition into Eclipse Foundation IoT) I need you to be an Eclipse Committer, else I could insert the pull request "spirit" into the the master by myself.

andsel:
I've integrated by hand part of the patch, keep out the downgrade of MapDB

PR #18, opened 2015-03-09
use moquette path property where possible

maggu2810:
We should always prefer the moquette.path system property before we fill
the home directory of the user with files.

The formatter removes some trailing whitespaces.

### DIFF
--- a/broker/src/main/java/org/eclipse/moquette/server/Server.java
+++ b/broker/src/main/java/org/eclipse/moquette/server/Server.java
@@ -19,29 +19,40 @@ import java.io.File;
 import java.io.IOException;
 import java.text.ParseException;
 import java.util.Properties;
-import org.eclipse.moquette.spi.impl.SimpleMessaging;
+
 import org.eclipse.moquette.server.netty.NettyAcceptor;
+import org.eclipse.moquette.spi.impl.SimpleMessaging;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
 /**
- * Launch a  configured version of the server.
+ * Launch a configured version of the server.
+ *
  * @author andrea
  */
 public class Server {
-    
+
     private static final Logger LOG = LoggerFactory.getLogger(Server.class);
-    
-    public static final String STORAGE_FILE_PATH = System.getProperty("user.home") +
-            File.separator + "moquette_store.mapdb";
+
+    public static final String STORAGE_FILE_PATH = getMoquettePath() + File.separator + "moquette_store.mapdb";
 
     private ServerAcceptor m_acceptor;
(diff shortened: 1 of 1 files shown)


maggu2810:
Sure, this would be possible, but not in scope of that PR.
This PR should just respect mosquitto.path for the storage similar to the config, so the current bundle implementation works with it.

maggu2810:
I think we (you?) should create an additional PR, so that all configuration values are part of the config properties. The location of the database should be a configuration parameter.
The static field is currently used by tests etc. So we could use that as a fallback if no configuration is set. In that case (configuration property is not set manually) I would prefer to keep this commits, so we could set mosquitto.path also for the test run and the user home directory is not filled with the storage files.

andsel:
I think that the configuration (except for logging) should be kept in one place. Is PR's scope to permit the user to customize the data path? If it's yes, in my opinion I think it should be in moquette.conf, what to you think?

maggu2810:
Yes, the intention was to change the data path.
Using moquette.conf (and so on the config properties) seems to be good.

But I think we should also change the default behavior if no configuration is available.
Running tests should never fill my user home directory, they should use a in-memory database (possibl using mapdb) or create a temporary directory to store all that stuff.

I used mosquitto.path in this PR because it was the simplest change.
But if I think about your comments, a config property would be the best (with a fallback to a "usefull" default behaviour, that keep my system clean).

Are there any requirements to keep a full backward compatibility? So, should we check if there is an existing database in user.home and move them to the "new" location?

swarwick:
I changed my version of the code to allow for

m_properties.put("persistent_store", System.getProperty("user.home") + File.separator + "moquette_store.mapdb");

In the ConfigurationParser.java file (and of course other changes where needed for this change).  This allows me to specify the db store path and if its not specified it defaults to what it was before.  I am not able to commit this change for review but if you want me to just send a diff or something for you guys to add let me know.

andsel:
I've merged the changes that @swarwick  created on his fork. Now the config file could use also "persistent_store" to define the path.

maggu2810:
Could not read the sources ATM.
Your OSGi bundle is using the startServer() method.
So, what is the default location for the database, is there still the home directory used or are some properties preferred that I could set starting my OSGi container?
What is the location of the database the unit test are using?

andsel:
In unit tests it has to be fixed, probably I'll set to string empty to force a memory DB (NB the MapDBPersistentStore default to memory if can't find the storage path).
For OSGi I've to fix completely the OSGi part due to bad visibility of classes that has to be used by MapDB, so the OSGi part for version 0.7 SNAPSHOT is to consider under development

maggu2810:
Thank you for your feedback.

Issue #20, opened 2015-03-18
config settings: allow system properties


maggu2810:
Do you think it would be ok to use system properties for the values of the properties in the configuration file?
We could extend the parse method of the ConfigurationParser to replace system properties in the value (using: https://commons.apache.org/proper/commons-lang/javadocs/api-2.6/org/apache/commons/lang/text/StrSubstitutor.html).

If someone would configure the the moquette database path in the configuration file and does not want to use absolute path he could use e.g. the value "${moquette.path}/storage/moquette_store.mapdb"

maggu2810:
I would create a pull request if it is okay for you and there are no requirements not using commons-lang of apache.

andsel:
Hi @maggu2810,
personally I don't like to import a library just to do simple string substitution, the library is also old. As stated in the pull request #18 I don't like to have two ways to configure the same thing, having the system property and the config file as options to configure a feature of the broker could lead to troubles,"I've set the value but the broker take another".

maggu2810:
> two ways to configure the same thing, having the system property and the config file as options to configure a feature of the broker could lead to troubles,"I've set the value but the broker take another".

This change does not introduce such a conflict, it allows just to use system properties in config file values.
How would you ship a OSGi based solution that is using your bundle if you do not know at which location the user is execution it.
How could we configure the database storage location in a relative way to the configuration file?

andsel:
At the moment there is no way, but I expect that OSGi provide a way to pass in configuration to bundles or define some Configuration bundles that are essentially "configuration carrier services". OSGi it's a little bit difficult to me at the moment.

andsel:
I've dropped the OSGi support because it's too complicated the guarantee a decent level of integration (pax-runner doesn't fit well) and it's time consuming for me to fix it against other libs, because Moquette bundle isn't simply not used I drop it

PR #21, opened 2015-03-18
add support for system properties in configuration

maggu2810:
Change the configuration parser of the MQTT parser, so we could use
system properties in the value of a configuration key.
The configuration file that the OSGi bundle is using points normally to
${moquette.path}/config/moquette.conf.
So we could use a configuration file that contains an entry e.g.
persistent_store ${moquette.path}/storage/moquette_store.mapdb
So the real location of the configuration and the dabase storage could
be changed for different environments without touching anything else but
a system property.

Signed-off-by: Markus Rathgeb maggu2810@gmail.com

### DIFF
--- a/broker/pom.xml
+++ b/broker/pom.xml
@@ -120,6 +120,12 @@
             <version>2.1.4</version>
         </dependency>
 
+        <dependency>
+          <groupId>org.apache.commons</groupId>
+          <artifactId>commons-lang3</artifactId>
+          <version>3.3.2</version>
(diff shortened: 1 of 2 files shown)


maggu2810:
create link: #20

andsel:
I @maggu2810 I close the issue with the same motivation of issue #20, I think that to avoid the problem of OSGi configuration it could be better to leverage an existing OSGi configuration tool as Configuration Admin service.

PR #29, opened 2015-04-02
Resolve #28

### LINKED ISSUE #28: Dependency injection
Hi,

my goal is to be able to use moquette with my own IAuthenticator.

What I think is necessary to achieve this :
- SimpleMessaging should not be a singleton anymore (pretty easy since it is only used by Server class)
- All the dependency of it should be injected instead of being created inside init method (ex: IAuthenticator)
- regroup all authentication stuff into a package

If you agree, I'd like to do that in my fork and then crate a PR.

Are you ok with that ?

christopheblin:
cf https://github.com/andsel/moquette/issues/28


andsel:
For the motivations explained in #28 I can't accept this pull request

christopheblin:
I've added a comment to #28, we can "talk" about it there

PR #76, opened 2015-08-02
Hability for intercepting broker events

wagnerluis1982:
Discussion for these changes started at moquette list (https://groups.google.com/d/topic/moquette-mqtt/yvShxpQqza0/discussion)

### DIFF
--- a/broker/src/main/java/org/eclipse/moquette/interception/DefaultInterceptHandler.java
+++ b/broker/src/main/java/org/eclipse/moquette/interception/DefaultInterceptHandler.java
@@ -1,0 +1,47 @@
+/*
+ * Copyright (c) 2012-2014 The original author or authors
+ * ------------------------------------------------------
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * and Apache License v2.0 which accompanies this distribution.
+ *
+ * The Eclipse Public License is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * The Apache License v2.0 is available at
+ * http://www.opensource.org/licenses/apache2.0.php
+ *
+ * You may elect to redistribute this code under either of these licenses.
+ */
+package org.eclipse.moquette.interception;
+
+import org.eclipse.moquette.proto.messages.*;
+import org.eclipse.moquette.spi.impl.subscriptions.Subscription;
+
+/**
+ * Default empty implementation that can be used to create custom handlers
+ * without defining every method.
+ *
+ * @author Wagner Macedo
+ */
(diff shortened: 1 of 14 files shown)


bessonm:
Hi Wagner,

I looked at your pull-request,
I have quick questions before the weekend!
- Is it your intention to provide one ThreadPool for SubscriptionsStore and another one for ProtocolProcessor ?
- If I understand well, Interface Interceptor is intended for Observer purpose (so the 'notify*' semantic) and InterceptHandler is intended for your others needs (those I disagree with :) ) like modifying the payload ?
  Am I right ?
  - Because, the SystemProperty is named "intercept.handler" but is used for instantiating a concrete Interceptor and not an InterceptHandler.
- Why not call the Interceptor from SimpleMessaging.onEvent() instead of calling it after annotation dispatch (and thus, maintaining calls from differents places like SubscriptionsStore or ProtocolProcessor) ?

Andrea's opinion is welcomed !

Atenciosamente :)

Mael

andsel:
Hi Mael
as I can see the instance of the Interceptor is the same, but it's shared among the ProtocolProcessor and the SubscriptionsStore, so the ThreadPool is shared because created inside the BrokerInterceptor.
There are somethings stuff that is not so clean, I mean that one thing is the BrokerInterceptor is bound to the instance of the Observer that it will signal, that's broken, my intention is to have one global thread pool to do all the offline signaling stuff.
In my opinion the Interceptor is an utility class inside that should manage the Listeners and the dispatching to these Listeners, so I accept the pull request but I've to rework it because is not good enough.

 Andrea

wagnerluis1982:
Hi Mael and Andrea,
Sorry for my absence in this talk.

Andrea, I'm glad for your acceptance. I'm sure that what you have in mind for changes, it will be good.

Mael, I removed interceptors that modify the message, as I said in discussion list I'll continue with this in other unrelated branch. What Andrea said is right, the Interceptor is only one instance, so we have only one ThreadPool for all intercepting job. Property "intercept.handler" is the canonical name of an `InterceptHandler` implementation, see https://github.com/wagnerluis1982/moquette/blob/master/broker/src/main/java/org/eclipse/moquette/spi/impl/SimpleMessaging.java#L198 where you can see:

```
InterceptHandler handler = Class.forName(props.getProperty("intercept.handler"))
```

At all, for better understanding, `BrokerInterceptor` is actually a bridge to `InterceptHandler`. In my opinion, this design was needed for more decoupled code and to implement the asynchronous intercepting in a more concise way.

For Mael's last question, I didn't put the calls on `SimpleMessaging` because the events that arrived there are input events that are not processed yet, so we don't know if client is really connected (may be blocked by the authenticator), message was really published or topic was really subscribed (they can be blocked by the authorizator).

Wagner

PR #77, opened 2015-08-04
Adding a plugin manager.

WilliamKinaan:
The current plugin that you can use is a plugin to set what additional
work the server should do when receiving a published message on a
specific topic.

For example, there is a plugin called ChangeTopicPublishingNeeds, if activated, all the published messages to the broker will be subscribable under NewTopic topic.

Another plugin is ChangePayloadPublishingNeeds, the broker will change the message.

You can activate the plugins in the config/plugins.conf file that is located in the broker project (later it will be changed to PluginConfiguration project).

You can add your own plugin as well. That is done be:
1- Implementing the ICustomPublishingNeeds interface, or extending the AbstractCustomPublishingNeeds class
2- set the file in the configuration file under ICustomPublishingNeedsService key

Other plugins are coming soon.

I didn't do any major changes to the broker. I just changed the org.slf4j version. Plus, I called my plugin manager from insider the broker module. My whole code exists in the PluginConfiguration module, which is a module from moquette project.
Nevertheless, you can find all the changes by searching for "William" word. you will have a comment for everything that I have changed.

I hope we collaborate working on the broker because we need to embed it in our M2M framework, and we have some specific business requirements.

Regards


WilliamKinaan:
A new plugin for connect/disconnect is added

PR #78, opened 2015-08-06
Improving plugin manager's code and adding connect/disconnect plugin

WilliamKinaan:
The call for the inject object of the Google Guice library is static now under the Server class. 

you just need to do Server.getInjector()

Plus, a new plugin for connect/disconnect users is added.  You can add your own plugin by implementing the IConDisPlugin or extending the AbstractConDisImpl class. Then, set your class under IConDisService entry in the plugins file that is located in the broker module under config folder


andsel:
Hi William,
thanks for your effort but there something that doesn't fit well, Moquette has to be really minimal, and Google Guice dependency probably doesn't get so much improvement,is a plus that is not needed at the moment.
The naming convention is really poor, I mean that "AbstractConDisImpl" isn't such expressive.
These implementation doesn't respect the discussion happened on the mailing list about threading separation, so I've decided to not accept this PR.
 Best regards
  Andrea

WilliamKinaan:
Hi Andrea,

Thanks for your feedback.
Regarding the name, you could be right when it comes to ConDis, which means
ConnectDisconnect. I just tried to make it shorter. However, what about the
other plugin? the doAdditionalJob pluging?

Can I understand that you refuse to use Google Guice at all? If yes, kindly
did you make any plugin for this "additional job before publishing"  ? or
we should try to build another plugin using native JRE?

Regards,

On Sat, Aug 8, 2015 at 9:04 AM, Andrea Selva notifications@github.com
wrote:

> Hi William,
> thanks for your effort but there something that doesn't fit well, Moquette
> has to be really minimal, and Google Guice dependency probably doesn't get
> so much improvement,is a plus that is not needed at the moment.
> The naming convention is really poor, I mean that "AbstractConDisImpl"
> isn't such expressive.
> These implementation doesn't respect the discussion happened on the
> mailing list about threading separation, so I've decided to not accept this
> PR.
> Best regards
> Andrea
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/andsel/moquette/pull/78#issuecomment-128929520.

## 

William.

andsel:
Hi William,
as I know Google Guice is dependency injection framework, it's not a plugin
runtime as could be OSGi os NetBeans Platofrom. Now don't want the
complexity of OSGi and similar stuff, but also I think that a DI framework
doesn't match, so the path I'm following is to user the ServiceProvider API
defined by JDK or explicit plugins declarations into the moquette config
file, I've to think about.

 Andrea

WilliamKinaan:
Hi Andsel

Yes Google Guice is a dependency injection framework, which I used in
Moquette just to define, during the run time, the implementation for each
plugin as it is written in the plugings configuration file.

It is really easy, small and effective at the same time.
Anyway, every plugin manager should have a plugins configuration file to
allow the developers to specify their custom plugins. (if you prefer a
different approach, kindly tell me).

Now, I am investing in adding the plugin manager as you wanted it. In other
words, by native JDK code. However, if you still want to check the
availability of adding it using Google Guice, kindly tell me.

Many thanks,
Regards,

On Mon, Aug 10, 2015 at 12:23 PM, Andrea Selva notifications@github.com
wrote:

> Hi William,
> as I know Google Guice is dependency injection framework, it's not a plugin
> runtime as could be OSGi os NetBeans Platofrom. Now don't want the
> complexity of OSGi and similar stuff, but also I think that a DI framework
> doesn't match, so the path I'm following is to user the ServiceProvider API
> defined by JDK or explicit plugins declarations into the moquette config
> file, I've to think about.
> 
> Andrea
> 
> On Mon, Aug 10, 2015 at 10:12 AM, William Kinaan <notifications@github.com
> 
> > wrote:
> > 
> > Hi Andrea,
> > 
> > Thanks for your feedback.
> > Regarding the name, you could be right when it comes to ConDis, which
> > means
> > ConnectDisconnect. I just tried to make it shorter. However, what about
> > the
> > other plugin? the doAdditionalJob pluging?
> > 
> > Can I understand that you refuse to use Google Guice at all? If yes,
> > kindly
> > did you make any plugin for this "additional job before publishing" ? or
> > we should try to build another plugin using native JRE?
> > 
> > Regards,
> > 
> > On Sat, Aug 8, 2015 at 9:04 AM, Andrea Selva notifications@github.com
> > wrote:
> > 
> > > Hi William,
> > > thanks for your effort but there something that doesn't fit well,
> > > Moquette
> > > has to be really minimal, and Google Guice dependency probably doesn't
> > > get
> > > so much improvement,is a plus that is not needed at the moment.
> > > The naming convention is really poor, I mean that "AbstractConDisImpl"
> > > isn't such expressive.
> > > These implementation doesn't respect the discussion happened on the
> > > mailing list about threading separation, so I've decided to not accept
> > > this
> > > PR.
> > > Best regards
> > > Andrea
> > > 
> > > —
> > > Reply to this email directly or view it on GitHub
> > > https://github.com/andsel/moquette/pull/78#issuecomment-128929520.
> > 
> > ## 
> > 
> > William.
> > 
> > —
> > Reply to this email directly or view it on GitHub
> > https://github.com/andsel/moquette/pull/78#issuecomment-129353835.
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/andsel/moquette/pull/78#issuecomment-129411857.

## 

William.

PR #84, opened 2015-08-17
Encapsulating every intercept data

wagnerluis1982:
- Each event method caller has a data class, for instance, onConnect has InterceptConnectMessage.
- This way, we organize things better and can attach other informations without change the protocol message classes.
- Removed not more necessary UnmodifiableConnectMessage and UnmodifiablePublishMessage
  classes.


wagnerluis1982:
Sorry, I left an error in BrokerInterceptorTest, the Intellij isn't so proactive as Eclipse. I'm closing and open a new.

PR #89, opened 2015-08-25
PubRelEncoder to encode following MQTT specification

kazvictor:
Fixing PubRelEncoder to encode PubRel message flag to match what is in the MQTT specification (both 3.1.1 and 3.1) 
http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718054

See Appendix B: MQTT-3.6.1-1


andsel:
Thanks for the fix

Issue #91, opened 2015-08-27
NettyAcceptor  - Severe error during pipeline creation


unclezhao:
I want to test SSL pub and sub. But just one of pub or sub can connected to moquette server, when another connected, there was an exception like this:

```
1630 [main] INFO  NettyAcceptor  - Starting SSL on port 8883
1632 [main] INFO  NettyAcceptor  - Server binded host: 0.0.0.0, port: 8883
1632 [main] INFO  NettyAcceptor  - SSL is disabled
Server started, version 0.7
10030 [pool-1-thread-1] INFO  ProtocolProcessor  - cleaning old saved subscriptions for client <SSLSubscribeClientTest>
10070 [pool-1-thread-1] INFO  ProtocolProcessor  - Create persistent session for clientID <SSLSubscribeClientTest>
10105 [pool-1-thread-1] INFO  ProtocolProcessor  - Connected client ID <SSLSubscribeClientTest> with clean session true
10106 [pool-1-thread-1] INFO  ProtocolProcessor  - <SSLSubscribeClientTest> subscribed to topic <MQTT> with QoS 0 - MOST_ONE
17551 [nioEventLoopGroup-3-2] ERROR NettyAcceptor  - Severe error during pipeline creation
io.netty.channel.ChannelPipelineException: io.netty.handler.ssl.SslHandler is not a @Sharable handler, so can't be added or removed multiple times.
        at io.netty.channel.DefaultChannelPipeline.checkMultiplicity(DefaultChannelPipeline.java:463)
        at io.netty.channel.DefaultChannelPipeline.addLast0(DefaultChannelPipeline.java:135)
        at io.netty.channel.DefaultChannelPipeline.addLast(DefaultChannelPipeline.java:128)
        at io.netty.channel.DefaultChannelPipeline.addLast(DefaultChannelPipeline.java:119)
        at org.eclipse.moquette.server.netty.NettyAcceptor$4.init(NettyAcceptor.java:221)
        at org.eclipse.moquette.server.netty.NettyAcceptor$1.initChannel(NettyAcceptor.java:129)
        at org.eclipse.moquette.server.netty.NettyAcceptor$1.initChannel(NettyAcceptor.java:124)
        at io.netty.channel.ChannelInitializer.channelRegistered(ChannelInitializer.java:69)
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRegistered(AbstractChannelHandlerContext.java:158)
[...]

PR #103, opened 2015-09-22
Fix multiple client connection using SSL issue #91

runafter:
Because SSLHandler cannot be share with another pipeline, changed PipelineInitializer to create SSLHandler whenever initialize pipeline.


andsel:
@runafter Thanks a lot for your PR!
 Andrea

PR #109, opened 2015-10-01
Stopping all server threads and fixing ServerShutdownTest

### LINKED ISSUE #81: Moquette server does not stop completely upon stopServer()
Hi, 

I'm trying to use the moquette server within an application, and it looks like it doesn't stop completely when the "stopServer()" method is invoked. 
I'm using version 0.7...

The following is the example I'm trying: 

```
package org.test.moquette;
import java.io.IOException;
import org.eclipse.moquette.server.Server;
public class Main {

    public static void main(String[] args) {
        Server  broker = new Server();

        try {
            broker.startServer();
        } catch (IOException e) {
            e.printStackTrace();
        }
        // do something here...
        try {
            Thread.sleep(10000);
        } catch (InterruptedException e) {
            e.printStackTrace();
        }

        broker.stopServer();
    }
}
```

When I run this I see the following in the logs:

```
Aug 10, 2015 10:28:36 AM org.eclipse.moquette.spi.impl.SimpleMessaging processInit
INFO: Starting without ACL definition
Aug 10, 2015 10:28:36 AM org.eclipse.moquette.server.netty.NettyAcceptor initFactory
INFO: Server binded host: 0.0.0.0, port: 1883
Aug 10, 2015 10:28:36 AM org.eclipse.moquette.server.netty.NettyAcceptor initFactory
INFO: Server binded host: 0.0.0.0, port: 8080
Aug 10, 2015 10:28:46 AM org.eclipse.moquette.server.Server stopServer
INFO: Server stopping...
Aug 10, 2015 10:28:46 AM org.eclipse.moquette.spi.impl.SimpleMessaging onEvent
INFO: onEvent processing messaging event from input ringbuffer org.eclipse.moquette.spi.impl.events.StopEvent@796a56e5
Aug 10, 2015 10:28:46 AM org.eclipse.moquette.server.netty.NettyAcceptor close
INFO: Msg read: 0, msg wrote: 0
Aug 10, 2015 10:28:46 AM org.eclipse.moquette.server.netty.NettyAcceptor close
INFO: Bytes read: 0, bytes wrote: 0
Aug 10, 2015 10:28:46 AM org.eclipse.moquette.server.Server stopServer
INFO: Server stopped
```

However, it looks like there are still some threads running, and the server doesn't seem to have been stopped completely (see attached screenshot). 

moquette-server-threads-after-stop

Am I doing something wrong, or is that a known issue? 
Can you suggest a workaround for it? 

Thanks, 
Stoyan

kazvictor:
Stopping the ProtocolProcessor and the BrokerInterceptor on server stop. 
Fixing the ServerShutdownTest so that it no longer depends on thread counts. It instead check that there are no new threads running after the broker is started and stopped.

This should fix https://github.com/andsel/moquette/issues/81


andsel:
Hi Victor,
the code in SimpleMessaging is almost ok, I would remove the dumpTree because it's just debug, 
and also the `processor.stop()` is going to be removed because I'm on the way to remove the ring buffer (the front-ring buffer is already remove) at the moment remains only the back, but in next commits I'll drop it. This because the ring buffer was great to have a single threaded ProtocolProcessor, but in the near future the broker will move to be multithreaded with persistent data structures.

probably I should maintain:

``` java
LOG.debug("shutdown invoked");

if (m_interceptor != null) {
    this.m_interceptor.stop();
    LOG.debug("interceptor firer stopped");
}
m_storageService.close();
```

What I don't agree is the test, counting the number of thread is evil to assert something, because you don't have control on it so the the test fails in intermittent way.

PR #135, opened 2015-12-22
Change the executing name of class in moquette.sh

stannislin:
Change the executing name of class from Server to io.moquette.server.Server in moquette.sh file.


andsel:
Could you do the same also for moquette.bat?
It yes apply to this PR and then I'm going to accept.

 Thanks a lot
 Andrea

stannislin:
Hi, I already edited moquette.bat the same as moquette.sh.

PR #149, opened 2016-02-17
Enables client authentification for SSL

konradmichael:
Client authentification can be set as with the SSL Properties to
require the client certificate to be signed by the server.
Added jUnit test and test certificates


konradmichael:
I rechecked and all test are running fine on Eclipse and using maven.
Are you doing any magic with your server instances?

andsel:
Hi  @konradmichael could you explain better the meaning of the PR, which is the issue you found.

Thanks
 Andrea

konradmichael:
I am sorry that I did not explain the issue more, so here it comes.

**The Situation**:
We use moquette in our jUnit tests as an embedded MQTT server.
A new requirement was a certificate validation process on the Client. As the production server uses client authentication, the embedded Server also needs client authentication.

**The Issue** (feature request):
Moquette had no configuration parameters for using Client Authentication. Also I have not found a way to set these with JVM args.

**The solution**:
Seen in PR.

I hope this helps

PR #157, opened 2016-03-04
Added onMessageConsumed logic.

redboy1972:
Related to:  https://github.com/andsel/moquette/issues/156

### DIFF
--- a/broker/src/main/java/io/moquette/interception/AbstractInterceptHandler.java
+++ b/broker/src/main/java/io/moquette/interception/AbstractInterceptHandler.java
@@ -38,4 +38,7 @@ public abstract class AbstractInterceptHandler implements InterceptHandler {
 
     @Override
     public void onUnsubscribe(InterceptUnsubscribeMessage msg) {}
+    
+    @Override
+	public void onMessageConsumed(InterceptConsumedMessage msg) {}
 }
--- a/broker/src/main/java/io/moquette/interception/AbstractInterceptHandler.java
+++ b/broker/src/main/java/io/moquette/interception/AbstractInterceptHandler.java
@@ -40,5 +40,5 @@ public abstract class AbstractInterceptHandler implements InterceptHandler {
     public void onUnsubscribe(InterceptUnsubscribeMessage msg) {}
     
     @Override
-	public void onMessageConsumed(InterceptConsumedMessage msg) {}
+	public void onMessageAcknowledged(InterceptAcknowledgedMessage msg) {}
 }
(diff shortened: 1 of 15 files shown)


andsel:
Hi @redboy1972  thank's for your PR but can'be accepted by 2 reason:
- the first trivial, the name of the "event" shouldn't be messagceConcumed, but messageAcknowledged, because that's the meaning,once you get an ack you could only say that the client acked the mssage, not that it has been cosnumed by the client, it's a question of perspective, and we need to keep the viewpoint from the server not the client.
  - the second, you changed signature of methods (added username) without discussing before and that's not good in general

redboy1972:
I will modify as requested to change messageConsumed -> messageAcknowledged.  MessageConsumed came first to mind as I come from writing ActiveMQ plugins which calls it such in org.apache.activemq.broker.BrokerFilter.

Instead of changing the method signature how would you suggest ferrying the username to the InterceptHandler?

From my use case the clientID is randomly generated by most clients on behalf of their API and isn't typically useful to me.  What I need is the username of the logged in user that is interacting with the system.

andsel:
Hi @redboy1972 you are suggesting to forward the concept of which user has been acknowledged, because you need the concept in you client code, you are using this as the a session but from MQTT specs and the broker standpoint that's the already exists an is the clientId. If you need the association between the clientID and the username there are 2 options:
-  the first that the embeding cde take care of this
- the SessionClient could be passed to the messageAcknowledged method, but it's a little bit risky because I can't trust what the embedding code does with that session handler

redboy1972:
I haven't changed anything to my knowledge.  The client id is still used as it always was and the username has not taken over any part of that role.  The only thing I did is provide to the InterceptHandler the logged in username.  

I too would not expose SessionClient.  If the cde (assuming this is the InterceptHandler) does this, which is possible, then an additional redundant mapping of clientId -> username would be required.  I will be running my broker with 200k or more concurrently connected clients.  That is a lot of ram to use to store redundant information the broker already knows.

My first implementation was using a Map to do ClientID -> username by plugging into onConnect (has info), onDisconnect (to remove) and having all the other methods get from the map.  Then when debugging other areas realized that the broker was storing the info too and it seemed redundant.  

I think a lot of people see messages being sent/read by users not by clients.  I see providing this accounting and auditing piece of info to a InterceptHandler as an enhancement.  If you still feel it a detriment, I will back it out as my real concern is the onMessageAcknowledged behavior.

andsel:
Merged by hand

PR #173, opened 2016-05-04
fix CSV list bug, collect CSV list into constant, provide better logging on disabled SSL

windbender:
There is a bad bug on the CSV list of subprotocols for the SSL web socket.  This was fixed

### DIFF
--- a/broker/src/main/java/io/moquette/server/netty/NettyAcceptor.java
+++ b/broker/src/main/java/io/moquette/server/netty/NettyAcceptor.java
@@ -54,6 +54,8 @@ import java.util.List;
  */
 public class NettyAcceptor implements ServerAcceptor {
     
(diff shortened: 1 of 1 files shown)


andsel:
Hi  @windbender, thank's for your PR, but where is the "bad bug" at the end
you just moved a string in constant, how the bug manifested it self before?

windbender:
Yeah, it's hard to see.    When setting up the handler chain for the secure web socket, there is supposed to be a CSV string containing the various subprotocols as a parameter to WebSocketServerProtocolHandler.  However there is a missing comma.  And since the missing comma was next to the subprotocol  "mqttv3.1"  and that is the most common mqtt version, it basically meant one could not connect via secure web sockets.  It took me a while to figure this out :-)

So yeah, "bad bug" in that there was zero expected functionality, but at the same time, it was very easy to fix.

andsel:
You area right,  now I see it!! Great catch!!
 Andrea

PR #174, opened 2016-05-05
allow turn off of nonSSL MQTT,  and also accepts empty string to indicate turn off

windbender:
I found I wanted to turn off non-SSL  MQTT

also I found that some properties systems didn't allow null as the value on a put(), so enabled use of the empty string as well.

### DIFF
--- a/broker/src/main/java/io/moquette/server/config/ClasspathConfig.java
+++ b/broker/src/main/java/io/moquette/server/config/ClasspathConfig.java
@@ -43,13 +43,13 @@ public class ClasspathConfig extends IConfig {
         }
         Reader configReader = new InputStreamReader(is);
         ConfigurationParser confParser = new ConfigurationParser();
+        m_properties = confParser.getProperties();
         assignDefaults();
         try {
             confParser.parse(configReader);
         } catch (ParseException pex) {
             LOG.warn("An error occurred in parsing configuration, fallback on default configuration", pex);
         }
-        m_properties = confParser.getProperties();
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


andsel:
I don't understand the meaning of the empty string check, I mean that if the configuration is correctly created by the IConfig subclasses than the value of "empty" or not specified properties is already null (probably a NullObject should be better)

windbender:
WRT the empty string check.   A  Properties object may have a "default"  inner Properties object.  If the value of a key is null then Properties.getProperties('key') will return the value of the inner "default" properties object.  In a situation where said default properties object contains a port value, then "null" in the top level properties object would be insufficient to turn off the listener since the default value would be returned.   I know that the Config code is not currently written this way (though perhaps it should ).

andsel:
To disable an acceptor what do you think about this idea:
If the three parameters that define the ports in moquette.conf file (_port_, _websocket_port_, _ssl_port_) could contains the value of ports (integer numbers) and also the value _undefined_ to disable that acceptor, instead of using empty strings as special value for the meaning "disable this port listener" ?

windbender:
I think for standalone moquette that would work.

My use case is that moquette is embedded in an instance of dropwizard. So I'm using dropwizard's config system, and then programmatically setting the values into moquette's config.  In this case the "type" of the variable is less clear.  did you mean to set the the value to a string "undefined"  or to null ?  For the reasons I described above null may not work since then things fall back to the default values.  I figured empty string would work.  Another option is an invalid port value such as -1 which is nice because it's still an integer.

andsel:
For me -1 or null are not so clear values,is -1 to be intended as "any
port" or as "disabled" the same reaspn is valid for empty string. I agree
with you that the type is not so clear but I prefer the clearing of
meaning,so the port value could be a int port number or the string value
"disabled"
Il giorno 12/mag/2016 08:09, "windbender" notifications@github.com ha
scritto:

andsel:
Merged by hand adding the 'disabled' keyword

PR #185, opened 2016-06-26
Add clientID to the IAuthenticator interface

bemisguided:
I have need for the clientID to be passed into my IAuthenticator implementation. The nodejs MQTT (mosca) supports this in their interface.

This is the "simple" fix which basically breaks compatibility for all IAuthenticator implementations. I updated all those included in this code base, however, of course external users will break upon upgrade.

I am happy to come at this again if you want a more backward compatible solution (I just figured maybe you're okay breaking compatibility at this point given it's still a minor release number... and it's not too hard for someone to update their code to become compatible again).

Let me know.

### DIFF
--- a/broker/src/main/java/io/moquette/spi/impl/ProtocolProcessor.java
+++ b/broker/src/main/java/io/moquette/spi/impl/ProtocolProcessor.java
@@ -58,7 +58,7 @@ import org.slf4j.LoggerFactory;
 
 /**
  * Class responsible to handle the logic of MQTT protocol it's the director of
- * the protocol execution. 
+ * the protocol execution.
  *
  * Used by the front facing class SimpleMessaging.
  *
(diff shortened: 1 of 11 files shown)


andsel:
Hi @bemisguided in which practical cases you authenticate using the triple (clientId, username, password) instead of usual credentials?

bemisguided:
In our IoT scheme the user credentials and the device are not the same entity and therefore the client ID needs to be factored into our authentication scheme.

windbender:
I think this is a fair concept.  +1  from my perspective.

Also nice job on collecting some of the inline constants into a single place for the tests! way to be a "leave it better than you found it" coder.

bemisguided:
Thanks for merging my PR (and the coding complement)! I'm happy to contribute (and do as best a job I could). 

If you could let me know when this is formally released I would appreciate. Also, if I come across anything else I need you may see other PRs (nothing so far but you never know).

PR #186, opened 2016-06-28
feature/cluster : add default clustering feature

mackristof:


diegovisentin:
Why is payload converted into a string in HazelcastMsg ?
And if this is necessary, maybe it's better to specify the charset
`this.payload = new String(payload, "UTF-8");`
as counterpart of
`publishMessage.setPayload(ByteBuffer.wrap(msg.getMessageObject().getPayload().getBytes("UTF-8")));`

mackristof:
well done @diegovisentin, thanks . payload can be byte array without problems

diegovisentin:
Is there any problem to merge this feature into the main branch?
(I hope that it's only a question of time)

andsel:
No, I'm studying a little bit more Hazelcast to become confident with the
PR. One I've finished, I'll review and merge.

 Andrea

andsel:
I gave a look  at PR and as a first step it's good. It's just a forwarder of the publishes, at the moment miss all the part of replication of sessions, replication of the message store and the join/disjoin of the members to the cluster; but these are stuff that has to be worked out with time.
 Before merge I ask if someone could the fix the test io.moquette.server.ServerIntegrationHazelcastHandlerInterceptorTest, it fail as reported by the TravisCI.
 Thanks 
   Andrea

kevto:
@mackristof regarding the failing test. I debugged your unit test and I found out as when the cluster replicates the messages over all the instances, it also replicates the message, multiplied by the amount of Moquette server instances connected to the cluster, into the BlockingQueue of the MessageCollector object.

In the screen below I added two more objects that will just poll the queue. In the variable dialog you will see the content of these new objects and each message replicated twice in the queue.

screenshot from 2016-07-17 02 09 24

To be sure of what I'm saying, I added another instance to the cluster and as expected, the message is multiplied in the queue by the amount of instances connected to the cluster.

screenshot from 2016-07-17 02 12 07

A fix would be to poll the published message twice (or the amount of instances you've connected to the cluster) and then the test would pass.

I hope this will help you out!

andsel:
I'm looking at this, seems like there is some interference on storage, probably the 2 broker instances share the same data storage files. I see from logs that <Publisher> connect on port 1883, but when publish on logs appears `ProtocolProcessor processPublish 358  - PUBLISH on server 1884 from clientID <Publisher>`, the publisher connect on node 1, but publish on the port of node 2, that ugly, probably some unwanted shared state.

PR #188, opened 2016-07-17
One param less in ClientSession subscribe method

kevto:
**subscribe** method of **ClientSession** class has a param named _topicFilter_. _topicFilter_ has no other use than for logging. Because when I was trying to work on #186 and tried to figure out why the test was failing, I looked into the subscribe method of the client in the broker to see if it was really overriding and all. Then I stumbled upon the unit test and tried to break the test by changing the value of _topicFilter_. Clearly it was giving me no different result. 

So that's why I basically removed it as I thought it'd be better and less confusing (not to mention it serves no real purpose in my opinion because you can get the _topicFilter_ from the Subscription object).

Also moved the unit test as was asked for in the TODO comment.

### DIFF
--- a/broker/src/main/java/io/moquette/spi/ClientSession.java
+++ b/broker/src/main/java/io/moquette/spi/ClientSession.java
@@ -94,9 +94,9 @@ public class ClientSession {
         return "ClientSession{clientID='" + clientID + '\'' +"}";
     }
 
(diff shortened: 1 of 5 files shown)


andsel:
It's a thing I was thinking to fix sometimes ago, well done

Issue #190, opened 2016-08-05
Null or Empty ClientID for Connect


peacecoder:
According to the specification of MQTT 3.1.1 (shown in the picture), I think if the implementation in moquette is not correct. Why it doesn't close the channel? Meanwhile, if the broker doesn't allow anonymous, why it close the channel when it doesn't close the channel?

![image](https://cloud.githubusercontent.com/assets/10008870/17434744/e17ef500-5b3e-11e6-9515-7e9afff121a7.png)

https://github.com/andsel/moquette/blob/master/broker/src/main/java/io/moquette/spi/impl/ProtocolProcessor.java#L164-L170

![image](https://cloud.githubusercontent.com/assets/10008870/17434831/46f51e82-5b3f-11e6-968d-06e255c96cc7.png)

kevto:
Done the work for you @peacecoder . I think this meets the expectation of MQTT's specification. Please take a look at the references above my comment, to see if there's room for any improvement.

peacecoder:
Thanks @kevto, I think you change meet the specification. And Another question is that if mqtt broker doesn't allow anonymous user, it should close the connection. Therefore why it doesn't close the channel? I am very confuse about it.

https://github.com/andsel/moquette/blob/master/broker/src/main/java/io/moquette/spi/impl/ProtocolProcessor.java#L172-L190

![image](https://cloud.githubusercontent.com/assets/10008870/17460805/77b02fe4-5ca7-11e6-9e0a-919a294eef47.png)

andsel:
Fixed by @kevto with PR #191

peacecoder:
@andsel @kevto does anyone answer my second question before close this issue?

kevto:
@peacecoder oh. Well, you were right on closing the channel and without notifying you, I made those changes so when anonymous clients try to connect to immediately close the channel. Those changes are merged by now in the master branch. Take a look at #192

peacecoder:
@kevto awesome, thanks for fix it.

PR #191, opened 2016-08-06
Allow zero byte client identifiers in the connect message.

kevto:
Work done based on issue #190 

All connect tests pass:
screenshot from 2016-08-06 11 23 35


kevto:
I can't see why Travis CI build failed for any reason. Only switched the names of two unit tests.

PR #195, opened 2016-08-11
dropMessagesInSession for clientId

myqhit:
clear the m_persistentMessageStore should use message guid to clear the map,not use clientId to clear the map,because we put message in the map use guid for the key


andsel:
Your assumption is incorrect, if the broker has a retained message, if the publishing client disconnects, the message has to still be persistent for future. Please next time, if you find something that doesn't work,  open an issue and discuss on it before wasting time.

PR #201, opened 2016-08-18
Bugfix - properly shutdown ProtocolProcessorBootstrapper

mjacoby:
Based on https://github.com/andsel/moquette/pull/109 if added a bugfix to properly shutdown the server and also a modified version of the integration test provided by @kazvictor.


andsel:
Hi  @mjacoby this test it's a little bit convoluted and hard to understand, just to to check that all threads terminate.
For example, in my opinion the scheduling at a fixed rate a thread to check the count of thread and then move forward the latch, it's a little bit non sense; in this case the thread that's running the test could do spin check of the condition (#number of current thread == #of threads at setup time).
Then another thing that doesn't sound good is the if, control structures has to be avoided in tests, else you are verifying  two things, and in this case is used just to debug the test.
 Sorry but this PR can't be accepted from my side.
  Regards
   Andrea

PR #202, opened 2016-08-18
Bugfix for ServerRestartIntegrationTest - Persistence file could not be deleted under Windows

mjacoby:
ServerRestartIntegrationTest failed because the the server was started before trying to delete the persistence files which works on Linux but fails on Windows because the files can't be deleted as they are opened by the server. 
This actually prevented building the broker on Windows because the test failed.


andsel:
Thanks for the PR

PR #205, opened 2016-09-10
Changing some private classes/methods to public

teranisi:
To embed moquette's MQTT protocol processor in an external program, some classes and methods are needed to be public. I hope this would be merged in the future release.

### DIFF
--- a/broker/src/main/java/io/moquette/server/DefaultMoquetteSslContextCreator.java
+++ b/broker/src/main/java/io/moquette/server/DefaultMoquetteSslContextCreator.java
@@ -37,7 +37,7 @@ import java.security.cert.CertificateException;
  *
  * Created by andrea on 13/12/15.
  */
-class DefaultMoquetteSslContextCreator implements ISslContextCreator {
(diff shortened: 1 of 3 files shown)


andsel:
To embed moquette I don't think this is the best way to open the ProtocolProcessor, this is a design flaws. Moquette has other extsion points (see the https://github.com/andsel/moquette/blob/master/embedding_moquette/src/main/java/io/moquette/testembedded/EmbeddedLauncher.java). I close this issue as not applicable.

teranisi:
OK, I understood that was a design flaw.
The requirements of my application were as follows:
1) want to access MessagesStore and SessionsStore to see which client sessions/messages were stored.
2) don't want to load HazelCast library.
I want the 'Server' class to have some interfaces to do above if ProtocolProcessor is not an open class.

andsel:
MessagesStore and SessionsStore is intended for Broker private access, it's correct that other application parts read/update the data managed by the broker. Probably you need an API to ask the broker some stuff. But the way to store messages and session is private broker responsibility. If you need such a thing, please open a feature request describing very well what's the API (SPI) you need.

Issue #207, opened 2016-09-14
Server API to get subscriptions information


teranisi:
If external programs that embed moquette using `Server` class want to know about subscriptions information (i.e. a set of `Subscription`s), they can implement InterceptHandler that overrides `onSubscribe(InterceptSubscribeMessage msg)` and collect the information from the msg argument.
But currently, the external programs cannot know about the stored subscriptions information after restarting the Server. (If a subscriber starts a session without `clean session` flag, the subscriber keeps the subscription state after reconnect even when they don't subscribe again in QoS 1,2).
There should be some API for `Server` to ask current subscriptions information from external programs. I think following API is enough.

`public List<Subscription> listAllSubscriptions();`

kevto:
Could this be something that you might be looking for? [Link to MQTT Wiki](https://github.com/mqtt/mqtt.github.io/wiki/SYS-Topics). It doesn't include your actual request to iterate through the subscriptions but it somewhat looks like an API. Perhaps we could add more SYS calls.

For SYS calls, there's already an issue. [Link](https://github.com/andsel/moquette/issues/6)

However, I wouldn't mind to have interfaces that allow projects to obtain such information that embed Moquette.

@andsel what are your thoughts?

teranisi:
As I wrote (and you wrote), at least we need an interface to iterate subscriptions on Server. But yes, it is good to have APIs like 'SYS calls' shown in the wiki.

andsel:
There are pros and cons in both solutions.
The "SYS" topic is agnostic of the language/platform, and is more portable, but you need to pass through a client, and if your are embedding Moquette it's counterintuitive and not clean.
The "API" way is more clean but it's tight to Moquette.

To solve @teranisi  issue the Java API is probably the best one, the SYS is another stuff to have in Moquette but it's not strictly related to this issue

PR #212, opened 2016-09-16
Add 'getSubscriptions' API on Server.

teranisi:
Here is an implementation of [#207](https://github.com/andsel/moquette/issues/207) and related test.

### DIFF
--- a/broker/src/test/java/io/moquette/server/ServerIntegrationEmbeddedPublishTest.java
+++ b/broker/src/test/java/io/moquette/server/ServerIntegrationEmbeddedPublishTest.java
@@ -97,6 +97,17 @@ public class ServerIntegrationEmbeddedPublishTest {
         m_server.internalPublish(message);
     }
 
+    private void connectNoCleanSession() throws Exception {
+        MqttConnectOptions opts = new MqttConnectOptions();
+        opts.setCleanSession(false);
+        m_subscriber.connect(opts);
+    }
+
+    private void subscribeToWithQosAndNoCleanSession(String topic, int qos) throws Exception {
+        connectNoCleanSession();
+        m_subscriber.subscribe(topic, qos);
+    }
+
     private void verifyNoMessageIsReceived() throws Exception {
         MqttMessage msg = m_callback.getMessage(false);
         assertNull(msg);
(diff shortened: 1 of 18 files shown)


broker/src/test/java/io/moquette/server/ServerIntegrationEmbeddedPublishTest.java
```diff
@@ -97,6 +97,17 @@ private void internalPublishToWithQosAndRetained(String topic, QOSType qos, bool
         m_server.internalPublish(message);
     }
 
+    private void connectNoCleanSession() throws Exception {
+        MqttConnectOptions opts = new MqttConnectOptions();
+        opts.setCleanSession(false);
+        m_subscriber.connect(opts);
+    }
+
+    private void subscribeToWithQosAndNoCleanSession(String topic, int qos) throws Exception {
```
kevto, line 106: Unit test is not verifying any constraints or whatsoever.

broker/src/test/java/io/moquette/server/ServerIntegrationEmbeddedPublishTest.java
```diff
@@ -97,6 +97,17 @@ private void internalPublishToWithQosAndRetained(String topic, QOSType qos, bool
         m_server.internalPublish(message);
     }
 
+    private void connectNoCleanSession() throws Exception {
```
kevto, line 100: Unit test is not verifying any constraints or whatsoever.

kevto (changes requested):
Two notes regarding unit tests.

kevto:
Perhaps also an idea to make these `List` instances immutable? I don't know what your exact intentions are but if you're only looking for to read data and not mutate any of it, to make these objects immutable rather than allowing out of package classes to modify `Subscription` data.

andsel:
I agree with @kevto no mutable state has to be shared. I'll review the pr
when I'm back from holiday
Andre
Il giorno 16/set/2016 16:19, "Kevin" notifications@github.com ha scritto:

teranisi:
My intention was to provide a read interface for an external program. The objects can be immutable.

PR #233, opened 2016-11-12
adds storage and api for keeping track of most recent message time

windbender:
As the commit states, this PR is around tracking most recent messages received for each client. And also provides a simple API to either find the most recent message per client or a list of clients which have messages more recently than a given Date object.

### DIFF
--- a/broker/src/main/java/io/moquette/server/ConnectionDescriptor.java
+++ b/broker/src/main/java/io/moquette/server/ConnectionDescriptor.java
@@ -17,6 +17,8 @@ package io.moquette.server;
 
 import io.netty.channel.Channel;
 
+import java.util.Date;
+
 /**
  * Value object to maintain the information of single connection, like ClientID, Channel,
  * and clean session flag.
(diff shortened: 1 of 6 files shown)


andsel:
Seems o as PR, but where should be useful this kind of fix, there is real cases where the time of last message received on a channel is necessary?

windbender:
@andel great question.  I think this is the sort of information which should be available via $SYS topics, but obviously this PR didn't get that far.  In my embedded app, I did actually convert the info in this API to a TOPIC, but given it's pretty application specific implementation I did not include it.

I also considered a unit test, but the only complexity in this thing is the time based stuff, and as you know this is pretty difficult to write a unit test for. ( Unless one goes to the effort of centralizing all time sources for the library ).  I suppose one could still write a unit test which would test for the existence of a mostRecentReceivedMsgTime value after a message is received.  If you are considering accepting this PR, then I will add that somewhat simple test.

OH and to answer your question. My application needs to know the most recent time any sort of messages is received from a client, including PINGREQ.  This is used to determine which clients are "currently" connected.  If you think there is a more appropriate way to determine this info, I'd be happy to consider changing.

andsel:
Hi @windbender if I'm right the query methods for last seen timestamp are getRecentClientPingTimes and mostRecentPing in ProtocolProcessor, but they should return also the time of a publish from a client not only the ping.
 Than, which is the client of this methods? The embedding application?

PR #245, opened 2016-12-27
MessageDigest is not thread-safe

nemecec:
Use DigestUtils instead, it is also more convenient (and Moquette already has a dependency on commons-codec).


andsel:
Great catch, thanks a lot @nemecec

PR #246, opened 2016-12-27
Remove compile-time dependency on log4j

nemecec:
Move log4j dependency from broker to dist - otherwise it is propagate transitively to projects using Moquette as a library.

Also removed redundant logback-core dependency from OSGi tests (it already has logback-classic).


andsel:
Thanks!

PR #251, opened 2017-01-10
Improve performance of message interception

lbarriosh:
Create a new property to set the size of the broker interceptor thread pool.
Make sure that the broker interceptor thread pool is always shut down.
Reduce interception latency by avoiding the calls to all the message interceptors. The intercept handlers can now specify which message types they will process.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,8 @@ broker/moquette.log
 broker/runner
 /bundle/target/
 /bundle/runner/
+osgi_test/target/
+perf/target/
 *~
 *log*
 .DS_Store
--- a/.gitignore
+++ b/.gitignore
@@ -14,8 +14,8 @@ broker/moquette.log
 broker/runner
 /bundle/target/
 /bundle/runner/
-osgi_test/target/
-perf/target/
+/osgi_test/target/
+/perf/target/
 *~
 *log*
 .DS_Store
--- a/broker/pom.xml
+++ b/broker/pom.xml
@@ -6,7 +6,7 @@
         <relativePath>../</relativePath>
         <artifactId>moquette-parent</artifactId>
         <groupId>io.moquette</groupId>
-        <version>0.9-SNAPSHOT</version>
+        <version>0.9.1.sofia</version>
     </parent>
 
     <artifactId>moquette-broker</artifactId>
--- a/broker/src/main/java/io/moquette/BrokerConstants.java
+++ b/broker/src/main/java/io/moquette/BrokerConstants.java
@@ -22,6 +22,7 @@ import java.io.File;
  */
 public class BrokerConstants {
     public static final String INTERCEPT_HANDLER_PROPERTY_NAME ="intercept.handler";
+    public static final String BROKER_INTERCEPTOR_THREAD_POOL_SIZE = "intercept.thread_pool.size";
     public static final String PERSISTENT_STORE_PROPERTY_NAME = "persistent_store";
     public static final String AUTOSAVE_INTERVAL_PROPERTY_NAME = "autosave_interval";
     public static final String PASSWORD_FILE_PROPERTY_NAME = "password_file";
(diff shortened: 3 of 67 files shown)


andsel:
Hi @lbarriosh 
 thanks a lot for your interest in the project and for PR, I have some questions about it:

 - Why the return type has been erased from methods add/removeInterceptHandler in broker/src/main/java/io/moquette/interception/Interceptor.java ?
 - Which is the reason to add an ID read-only property in InterceptHandler?
 - The design that make all Intercept<MessageType>Message extend the marker interface InterceptMessage seems more complicated than using a simple enum with message types; I don't expect to continue adding Intercept<MessageType>Message to justinfy this dynamicity.

Thanks,
 Andrea

lbarriosh:
Hi @andsel

First of all, thanks a lot for your quick response. This pull request (and two extra ones that I'll be sending you this week) can seem a bit odd, so I'd like to give you some context first.

For the last two years, we've been using moquette in [Sofia2](http://sofia2.com/), Indra's IoT platform. In order to make it dependable for a production environment, we have included several patches in the 0.6.x codebase, and we also have added a lot of log traces. 

I am merging our patches and our log traces into the latest moquette codebase. Please, feel free to include or omit them, but have a look at our error handling patches if you can: they prevent some strange crashes of the broker.

That being said, I'll try to answer your questions as best as I can:

> Why the return type has been erased from methods add/removeInterceptHandler in broker/src/main/java/io/moquette/interception/Interceptor.java

In the current implementation, a lot of no-op tasks are submitted to the interceptor thread pool. This increases the latency of the message interception. 
Now, the message interceptors are stored in several lists (one for each intercepted message type). The boolean value that the methods add/removeInterceptHandler can still be generated, but I preferred to remove them to keep the methods as simple as possible.

> Which is the reason to add an ID read-only property in InterceptHandler?

I'll include new log traces in the BrokerInterceptor class. The getID() method will be used to identify which interceptors will be added or removed at each call to add/removeInterceptHandler. The name of the class is not relevant enough when you have several instances of the same class configured in different ways. This usually happens when you rely on an IoC container (like spring's).

The AbstractInterceptHandler class contains a default implementation of the getID() method, so we don't have to implement it at each existing InterceptHandler.

> The design that make all InterceptMessage extend the marker interface InterceptMessage seems more complicated than using a simple enum with message types; I don't expect to continue adding InterceptMessage to justinfy this dynamicity.

The intercept messages do not have a common type. When you use generics, having a base type at the root of the hierarchy is usually handy. But you can remove the InterceptMessage interface: I'll not be using it for now.

andsel:
Hi Luis,
thanks a lot for the great work, but I have some questions:
- why adding NETTY_CHANNEL_TIMEOUT_SECONDS_PROPERTY_NAME as timeout on netty channel, while this could potentiallly interfere with the same IdleStateHandler class that is used for the keepAlive timeout.
- then I've have seen all debug log decorated with If (LOG.isDEbugEnabled()) but that line is no usafull, SLF4J doesn't any string manipulation if log level doesn't match, I'll remove all once integrated the patch.

 Andrea

lbarriosh:
Hi @andsel 

Thanks for your response. 

> why adding NETTY_CHANNEL_TIMEOUT_SECONDS_PROPERTY_NAME as timeout on netty channel, while this could potentiallly interfere with the same IdleStateHandler class that is used for the keepAlive timeout.

In our deployments, the low-level netty configuration properties are usually left unchanged. But sometimes the environments require a little bit of fine tuning, and having to recompile the broker and re-upload it several times is not very convenient.

You're right, the property can interfere with the idle state handler. That's why I've included a default value (the same it had before). Maybe we can move that constant out of the BrokerConstants class to another constants class because the property is meant to be used only when necessary. 

> then I've have seen all debug log decorated with If (LOG.isDEbugEnabled()) but that line is no usafull, SLF4J doesn't any string manipulation if log level doesn't match, I'll remove all once integrated the patch.

You are right, no string manipulation is performed on this cases. But the arguments of the logger call are evaluated anyway. For instance, if we had a class like this

public class PayloadType ... {
// Field declarations
...
public String toString() {
// serialization code
}
...
}

and a log trace like this

logger.debug("Sending object. MqttClientId = {}, payload = {}", clientId, payload.toString())

the payload object  would be always converted to string, even if debug traces are disabled.

The broker has very verbose debug traces in some points (for example, whenever subscription trees are printed out). And this reduces performance if we don't decorate them. In less extreme scenarios (for example, when the arguments of the logger call are attributes or getters), I agree with you: the if's can be left out.

If you have any other questions, just let me know.

Luis

PR #254, opened 2017-01-19
Subscritions store cleanup

AndreasMager:


andsel:
Hi Andreas
thankyou for your PR, but before creating a PR usually is better to discuss the issue the PR is solving in GoogleGroups or creating a question issue. Said that, could you better describe you PR please? 

For example you said that LinkedQueue is worst than ArrayList, have you got done any measurement on this improvement inside Moquette, because that queue was originally an ArrayList and was decided to move to Queue because is FIFO and not random access list.

AndreasMager:
I will splitt this PR and add some measurement data. The LinkedBlockingQueue is not the best option for a singel Thread enviroment.

PR #255, opened 2017-01-23
PerformanceTest added for the SubscriptionsStore#matches

AndreasMager:


andsel:
Hi @AndreasMager 
thanks for the contribution, being this a hand-launched main class used only to do measurement of some internal stuff of the broker, could you move the code in the perf/src/java?
 Thanks
  Andrea

AndreasMager:
Ok I will move it tomorrow.

AndreasMager:
@andsel done

PR #256, opened 2017-01-23
MemorySessionStore#m_persistentSubscriptions: Map Layout changed

AndreasMager:
If one clientid subscribes a lot of topics the old implementation
struggles with the for loops


andsel:
Hi @AndreasMager 
the contribution is almost ok, but don't like so much the usage of implementation classes (HashMap for example) when I can use the abstract interface (Map)

For example
```java
HashMap<String, Subscription> clientSubscriptions = m_persistentSubscriptions.get(clientID);
```

should be rephrased as 
```java
Map<String, Subscription> clientSubscriptions = m_persistentSubscriptions.get(clientID);
```

PR #259, opened 2017-01-26
Protocol processor test fixed

AndreasMager:


andsel:
Hi @AndreasMager  could you describe better which problem fixes this PR?
Thanks
   Andrea

broker/src/test/java/io/moquette/spi/impl/ProtocolProcessorTest.java
```diff
@@ -426,14 +426,22 @@ public void testCleanRetainedStoreAfterAQoS0AndRetainedTrue() {
         //prepare and existing retained store
         NettyUtils.clientID(m_channel, "Publisher");
         ByteBuffer payload = ByteBuffer.allocate(5).put("Hello".getBytes());
+        payload.flip();
```
AndreasMager: If the flip method isn't called the internal structure handles this message as empty.
Empty messages are not stored in the RetainStore

broker/src/test/java/io/moquette/spi/impl/ProtocolProcessorTest.java
```diff
@@ -426,14 +426,22 @@ public void testCleanRetainedStoreAfterAQoS0AndRetainedTrue() {
...
         m_processor.processPublish(m_channel, msg);
-        
+
+        Collection<IMessagesStore.StoredMessage> messages = m_messagesStore.searchMatching(new IMatchingCondition() {
+            public boolean match(String key) {
+                return  SubscriptionsStore.matchTopics(key, FAKE_TOPIC);
+            }
+        });
+        assertFalse(messages.isEmpty());
+
```
AndreasMager, line 471: This part checks if there is a message in the retain storage. The old implementation would fail on this assert.

netty_parser/src/main/java/io/moquette/parser/proto/messages/PublishMessage.java
```diff
@@ -42,7 +42,7 @@ public void setTopicName(String topicName) {
     }
 
     public ByteBuffer getPayload() {
-        return m_payload;
+        return m_payload.duplicate();
```
AndreasMager, line 45: If someone reads the buffer with the wrong methods he won't change the position pointers.

andsel:
Hi, I've changed the protocol codec part, passingo to Netty's 4.1 codec, this shouldn't be yet applicable after I merge.

AndreasMager:
No problem i'll rebase it tomorrow.

AndreasMager:
I like the new netty codec the test setup error is gone. I have rebase the check for a correct test setup

PR #261, opened 2017-01-27
Adding assertj to test dependencies

AndreasMager:
Assertj is a great library for writing junit test.

### DIFF
--- a/broker/pom.xml
+++ b/broker/pom.xml
@@ -136,6 +136,15 @@
             <version>9.2.0.M1</version>
             <scope>test</scope>
         </dependency>
+
+        <dependency>
+            <groupId>org.assertj</groupId>
+            <artifactId>assertj-core</artifactId>
+            <!-- use 2.6.0 for Java 7 projects -->
+            <version>2.6.0</version>
+            <scope>test</scope>
+        </dependency>
+
         <dependency>
             <groupId>com.h2database</groupId>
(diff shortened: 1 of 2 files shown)


andsel:
AssertJ seems promising, but has few stars on GitHub. Probably Hamcrest matchers or probably I would introduce Spckframework or JUnit 5 which should be GA for Q1 of 2017 (practically tomorrow). What do you think?

AndreasMager:
I work for 2 years with assertj now. The collection methods and error checks in java 1.8 are nice. I'll take a look on your Suggestions. My preferred test library's are awaitility, mokito and assertj.

AndreasMager:
JUnit 5 is java 1.8.
I thing Hamcrest matcher can be use in assertj. They are not so nice to use, because you must know the methods autocompletion struggles with this. 
Spockframework locks like python :D not my favorite
https://github.com/google/truth looks a bit like assertj.

andsel:
I would move to Java 8 for release 0.10, surely version 1.0 is using Java
8, JUnit5 could be a good step  forward.

andsel:
Really Spock is Groovy ;-)  You convinced me, I'll give a try to AssertJ, if the project in future is abandoned (that's my unique trouble about it) we can switch to Google's Truth or fallback on Hamcrest. JUnit 5 at the end to use fluent assert API suggest to use Hamcrest, Truth or AssertJ

PR #264, opened 2017-01-31
Java 1.8

AndreasMager:
I will keep this PR in sync with your master branch.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,1 +1,3 @@
 language: java
+jdk:
+  - oraclejdk8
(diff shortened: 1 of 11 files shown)


windbender:
+1  on moving to java 8.   Java 9 is basically out.  Java 7 is EOL.  what are we waiting for?

andsel:
Moquette could potentially run on Android,what do you think to move to
Kotlin?it nice modern functional needs Java 1.6 and Android support it!
Il giorno 20/mag/2017 06:35, "windbender" <notifications@github.com> ha
scritto:

AndreasMager:
I have written some Android apps. There is no clean way you can run a proper server environment like moquette in Android thread and memory system. Its an anti pattern for android apps.

Kotlin has some nice features but it has also some drawbacks. A build system setup mixed with java is a pain even in gradle. I have rolled back kotlin code from our server development. The last time i tried kotlin 6 month ago the editor language support was still not good.

PR #267, opened 2017-02-09
Refactor of SubscriptionsStore.matches

AndreasMager:
Motivation:
LinkedBlockingDeque is a bad choice in this use case. It is designed for concurrent scenarios.
This implementaition is single threaded.

I have measured some implementations  the winner was a array of Token.

### DIFF
--- a/broker/src/main/java/io/moquette/spi/impl/subscriptions/SubscriptionsDirectory.java
+++ b/broker/src/main/java/io/moquette/spi/impl/subscriptions/SubscriptionsDirectory.java
@@ -21,6 +21,7 @@ import io.moquette.spi.ISubscriptionsStore.ClientTopicCouple;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import java.util.*;
+import java.util.concurrent.LinkedBlockingDeque;
 import java.util.concurrent.atomic.AtomicReference;
 
 /**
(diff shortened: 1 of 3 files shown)


andsel:
Could you share such measurements?
Deque was used because the work on stacked tokens, so the abstract data type of Deque must be maintained, we could change the implementation, for example if arrays is more performant, we can use ArrayDeque

AndreasMager:
Ok @andsel if have tested the ArrayDeque and its the fastes of all even with all the copy operations.
If it's stupid but it works it ain't stupid.

ArrayDeque vs []
deque wins with 95% less time

[] vs LinkedBlockingDeque
[] wins with 89% less time

ArrayList and LinkedList are slightly slower than [].

andsel:
Before to decide if a thing is good or bad I would like to see the test code, so that I can reproduce it on my laptop, or at least the description of the testbed plus the results in terms of number of insertion/deletion per seconds.

AndreasMager:
The test code is pref/../SubscriptionStoreTest. I have  change the code to test both implementation in the same test run.

AndreasMager:
Hi @andsel . I have added a temporary commit with both implementations.

Minimum time of 500 Testruns:
new implementation took 225ms
old implementation took 263ms

AndreasMager:
The code became removed.

PR #268, opened 2017-02-09
MessageGUID: class replaced by UUID

AndreasMager:
Motivation:
MessageGUID don't add any advantages to the project. 
Instead the are some disadvantages: 
They add more object's (higher gc load)
To store the UUID as Strings needs more memory


AndreasMager:
This Update will break old MapDB files!

andsel:
Hi @AndreasMager MessageGUID instead carry a big advantage to the project, it's a value object in the domain model, it say what it needs; I mean a UUID is generic , could be a uuid for subscription, for message or for anythiing, MessageGUID say that's a uuid for message. So The class has a big reason to stay.
We can drop the String of UUID and keep just the byte[] representing that UUID, this eliminates the 34 bytes for string chars + the reference to it, keeping just a byte[] of 16 bytes; but I wan to keep the MessageGUID

AndreasMager:
I still see no benefit only trouble.  Do you wan't create a class for each Datatype you use like class MessageCounter { int counter} or class Topic { String topic}?
With the Streams Api's from java 1.8 you will hate this stuff.

andsel:
Sorry but don't agree with you, having some Map<UUID, Something> for me it's like having Map<String, something> so I prefer to use some more meaningful concept as key. In this case I think that using a domain object carry great value to readability.

PR #274, opened 2017-02-24
The subscriptions tree must be pruned in order to prevent severe performance issues

lbarriosh:
We have detected a severe performance issue in the following scenario:
- clients connect with the cleanSession flag setted to true.
- after establishing a new connection against the broker, clients subscribe to a different MQTT topic of their own. These topics are not shared among clients.

Each time a client disconnects, its subscriptions are removed from the subscription tree. But the nodes remain on the tree with an empty subscriptions list.

When the broker processes a publish message, it traverses the whole subscription tree. The number of empty nodes grows each time a client disconnects, and this increments the time the broker takes to handle a PUBLISH message (from less than 20 ms after starting up the broker to 200 ms after several hours).

This issue also makes the broker vulnerable to very simple DDoS attacks, an can be easily fixed by modifying the SubscriptionsStore and TreeNode classes.

### DIFF
--- a/broker/src/main/java/io/moquette/connections/IConnectionsManager.java
+++ b/broker/src/main/java/io/moquette/connections/IConnectionsManager.java
@@ -10,12 +10,6 @@ import java.util.Collection;
  *
  */
 public interface IConnectionsManager {
-	/**
-	 * Returns the number of physical connections
-	 * 
-	 * @return
-	 */
-	public int getActiveConnectionsNo();
 
 	/**
 	 * Determines wether a MQTT client is connected to the broker.
--- a/broker/src/main/java/io/moquette/server/ConnectionDescriptorStore.java
+++ b/broker/src/main/java/io/moquette/server/ConnectionDescriptorStore.java
@@ -97,11 +97,6 @@ public class ConnectionDescriptorStore implements IConnectionsManager {
     }
 
     @Override
-    public int getActiveConnectionsNo() {
-        return connectionDescriptors.size();
-    }
-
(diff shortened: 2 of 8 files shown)


andsel:
Thanks @lbarriosh fro the PR, but there some things I don't understand.
How the modifications to NettyAcceptor influence the solution to the bug? if they are not part of the solution, they shouldn't be part of the PR.

Instead wrapping TreeNode with PrunedTreeNode I would have fixed directly the TreeNode, at the end PrunedTreeNode is just a bean to wrap a TreeNode and a flag, so probably is better to move that flag into TreeNode.

lbarriosh:
Hi @andsel 
> How the modifications to NettyAcceptor influence the solution to the bug? if they are not part of the solution, they shouldn't be part of the PR.

You're right. You can ignore all the changes except the ones at the TreeNode.java file. It seems that your codebase has evolved a lot since our last PR: that's where the other changes come from.

> Instead wrapping TreeNode with PrunedTreeNode I would have fixed directly the TreeNode, at the end PrunedTreeNode is just a bean to wrap a TreeNode and a flag, so probably is better to move that flag into TreeNode.

If the TreeNode instances are not serialized anywhere, the alternative you suggest is the best one. The TreeNode class is used at the core of the broker, and we didn't want to break anything up.

Regards,

Luis

andsel:
Hi @lbarriosh thanks for your suggestion, fixed this PR directly on master with e5c46e7a55591840fd4fc7eebedb9c719d5d0ba9

PR #277, opened 2017-03-08
Changes on interception package

AndreasMager:
Field and Properties moved to **AbstractInterceptHandler** for code reuse.

New intercept event **WipeSubscriptionsMessage** added. This event is needed to build a consistent state of all subscription in the broker

Topic class used in topics related intercept event messages.

Config file property **hazelcast.topic_name**  added to choose the topic name.


AndreasMager:
I won't maintain this branch any more. I switch to RxJava

PR #278, opened 2017-03-08
ScheduledExecutorService moved from MapDBPersistentStore to Server

AndreasMager:
To provide a global scheduler executor in the broker the field has been moved.

### DIFF
--- a/broker/src/main/java/io/moquette/spi/impl/ProtocolProcessorBootstrapper.java
+++ b/broker/src/main/java/io/moquette/spi/impl/ProtocolProcessorBootstrapper.java
@@ -98,22 +98,24 @@ public class ProtocolProcessorBootstrapper {
             messagesStore = store.messagesStore();
             m_sessionsStore = store.sessionsStore();
             storeShutdown = new Runnable() {
-				@Override
-				public void run() {
-					store.close();
-				}
-			};
+
+                @Override
+                public void run() {
+                    store.close();
+                }
+            };
         } else {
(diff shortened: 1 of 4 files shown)


broker/src/main/java/io/moquette/spi/impl/ProtocolProcessorBootstrapper.java
```diff
@@ -98,22 +98,24 @@ public ProtocolProcessor init(IConfig props, List<? extends InterceptHandler> em
...
-			};
+
+                @Override
+                public void run() {
+                    store.close();
+                }
+            };
         } else {
-            final MapDBPersistentStore m_mapStorage = new MapDBPersistentStore(props);
+            final MapDBPersistentStore m_mapStorage = new MapDBPersistentStore(props, server);
```
andsel: Instead of pass the server, creating a strong dependency between MapDBPersistentStore and Server, is better to catch in the ScheduledExecutorService

AndreasMager: Good point.

andsel:
Thanks @AndreasMager if you squash the commits I merge the PR.

AndreasMager:
@andsel ready for merge..

andsel:
Thanks @AndreasMager great!! I plan to use this centralised scheduler service for all sched operations inside the broker. Problably I'll extend the concet to have a central executor service for every asych operation (but this is just an idea)

AndreasMager:
With the upgrade to java 1.8 you will get ForkJoinPool.commonPool()

PR #282, opened 2017-03-13
LMDB persistent store implementation for Moquette

jan-zajic:
see
https://en.wikipedia.org/wiki/Lightning_Memory-Mapped_Database
https://symas.com/offerings/lightning-memory-mapped-database/
https://github.com/lmdbjava/lmdbjava

### DIFF
--- a/broker/pom.xml
+++ b/broker/pom.xml
@@ -82,6 +82,12 @@
         </dependency>
 
         <dependency>
+            <groupId>org.lmdbjava</groupId>
+            <artifactId>lmdbjava</artifactId>
+            <version>0.0.5</version>
+        </dependency>
+
+        <dependency>
             <groupId>org.hdrhistogram</groupId>
             <artifactId>HdrHistogram</artifactId>
             <version>2.1.4</version>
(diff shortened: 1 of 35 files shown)


andsel:
Hi @jan-zajic I'm already int he way to move the key-value storage part to H2 MV Store, like discusse in an old issue. Personally I prefere to maintain just one persistence engine, instead of plug them, at the end they are almost equivalent to each other. I move to H2 because is more solid and maintaned then MapDB 1.0. MapDB 3.x is not a viable solution because it has huge deps and I wnat to keep Moquette as slim as possible.
 So, given this motivation the PR is not beeing accepted, thanks a lot for you effort!!!
 Andrea

jan-zajic:
@andsel in that case what about make storage engine plugable via configuration like in my PR or in https://github.com/andsel/moquette/pull/269, or any other way so we can make our storage engine as aditional optional module? I can publish it to maven central and maintain it myself but you must keep IPersistentStore interface or IStore from PR 269, now MapDBPersistentStore is hardcoded in code.

jan-zajic:
btw @andsel when you be done with h2 implementation? I known h2 pretty well, so i can help if necessary...

andsel:
I'm using H2MVStore that's not the H2 database. Currently I'm grooming to think hot to implement efficiently a persistent queue.

PR #289, opened 2017-03-22
Build improvements

larroy:


larroy:
ping

andsel:
Hi, @larroy I 've some questions
- what ping means?
- you introduced logback but not configured, so you bring a dependency that's not used.
- the maven-assembly-plugin plugin is not the maven plugin to handle OSGi stuff inside maven, in which sense it generated the OSGi headers?
- I don't find useful the  broker/build.sh which skips tests during the build process, which is the intended use case?

larroy:
but logback has a default configuration. It provides a slf4j back end right?

What I wanted to do was to generate a single jar that you can use to run the broker. Is there a better way to do it?

Regarding build.sh the goal is to generate a fat jar.

Is there a better alternative to all this? I'm not super familiar with Maven...

andsel:
The spirit is that Moquette should be easily embeddable, so this means that when it's run in self mode it should be easy configurable, and minimal steps are download the tar gz, unpack and run it from bin.
Instead when it's used inside other project you have simply to add the dependency to the broker jar.
I don't love the uber Jars, where every system you integrate brings packed all the necessary dependencies, I prefer maven/gradle resolves the dependencies transitively. These improve the sharing of artifacts and lover the bandwidth requested on updates

PR #293, opened 2017-03-31
add synchronization for inFlightForClient

fbeister:
This fixes a race condition in

`int maxId = inFlightForClient.isEmpty() ? 0 : Collections.max(inFlightForClient);`

where `isEmpty()` returns `false`, then another threads removes the last element and then `max()` throws a `NoSuchElementException`

There might be better ways to accomplish this (special methods of ConcurrentMap or such, e.g. `putIfAbsent` or `computeIfAbsent`) but this fixes the problem at hand.

Apart from that, `m_inFlightIds` should also be checked for race conditions because checking if there is an entry for a clientId and creating its Set also isn't atomic.


andsel:

Issue #237, opened 2016-11-29
Potential Bug in NextPacketID


peacecoder:
Hi,

I find that there is a potential bug of getting next packet id. Here is the source code

```
    /**
     * Return the next valid packetIdentifier for the given client session.
     * */
    @Override
    public int nextPacketID(String clientID) {
        Set<Integer> inFlightForClient = this.m_inFlightIds.get(clientID);
        if (inFlightForClient == null) {
            int nextPacketId = 1;
            inFlightForClient = Collections.newSetFromMap(new ConcurrentHashMap<Integer, Boolean>());
            inFlightForClient.add(nextPacketId);
            this.m_inFlightIds.put(clientID, inFlightForClient);
            return nextPacketId;
        }

        int maxId = inFlightForClient.isEmpty() ? 0 : Collections.max(inFlightForClient);
        int nextPacketId = (maxId + 1) % 0xFFFF;
        inFlightForClient.add(nextPacketId);
        return nextPacketId;
    }
```

However according to the protocol

> SUBSCRIBE, UNSUBSCRIBE, and PUBLISH (in cases where QoS > 0) Control Packets MUST contain a non-zero 16-bit Packet Identifier [MQTT-2.3.1-1]. Each time a Client sends a new packet of one of these types it MUST assign it a currently unused Packet Identifier [MQTT-2.3.1-2]. If a Client re-sends a particular Control Packet, then it MUST use the same Packet Identifier in subsequent re-sends of that packet. The Packet Identifier becomes available for reuse after the Client has processed the corresponding acknowledgement packet. In the case of a QoS 1 PUBLISH this is the corresponding PUBACK; in the case of QoS 2 it is PUBCOMP. For SUBSCRIBE or UNSUBSCRIBE it is the corresponding SUBACK or UNSUBACK [MQTT-2.3.1-3]. The same conditions apply to a Server when it sends a PUBLISH with QoS > 0 [MQTT-2.3.1-4].

There are 3 bugs.

1. We know that packet Id range from 1 to 65535, but the code will generate from 1 ~ 65534.
2. It will get a packet id that is still in flight. since it doesn't check the next packet id is in the set or not.
3. There is only one ProtocolProcess share among muliti-thread, this will cause inconsistence, I mean it might cause each thread get the same packet id at the same time.

Tell me if i get a wrong way.

hylkevds:
This was fixed during a recent rewrite of that bit of code.

Issue #208, opened 2016-09-14
Intermittent concurrentModificationExceptions 


kjayashankar:
Hi, 
I'm stress testing moquette and I see lot of concurrentmodification exceptions, 
first it was in Collections.sort of MapDBSessionStore class, nextPacketID( ) method. I modified code base to not use collections.sort and used a method to get the max of collection using iterator, still I'm seeing intermittent exceptions like these :

```
22:16:09,503 [nioEventLoopGroup-3-2] ERROR NettyMQTTHandler channelRead 84  - Bad error in processing the message
java.util.ConcurrentModificationException
    at java.util.HashMap$HashIterator.nextNode(HashMap.java:1437)
    at java.util.HashMap$KeyIterator.next(HashMap.java:1461)
    at io.moquette.spi.persistence.MapDBSessionsStore.max(MapDBSessionsStore.java:179)
    at io.moquette.spi.persistence.MapDBSessionsStore.nextPacketID(MapDBSessionsStore.java:168)
    at io.moquette.spi.ClientSession.nextPacketId(ClientSession.java:170)
    at io.moquette.spi.impl.ProtocolProcessor.route2Subscribers(ProtocolProcessor.java:555)
    at io.moquette.spi.impl.ProtocolProcessor.processPublish(ProtocolProcessor.java:418)
    at io.moquette.server.netty.NettyMQTTHandler.channelRead(NettyMQTTHandler.java:61)
    at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:318)
    at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:304)
    at io.moquette.server.netty.metrics.MessageMetricsHandler.channelRead(MessageMetricsHandler.java:46)
    at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:318)
    at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:304)
    at io.netty.handler.codec.ByteToMessageDecoder.fireChannelRead(ByteToMessageDecoder.java:276)
    at io.netty.handler.codec.ByteToMessageDecoder.channelRead(ByteToMessageDecoder.java:263)
    at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:318)
[...]

PR #294, opened 2017-04-10
Bugfixes in package persistence

AndreasMager:


AndreasMager:
Fixes bug #208
Fixes bug #237

PR #298, opened 2017-04-20
IMessagesStore#storeRetained: this method should decide to store QOS0 messages

AndreasMager:
Motivation:
The storage implementation gets the possibility to choose how the different QOS levels for retained messages are handled.
The current implementation suppresses QOS 0 retained messages, the MessagesStore won't get them.
 
My storage implementation has different levels of reliability and speed. 
For example:
QOS 0 : persistent, fast and non reliable
QOS 1 : persistent, fast and reliable
QOS 2 : persistent, back pressure and reliable

This PR is reducing the Topic class creation.


AndreasMager:
Closes: #304

windbender (commented):
Looking forward to the QOS0 fix!

AndreasMager:
This PR contains no fix. I am sorry i added your issue number to this but it was the wrong PR. The deletion of the link comment didn't have a effect to the linking.

PR #299, opened 2017-04-21
Protocol processor added db type field

AndreasMager:
If a global database is used which store retain messages, doesn't need to restore them in ProtocolProcessor#internalPublish .


andsel:
It's not clear to me what you intend for "global database"

AndreasMager:
I have a cassandra cluster which stores the retain messages. The hazelcast implementation redistributes the publish messages and stores them again.

PR #300, opened 2017-04-21
PersistentQueueMessageSender#sendPublish: warning should not be loged in OK case

AndreasMager:


AndreasMager:
Closes: #304

AndreasMager:
implement in https://github.com/andsel/moquette/commit/23b903e2f282463057862d9288b1d0b6467db5a5

PR #301, opened 2017-04-25
Missing message logging added

AndreasMager:


andsel:
Message type must be uppercase for grepping in logs, I don't see any added value in using the type.toString()

AndreasMager:
MqttMessageType enum is defined in uppercase. The first commit should be merged. There a lot of errors in the log at the moment.

PR #321, opened 2017-05-31
MemorySessionStore: Session class added

AndreasMager:
A new class added for easier session cleanup and LRU Cache etc.


andsel:
Hi @AndreasMager  could you explain/comment a little bit about this PR?

AndreasMager:
The whole data structure of the a session is broken splitted over several Map in Maps because of the old MapDB implementation. The MemoryStorage implementation doesn't need this burden.

The Session cleanup should be done by the Storage implementation. Your new commit goes in this way with dropInFlightMessagesInSession Method. 

The MemoryStorage needs a method to cleanup old outdated session like a LRU Cache to avoid OutOfMemoryExceptions. 

I want to implement the guava CacheBuilder for this task.

andsel:
The PR seems good to me, I agree with you. Once you squash all the commits I merge it. Thanks

AndreasMager:
Fixes #325

AndreasMager:
@andsel rebased and squashed

Issue #310, opened 2017-05-17
ACL for topic that contains #


pedrocarvalho:
Hi,

I'm getting a ParseException when using a topic that contains a # character. Example:

```
user someUser
topic readwrite mytopic/#
```

Looking at source in ACLFileParser.java seems obvious what the error is. As it is, if the line contains a # in other position than the beginning  of the line, the ParseException is thrown.

```
       while ((line = br.readLine()) != null) {
                int commentMarker = line.indexOf('#');
                if (commentMarker != -1) {
                    if (commentMarker == 0) {
                        // skip its a comment
                        continue;
                    } else {
                        // it's a malformed comment
                        throw new ParseException(line, commentMarker);
                    }
                } else {
                    if (line.isEmpty() || line.matches("^\\s*$")) {
                        // skip it's a black line
                        continue;
                    }

                    collector.parse(line);
                }
            }
```

andsel:
Fixed by PR #340

PR #340, opened 2017-07-25
#310 - Allow for topic/patterns with '#' and end of line comments

nMoncho:
Solves #310

### DIFF
--- a/broker/src/test/java/io/moquette/spi/impl/security/ACLFileParserTest.java
+++ b/broker/src/test/java/io/moquette/spi/impl/security/ACLFileParserTest.java
@@ -44,9 +44,12 @@ public class ACLFileParserTest {
     }
 
     @Test(expected = ParseException.class)
-    public void testParseInvalidComment() throws ParseException {
+    public void testParseInvalidPaddedComment() throws ParseException {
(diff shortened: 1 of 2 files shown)


broker/src/test/java/io/moquette/spi/impl/security/ACLFileParserTest.java
```diff
@@ -43,10 +43,13 @@ public void testParseValidComment() throws ParseException {
         assertTrue(authorizations.isEmpty());
     }
 
-    @Test(expected = ParseException.class)
```
andsel: Comment lines (like in Mosquitto conf) must start with # this test verifies that if there a space a before the file is invalid.

andsel: This invalidate the PR

nMoncho: I reverted to invalidate comments lines that start with a space. Thought that would be more fault tolerant if comment lines starting with spaces would be treated the same as without them.

andsel:
Thanks @nMoncho merged
 Andrea

Issue #171, opened 2016-04-29
Last Will is never treated as retained


knr1:
We noticed that moquette does not seem to retain the last will published by moquette, even though the message has been prepared this way during connection.

However, according to the specification OASIS 3.1.1, 
"If the Will Flag is set to 1, the Will QoS and Will Retain fields in the Connect Flags will be used by the Server, and the Will Topic and Will Message fields MUST be present in the payload [MQTT-3.1.2-9]."

In order to provoke this behaviour, we have done the folowing:
Create a connection with a last will: 
(clientid "a", qos "1", retain "true", topic "status", message "offline")
when the connection is disconnected forcefully (by killing the process), the broker sends the will message.
We checked this with a client via WebSocket: (clientId "b", subscribe "#", qos "1")

But when connecting and checking with a client c some moments after client b has confirmed the last will, the message is not retained, hence cannot be seen by c (or any other client) anymore....

Could you elaborate on that?

andsel:
Hi @knr1 you are right and Moquette implement it. I've just added an integration test to check your use case (commit: dbe9c7bcaa179a4c4b5ca0633862b12e09f1a6eb) could you provide the log of your real case? (also witch version on Moquette you are using, if snapshot, try to update it, which OS, possibly a unit test or code to make it repeatable).

knr1:
Sorry for the long absence... Here I am again.
For me, the situation is still the same even with the newest checkout of the 0.9 broker.

I started the broker... here is its printout:

---

Building Moquette - broker 0.9-SNAPSHOT

---

Downloading: https://repo.eclipse.org/content/repositories/paho-releases/io/moquette/moquette-netty-parser/0.9-SNAPSHOT/maven-metadata.xml
Downloading: https://oss.sonatype.org/content/repositories/snapshots/io/moquette/moquette-netty-parser/0.9-SNAPSHOT/maven-metadata.xml

Downloading: https://repo.eclipse.org/content/repositories/paho-releases/io/moquette/moquette-parent/0.9-SNAPSHOT/maven-metadata.xml
Downloading: https://oss.sonatype.org/content/repositories/snapshots/io/moquette/moquette-parent/0.9-SNAPSHOT/maven-metadata.xml

--- exec-maven-plugin:1.2.1:exec (default-cli) @ moquette-broker ---
log4j:WARN No appenders could be found for logger (io.moquette.server.Server).
log4j:WARN Please initialize the log4j system properly.
[...]

PR #343, opened 2017-08-05
Fix issue #171

Dolychuck:
As per issue #171 using a similar example:  upon connecting to the broker a client can connect with a retained message status of "online" and for unexpected disconnection a retained will message of "offline". However, currently the will message is not being retained resulting in "online" being retained to the topic after the will message has been sent.

### DIFF
--- a/broker/src/main/java/io/moquette/spi/impl/ProtocolProcessor.java
+++ b/broker/src/main/java/io/moquette/spi/impl/ProtocolProcessor.java
@@ -561,15 +561,20 @@ public class ProtocolProcessor {
     /**
      * Specialized version to publish will testament message.
      */
-    private void forwardPublishWill(WillMessage will, String clientID) {
-        LOG.info("Publishing will message. CId={}, topic={}", clientID, will.getTopic());
-        // it has just to publish the message downstream to the subscribers
-        // NB it's a will publish, it needs a PacketIdentifier for this conn, default to 1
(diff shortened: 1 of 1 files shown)


andsel:
Sorry but for me there is a bit confusion. which is the workflow you are describing?
- a client connect to the broker with "will retains" flag set to true
- the will message has the payload "offline"
- the client disconnects abruptly form the broker
- the broker publish the "offline" message as retained

which is the expected behaviour (from spec of MQTT 3.1.1) and how the broker violate it?
Without a clear understanding of the bug and solution I can't accept this PR

Dolychuck:
Sorry for the confusion I didn't explain myself very well earlier. Perhaps there is a mistake in the logic of what I am trying to do. The intended workflow is as follows: 

   1) A client connects to the broker with "will retains" flag set to true to the topic of "status"
   2) The will message has the payload "offline"
   3) Immediately after connecting to the broker the client sends a message with payload "online" to the 
   topic "status" set as a retained.
   4) The client disconnects abruptly from the broker.
   5) The broker publishes the "offline" message as retained. (overwriting "online") 
   6) The topic "status" retains "offline"

I have verified that the will message is being sent, but when a new client subscribes to the topic the message being received is "online". 
My assumption is that the LWT message would be the last message sent to the topic and it should therefore be the message received by any clients who subscribe to the topic after this has taken place.

andsel:
Your fix seems to solve the bug, thanks I'll cover with a test

PR #344, opened 2017-08-06
Allow custom server path for MQTT over websockets

jeremymailen:
Some environments may require a different route for a broker.
https://wiki.eclipse.org/Paho/Paho_Websockets suggests that /mqtt should be the
default with an option to specify an alternative.

This change also turns on subpath matching in case a user wants to configure "/" to handle all paths.


andsel:
Sounds good, thanks

jeremymailen:
Did you mean to merge this or did you plan to reimplement?

Issue #347, opened 2017-08-17
memory leak on MemorySessionStore


arifainchtein:
Hello,
I am embedding  Moquette using the following code:

`IResourceLoader classpathLoader = new ClasspathResourceLoader();
 	        final IConfig classPathConfig = new ResourceLoaderConfig(classpathLoader);

 	        final Server mqttBroker = new Server();
 	       
 	        List<? extends InterceptHandler> userHandlers = Collections.singletonList(new PublisherListener());
 	        mqttBroker.startServer(classPathConfig, userHandlers);`

The class PublisherListener is

`public class PublisherListener  extends AbstractInterceptHandler {
	Logger logger;
	
	public PublisherListener() {
		logger = Logger.getLogger(getClass());
		
	}
	public void onConnect(InterceptConnectMessage message) {
		logger.info(" received connection from " +  message.getClientID());
	}
	 
	    public void onDisconnect(InterceptDisconnectMessage msg) {
	    	logger.info(" received onDisconnect from " +  msg.getClientID());
	    } 

	    public void onConnectionLost(InterceptConnectionLostMessage msg) {
	    	logger.info("onConnectionLost from " +  msg.getClientID());
	    }
	    
	public void onPublish(InterceptPublishMessage message) {
		logger.info(" received a message on topic: " + message.getTopicName()
		+ ", from: " + message.getClientID());
				
	}
	@Override
	public String getID() {
		// TODO Auto-generated method stub
		return "Publisher Listener";
	}
}
`
after the program runs for a while, it crashes because of an out of memory error.
I ran the code with the flag to produce and memory dump which generated an hprof file.  When I opened it in eclipse, and ask for the memory leak report, i got the following:

**One instance of "io.moquette.persistence.MemorySessionStore" loaded by "sun.misc.Launcher$AppClassLoader @ 0x6eaae6f8" occupies 125,642,208 (97.21%) bytes. The memory is accumulated in one instance of "java.util.concurrent.ConcurrentHashMap$Node[]" loaded by "<system class loader>".**

Not sure what object is not garbae collected.  Any suggestion would be appreciated.

andsel:
I've reworked a lot the session management with commit 1b272e584933fc0b78919bad767d7e73dbcc01b6 so this should be attenuated

PR #348, opened 2017-08-18
MemorySessionStore configurable Cache for Session added

AndreasMager:
In the `MemorySessionStore` the sessions doesn't get deleted. 
If clients uses for each connection a random ClientID the memory leaks away.

This PR contains Guava Cache that will start to drop old Session. 
The eviction happens by time and size

It adds 2 new configuration properties:
`storage_class.max_sessions` The default value is 5000 session
If 5000 session is reached the old session will get dropped

`storage_class.session_expire` The default value is 28 days
If a session isn't used for 28 days it will become dropped

Fixes: #347

### DIFF
--- a/broker/src/main/java/io/moquette/BrokerConstants.java
+++ b/broker/src/main/java/io/moquette/BrokerConstants.java
@@ -67,6 +67,9 @@ public final class BrokerConstants {
 
     public static final String STORAGE_CLASS_NAME = "storage_class";
 
+    public static final String STORAGE_CLASS_MAX_SESSIONS = "storage_class.max_sessions";
+    public static final String STORAGE_CLASS_SESSION_EXPIRE = "storage_class.session_expire";
+
     private BrokerConstants() {
     }
 }
--- a/broker/src/main/java/io/moquette/persistence/MemorySessionStore.java
+++ b/broker/src/main/java/io/moquette/persistence/MemorySessionStore.java
@@ -16,7 +16,9 @@
 
 package io.moquette.persistence;
 
+import io.moquette.BrokerConstants;
 import io.moquette.server.Constants;
+import io.moquette.server.config.IConfig;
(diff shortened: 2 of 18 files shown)


andsel:
Hi @AndreasMager 
Do we really need google guava? One of the principle of Moquette it to have minimal deps as possible, and google guava brings in a lot of code.

The problem of session expiration is due to a protocol deficiency, and solving only for Memory implementation is not good.

What I do is to add days as a field in the Session, and than having a cleaner task on a schedule base. Than at every new session creation in the SessionStore verify if excedes the capacity and than do the clean in case. 
As final point, hearbeat name of a method is not so meaningful, probable a "isPresent" or "stillSessionValid" are better choices

AndreasMager:
I have used Guava because its already in the dependencie tree. I can switch to an own implementation if want it? 
What do you think a about "updateValidity" for the method name.
I will add an implementation for both other Storage systems.

andsel:
Guava is already included?! Not on the current trunk, there's no reference
in the pom.xml, is it transitive?
But the method actually doesn't update anything, just check if the session
is still present in the cache, renaming to updateValidity and just doing a
check is little bit misunderstanding for me.

I think that changes like this, that essentially are structural has to be
discussed in the mailing list, to keep track and share the thoughts that
drove to certain decisions.

AndreasMager:
The guava dependencie comes transitive with bugsnag 
+--- com.bugsnag:bugsnag:[3.0,4.0) -> 3.0.2
|    +--- com.fasterxml.jackson.core:jackson-databind:2.6.7 -> 2.8.8 (*)
|    +--- com.google.guava:guava:12.0.1

This checking in the method raises the access counter of Cache. So it doesn't become deleted.
Is google group the mailing list?

andsel:
:-( nice bugsnag!
yes it's the google group

PR #351, opened 2017-08-25
StandartCharset used

AndreasMager:
DefaultCharset - errorprone warnings removed.


andsel:
It's really needed ? As I know by default getBytes() return the bytes in UFT-8 when you don't specify the charset

AndreasMager:
The charset thing is a real mess. I don't like it too. The default charset depends on the OS and the jvm startup parameter. It should be UTF-8 and it will be cached. I hate ́`SHOULD BE` in rfc and code.
If you don't like this PR `-Dfile.encoding="UTF-8"` should be added to the startup script.

http://javarevisited.blogspot.de/2012/01/get-set-default-character-encoding.html

andsel:
Hi Andreas,
good catch, I prefer the -Dfile.encoding="UTF-8"on the .sh and .bat scripts

 Andrea

PR #364, opened 2017-10-21
CTrie TNode fix, concurrency test fixes, client_id locking

mwilliford:
Please see the attached commits for consideration into master.  In my tests, these changes have improved the stability of the new CTrie implementation under a bit of controlled load.  Discussion encouraged.

### DIFF
--- a/broker/src/main/java/io/moquette/connections/MqttSession.java
+++ b/broker/src/main/java/io/moquette/connections/MqttSession.java
@@ -31,6 +31,7 @@ public class MqttSession {
     private int secondPhaseAckPendingMessages;
     private Collection<MqttSubscription> activeSubscriptions;
     private MqttConnectionMetrics connectionMetrics;
+    private String clientID;
 
     public boolean isConnectionEstablished() {
         return connectionEstablished;
@@ -88,4 +89,9 @@ public class MqttSession {
         this.connectionMetrics = connectionMetrics;
     }
 
+    public String getClientID() { return clientID; }
+
+    public void setClientID(String clientID) {
+        this.clientID = clientID;
+    }
 }
--- a/broker/src/main/java/io/moquette/server/ConnectionDescriptorStore.java
+++ b/broker/src/main/java/io/moquette/server/ConnectionDescriptorStore.java
@@ -129,7 +129,7 @@ public class ConnectionDescriptorStore implements IConnectionsManager {
     public Collection<MqttSession> getSessions() {
         LOG.info("Retrieving status of all sessions.");
         Collection<MqttSession> result = new ArrayList<>();
-        for (ClientSession session : this.sessionsRepository.getAllSessions()) {
+        for (ClientSession session : this.sessionsRepository.getAllSessions(true)) {
             result.add(buildMqttSession(session));
         }
         return result;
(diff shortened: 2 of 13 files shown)


andsel:
Correct if I'm wrong, but substantially you are suggesting to use a locking per client? If it's, it's not what I wanted to implement using the CTrie, as CTrie original paper state, they are trees structures thought to avoid locking by VM and using versioning and CAS operations.

mwilliford:
I should have been more clear about the PR.  There are 4 separate changes, each which address a different problem.  I can break this PR into 4 PR's if you would like, so we can discuss each one separately.  I do understand that you have a goal of being as lock-free as possible.  Let me know how to proceed.

First commit - CTrie fix 09ab912c08f656370a01b3ae8bcd27bc03e33276
Our use of the CTrie deviates from the paper in a few ways, some of which are a bug, some which are ok for now.  
We allow for a node to exist without any subscriptions, and prior to my recommended change (this commit), we were not compressing TNodes.  This commit compresses just the last TNode after unsubscribe.  This is essential because you can never reuse an INode which points to a TNode, it is tombed forever.  So the next insert would need to handle this condition prior to insert.  
There was another issue with concurrency inside the CNode instance itself.  Concurrent access to a single CNode is typical under load, and I did in fact hit concurrency exceptions with subscriptions and clients datastructures.  I make the copy constructor generate new datastructures, I think the code was only passing on the reference to the new CNode before.  We may need to look further into when these concurrency conditions occur, but I am 100% sure they do.
The There are a few remaining improvements that we should probably consider in the future:  Our remove only removes the last node, so if you have a long chain, we are left with unused nodes.  I think the paper also suggests that we check for TNodes in all operations, and handle them prior to any insert, update, or delete.
 
If you run the unit tests from this commit (without the fixes), you should notice that clients get into a situation where they cannot remove a subscription.  Most other bugs were observed in testing under heavy load.


Commit:  77759768ec4d5714c220a30d66633fb46c3eb6aa
This is a feature we need, I'd like to be able to get all sessions, including transient sessions.  Also, added the client_id to the DTO. 

Commit:  3608c76b100618df12cffb89edfbb53804aa0636
Do not allow simultaneous mqtt messages from a given client to enter.  There are a few ideas to discuss here.  First, a session seems like it should have affinity on a NIO channel, or rather TCP socket.  I notice now that you store the client_id as a Netty channel atrib.  So, under normal conditions, the locking should never restrict, would you agree?  I think MQTT restricts a client_id to a single connection at a time anyways.  In practice, under heavy load, with clients attempting to reconnect, I believe that there are possible conditions which a client may attempt to violate these rules.  I see that there are places in publish/unpublish or maybe login, which are potentially not safe to run concurrently for the same client_id.  Since it should not happen anyways, I thought this would be a stability improvement to enforce.  Let me know what you think about this idea.

f4c53822c3cf5ae1ada74842d97cf2aa34a2ee56
Try/Catch around canRead or CanWrite.
If the implementation (embedded use-case) throws an exception, we should probably catch it and deny access.  If we allow the exception to be uncaught all the way to NettyMQTTHandler, then we may leave subscriptionInCourse state stuck in verified, and and at the very least we don't return a failed SUBACK.  This is not moquette's bug, but the implementations of IAuthorizator might be careless and this is a way for us to reduce bugs on their side.

mwilliford:
Probably it is best to close this PR, and let me know which commits to bundle together into a new PR.  For example, I could create a PR for all changes except the client level locking.

andsel:
I think that each commit request a PR, because it's almost unrelated.. But give me some time to review carefully.

PR #365, opened 2017-10-21
Fix for CTrie TNode exception and other improvements to CTrie related

mwilliford:
subscribe and unsubscribe issues

-  Added unit tests to cover a few more cases which I my system ran into
   under concurrent load testing.
-  Remove and detect TNode, implement more of
   http://bravenewgeek.com/breaking-and-entering-lose-the-lock-while-embracing-concurrency/
   related to TNode state
-  NOTE:  The graph is left with nodes which have no subscriptions, the
   TNode cleanup only cleanss the last node for now.  We should
   autoprune further up by placing more TNodes and pruning them off.
-  Adding concurrency Sets and Lists to CNode, because those ARE edited
   and read in parallel in our use-case.  Before a CNode is CAS
   replaced, more than 1 thread may be attempting to modify and read the
   subscriptions and children CNode datastructures, and I hit this
   during testing.
-  Fixed an issue where we unsubscribe all sub-topics if an unsubscribe
   is sent on a parent node.
-  Fixed an issue with unsubscribing multiple times when no subscrition
   exists, we should handle this gracefully.

### DIFF
--- a/broker/src/main/java/io/moquette/spi/impl/subscriptions/CNode.java
+++ b/broker/src/main/java/io/moquette/spi/impl/subscriptions/CNode.java
@@ -24,9 +24,7 @@ class CNode {
 
     Token token;
     private List<INode> children;
-    Set<Subscription> subscriptions = new HashSet<>();
-
-    //private int subtreeSubscriptions;
+    Set<Subscription> subscriptions;
 
     CNode() {
         this.children = new ArrayList<>();
@@ -35,9 +33,9 @@ class CNode {
 
     //Copy constructor
     private CNode(Token token, List<INode> children, Set<Subscription> subscriptions) {
-        this.token = token;
-        this.children = children;
-        this.subscriptions = subscriptions;
+        this.token = token; // keep reference, root comparison in directory logic relies on it for now.
+        this.subscriptions = new HashSet<>(subscriptions);
+        this.children = new ArrayList<>(children);
     }
 
     boolean anyChildrenMatch(Token token) {
@@ -75,6 +73,9 @@ class CNode {
     public void add(INode newINode) {
         this.children.add(newINode);
     }
+    public void remove(INode node) {
+        this.children.remove(node);
+    }
 
     CNode addSubscription(String clientId, Topic topic) {
         this.subscriptions.add(new Subscription(clientId, topic));
@@ -83,6 +84,7 @@ class CNode {
 
     /**
      * @return true iff the subscriptions contained in this node are owned by clientId
+     *   AND at least one subscription is actually present for that clientId
      * */
     boolean containsOnly(String clientId) {
         for (Subscription sub : this.subscriptions) {
@@ -90,7 +92,7 @@ class CNode {
                 return false;
             }
         }
-        return true;
+        return !this.subscriptions.isEmpty();
     }
 
     //TODO this is equivalent to negate(containsOnly(clientId))
--- a/broker/src/main/java/io/moquette/spi/impl/subscriptions/CTrieSubscriptionDirectory.java
+++ b/broker/src/main/java/io/moquette/spi/impl/subscriptions/CTrieSubscriptionDirectory.java
@@ -31,6 +31,7 @@ public class CTrieSubscriptionDirectory implements ISubscriptionsDirectory {
     private static final Logger LOG = LoggerFactory.getLogger(CTrieSubscriptionDirectory.class);
 
     private static final Token ROOT = new Token("root");
+    private static final INode NO_PARENT = null;
 
     INode root;
     private volatile SessionsRepository sessionsRepository;
@@ -49,10 +50,13 @@ public class CTrieSubscriptionDirectory implements ISubscriptionsDirectory {
         @Override
         public void visit(CNode node, int deep) {
             String indentTabs = indentTabs(deep);
-            s += indentTabs + (node.token == null ? "" : node.token.toString()) + prettySubscriptions(node) + "\n";
+            s += indentTabs + (node.token == null ? "''" : node.token.toString()) + prettySubscriptions(node) + "\n";
         }
 
         private String prettySubscriptions(CNode node) {
+            if (node instanceof TNode) {
+                return "TNode";
+            }
             if (node.subscriptions.isEmpty()) {
                 return StringUtil.EMPTY_STRING;
             }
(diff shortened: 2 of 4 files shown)


mwilliford:
New pull request with only the CTrie fixes, more PR's to come later.  Broke it up to allow for easier discussions.  

Marcus

broker/src/main/java/io/moquette/spi/impl/subscriptions/CNode.java
```diff
@@ -75,6 +73,7 @@ CNode copy() {
     public void add(INode newINode) {
         this.children.add(newINode);
     }
+    public void remove(INode node) { this.children.remove(node); }
```
andsel: Please rewrite with newlines

public void remove(INode node) { 
    this.children.remove(node); 
}

broker/src/main/java/io/moquette/spi/impl/subscriptions/CTrieSubscriptionDirectory.java
```diff
@@ -205,7 +226,8 @@ public void add(Subscription newSubscription) {
     }
 
     private Action insert(String clientId, Topic topic, final INode inode, Topic fullpath) {
-        Token token = topic.headToken();
+
```
andsel: Please, avoid spacing on the first lines of methods

broker/src/main/java/io/moquette/spi/impl/subscriptions/CTrieSubscriptionDirectory.java
```diff
@@ -259,25 +281,43 @@ private INode createPathRec(String clientId, Topic topic, Topic fullpath) {
...
+     *
+     *
+     * @param topic
+     * @param clientID
+     */
     public void removeSubscription(Topic topic, String clientID) {
         Action res;
         do {
-            res = remove(clientID, topic, this.root);
+            res = remove(clientID, topic, this.root, null);
```
andsel: If we can, we should not use null as special values, prefer a constant named NO_PARENT

broker/src/test/java/io/moquette/spi/impl/subscriptions/CTrieSubscriptionDirectoryTest.java
```diff
@@ -122,6 +128,8 @@ static Subscription clientSubOnTopic(String clientID, String topicName) {
 
     @Test
     public void givenTreeWithSomeNodeWhenRemoveContainedSubscriptionThenNodeIsUpdated() {
+        System.out.println(sut.dumpTree());
```
andsel: No output in tests, if we can, they are just debug strings, if we need prefer logging with trace levels

broker/src/test/java/io/moquette/spi/impl/subscriptions/CTrieSubscriptionDirectoryTest.java
```diff
@@ -130,8 +138,127 @@ public void givenTreeWithSomeNodeWhenRemoveContainedSubscriptionThenNodeIsUpdate
...
+
+    }
+
+    @Test
+    public void givenTreeWithSomeNodeWhenRemoveMultipleTimes() {
+        System.out.println(sut.dumpTree());
+
+        sut.add(clientSubOnTopic("TempSensor1", "test"));
+        System.out.println(sut.dumpTree());
+
```
andsel: no double newlines n body methods

broker/src/test/java/io/moquette/spi/impl/subscriptions/CTrieSubscriptionDirectoryTest.java
```diff
@@ -130,8 +138,127 @@ public void givenTreeWithSomeNodeWhenRemoveContainedSubscriptionThenNodeIsUpdate
...
+        sut.removeSubscription(asTopic("test"), "TempSensor1");
+        sut.removeSubscription(asTopic("test"), "TempSensor1");
+        sut.removeSubscription(asTopic("test"), "TempSensor1");
+        sut.removeSubscription(asTopic("test"), "TempSensor1");
+
+
+        //Verify
+        final Optional<CNode> matchedNode = sut.lookup(asTopic("/temp"));
+        assertFalse("Node on path /temp can't be present", matchedNode.isPresent());
+
```
andsel: No start or end methods with white lines

mwilliford:
fixed formatting, fixed another bug, created test for this bug.  The bug was around sub-topic remove.  I believe if you belong to both /temp and /temp/1, that removing subscription from /temp should still keep you subscribed to /temp/1.  And obviously that is the case if clientID's are different.  Test cases added.

I am trying to come up with a better methodology on testing concurrency in the CTrie.  I may write some test code where I insert delays in the middle of CNode updates, and also in between adding a TNode and clearing it.  I don't have time to do this deeper investigation now, for now I just run a simulated production load and watch for any exceptions (and test that I receive expected messages).  With the production test, I was hitting that original TNode exception, with this PR I am not. 

When I have more time (TODO):
- Simulate concurrency by injecting delay or code to specifically freeze CTrie state in the middle of an update, start another operation in parallel.
- Refactor CTrieDirectory.remove with more clear if/then/else branching or case, so it is more readable.  I don't like when if conditions start to become unclear.

andsel:
Hi Marcus,
probably there are some library to do similar stuff, but it's clean to
insert sleeps, only for test.

When you speak about production traffic, could you share some data?
- clients connected
- subscriptions used
- messages per second

It's just to understand which are the real performances of Moquette,
because I don't have a such a real world use case.

 Thanks
  Andrea

mwilliford:
about 3500 concurrent devices
I think around 6 subscriptions per device
I need to instrument better on the messages per second.

andsel:
That' s a nice use case :-D do you have link to service you are developing
or it's a custom private project?

 Andrea

On Mon, Oct 23, 2017 at 10:09 AM, Marcus Williford <notifications@github.com

mwilliford:
Andrea,

Industrial AgTech lighting, we control lights.  :)

Let me know what you think of the CTrie changes.  I have more PR's behind this one to discuss.  Also, I want to make sure that I understood mqtt correctly, with respect to unsubscribing sub-topics should not unsubscribe you from deeper topic, of which we had a bug and I created unit tests for.

andsel:
Hi Marcus,
for me the PR is OK, if squash all commits I'll accept it.
Thanks
 Andrea

mwilliford:
Ok, all squashed.  thanks!  Standby for additional PR's after this one.  :)

Issue #316, opened 2017-05-19
if unchecked exception is throw while running processSubscription then remove() will never be called


windbender:
for example if   m_authorizator.canRead()  throws an unchecked exception,  then this will bubble up to processSubscription and the execution pathway will never reach 
```
boolean success = this.subscriptionInCourse.remove(executionKey, SubscriptionState.STORED);
```
Hence the broker would never be able to handle another subscription event from that executionKey again.

PR #368, opened 2017-10-29
Added try/catch blocks around calls to canRead and canWrite to protect subscription state.

mwilliford:
Addresses issue #316  

Wrap canRead and canWrite, assume false if any uncaught exceptions occur.  This prevents a situation where a client_id can never subscribe again, even on new connections.

### DIFF
--- a/broker/src/main/java/io/moquette/spi/impl/QosPublishHandler.java
+++ b/broker/src/main/java/io/moquette/spi/impl/QosPublishHandler.java
@@ -36,10 +36,19 @@ abstract class QosPublishHandler {
     public boolean checkWriteOnTopic(Topic topic, Channel channel) {
         String clientID = NettyUtils.clientID(channel);
         String username = NettyUtils.userName(channel);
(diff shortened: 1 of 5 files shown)


broker/src/main/java/io/moquette/spi/impl/QosPublishHandler.java
```diff
@@ -36,10 +36,19 @@ protected QosPublishHandler(IAuthorizator m_authorizator) {
...
+
+        try {
+            canWrite = m_authorizator.canWrite(topic, username, clientID);
+        } catch (Exception e)  {
+            LOG.error("Caught exception from canWrite, please wrap and return false instead throwing", e);
+        }
+
+        if (!canWrite) {
             LOG.error("MQTT client is not authorized to publish on topic. CId={}, topic={}", clientID, topic);
-            return true;
```
mwilliford, line 41: Even though I'm not sure this is called, please review these changes to the return value.

andsel:

PR #373, opened 2017-11-07
[Feature Request] Intercepting qos1/qos2/internal messages published to subscribers

ScienJus:
Hi andsel. i'm trying to use moquette broker with an embedded form in my web application, and want to use `InterceptHandler` to trace every published message (trace message posted by the client and message received by the client, and so forth).

In `InterceptHandler#onPublish` I can do the above, but I have no idea which clients the message published to. So I add a new method `onPublishedToSubscriber` which will be called on qos1/qos2/internal messages published in `BrokerInterceptor#notifyPublishedToSubscriber`.

How do you think of this implementation? Any better suggestions? Thanks.


andsel:

PR #379, opened 2017-11-22
feature: upgrade the test case of EmbeddedLauncher.

ningg:
Fixes https://github.com/andsel/moquette/issues/341

* feature: use `ByteBuf.getBytes()` instead of `ByteBuf.array()`, because `ByteBuf.array()` will append multi blanks to the tail.
* feature: upgrade the test case of EmbeddedLauncher.


andsel:
The commit on the EmbeddedBroker for me doesn't bring so much value, the objective of the test was just to show how it could be embedded, the internalPublish was the simplest way to show how, from  an embedder perspective, a message could be sent. Using an EmbeddedChannel (which is usable only for tests and not in prod applications) is useless in this case.

PR #395, opened 2018-03-18
To fix #366，using ConnectionDescriptor instead of clientID as the key of will msg

### LINKED ISSUE #366: Will Message lost if client reconnect before server clear the old connection
### Expected behavior
If client detected a broken connection before server， and do the reconnection, the `ProtocolProcessor#processConnect(Channel, MqttConnectMessage)` force close the old one, then accepted the new connection.
We expected that all event of old connection happened before the new connection, such as the will message, the `Interceptor#notifyClientConnectionLost()`, `Interceptor#notifyClientConnected()`
### Actual behavior
`ProtocolProcessor#processConnect(Channel, MqttConnectMessage)`  aborted the old connection, but it's asynchronous. 
The `storeWillMessage()` replaced the will message of old connection by the new one from new connection.
Then `processConnectionLost(String, Channel)` of old connection happened, mistakenly publish and cleared the will messge of new connection. 
So, **the old will message is replaced by the new one, and the new will message is lost.**

fjcyue:
To fix #366 (Will Message lost if client reconnect before server clear the old connection) :

1. using ConnectionDescriptor instead of clientID as the key of will msg store, 
2. and force fire the old will msg when the client reconnect with the same clientID.

Change-Id: Ib79c05668699a1824879429bb3fac73b903c488e

### DIFF
--- a/broker/src/main/java/io/moquette/spi/impl/ProtocolProcessor.java
+++ b/broker/src/main/java/io/moquette/spi/impl/ProtocolProcessor.java
@@ -124,7 +124,7 @@ public class ProtocolProcessor {
     SessionsRepository sessionsRepository;
 
     // maps clientID to Will testament, if specified on CONNECT
-    private ConcurrentMap<String, WillMessage> m_willStore = new ConcurrentHashMap<>();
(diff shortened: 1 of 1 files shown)


andsel:

PR #414, opened 2018-06-21
enable the usage of Netty OpenSSL provider implementation and PKCS12 key stores

mscheibler:
In order to enable the usage of additional TLS cyphers which are not supported by the JDK SSL implementation it would be nice to have the option to use the Netty OpenSSL provider implementation.

Also the use of PKCS12 key stores is now supported.

The defaults are set to match the previous behavior.

### DIFF
--- a/broker/src/main/java/io/moquette/server/DefaultMoquetteSslContextCreator.java
+++ b/broker/src/main/java/io/moquette/server/DefaultMoquetteSslContextCreator.java
@@ -16,19 +16,34 @@
 
 package io.moquette.server;
 
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.URL;
+import java.security.GeneralSecurityException;
+import java.security.KeyManagementException;
+import java.security.KeyStore;
+import java.security.KeyStoreException;
+import java.security.NoSuchAlgorithmException;
+import java.security.PrivateKey;
+import java.security.cert.Certificate;
+import java.security.cert.X509Certificate;
+import java.util.Collections;
+import java.util.Objects;
+
 import io.moquette.BrokerConstants;
 import io.moquette.server.config.IConfig;
 import io.moquette.spi.security.ISslContextCreator;
+import io.netty.handler.ssl.ClientAuth;
+import io.netty.handler.ssl.SslContext;
+import io.netty.handler.ssl.SslContextBuilder;
+import io.netty.handler.ssl.SslProvider;
+import javax.net.ssl.KeyManagerFactory;
+import javax.net.ssl.TrustManagerFactory;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import javax.net.ssl.KeyManagerFactory;
-import javax.net.ssl.SSLContext;
-import javax.net.ssl.TrustManager;
-import javax.net.ssl.TrustManagerFactory;
-import java.io.*;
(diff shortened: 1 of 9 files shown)


mscheibler:
Hi @andsel 
If you could spare a few minutes it would be good to know whether my change to ISslContextCreator gets your approval. If not, then i would have to think of another approach.
Thanks a lot.

andsel:
Hi Mike, I've just to find the time to review it.

broker/src/main/java/io/moquette/server/DefaultMoquetteSslContextCreator.java
```diff
@@ -38,77 +53,120 @@
...
         if (jksPath == null || jksPath.isEmpty()) {
-            // key_store_password or key_manager_password are empty
             LOG.warn("The keystore path is null or empty. The SSL context won't be initialized.");
             return null;
         }
-
-        // if we have the port also the jks then keyStorePassword and keyManagerPassword
-        // has to be defined
         final String keyStorePassword = props.getProperty(BrokerConstants.KEY_STORE_PASSWORD_PROPERTY_NAME);
-        final String keyManagerPassword = props.getProperty(BrokerConstants.KEY_MANAGER_PASSWORD_PROPERTY_NAME);
```
andsel, line 61: Why the key manager password has been removed,it's not anymore necessary?

broker/src/main/java/io/moquette/server/netty/NettyAcceptor.java
```diff
@@ -86,9 +101,9 @@ protected void encode(ChannelHandlerContext chc, ByteBuf bb, List<Object> out) t
         }
     }
 
-    abstract static class PipelineInitializer {
+    private abstract static class PipelineInitializer {
 
-        abstract void init(ChannelPipeline pipeline) throws Exception;
```
andsel, line 91: Why this change,  PipelineInitializer speacks about a ChannelPipeline, why moving to SocketChannel?

andsel (commented):
I think that's almost ok. It's not clear to me why in PipileneInitializer you switched the argument of init from ChannelPipeline to SocketChannel, considering that you then retrieve the ChannelPipeline from SocketChannel itself
Then please add a section in the distribution/src/main/resources/moquette.conf with the various option of configuration for the SSL handler, with line commented.

broker/src/main/java/io/moquette/server/DefaultMoquetteSslContextCreator.java
mscheibler: The `KEY_MANAGER_PASSWORD_PROPERTY_NAME` is still used to initialize the KeyManagerFactory (when the JDK provider is used) and also to access the private key (for handover to the OpenSSL provider). However i did rename the variable to `keyPassword` as i think this describes the purpose of this password better. According to the Javadoc it is *the password for recovering keys in the KeyStore*
Just the `loadKeyStore()` method does not need it.

broker/src/main/java/io/moquette/server/netty/NettyAcceptor.java
mscheibler: Because i handover the `SocketChannel` to the `createSslHandler` method.
It would also be possible to retrieve the channel from the pipeline, but then i would need to add a cast, as the pipeline just returns a `Channel` not a `SocketChannel` and i wanted to avoid that.
The channel information about the remote peer is then used to initialize the SSL engine.

andsel:
Thanks @mscheibler

Issue #412, opened 2018-06-15
Protocol Violation PUBLISH accepts zero length topic


ckrey:
### Expected behavior
Connection close

### Actual behavior
accept and publish to subscribers with '#' topicfilter

### Steps to reproduce

publish a message with zero length topic

### Moquette MQTT version
0.11

### OS version (e.g. `uname -a`)

> All Topic Names and Topic Filters MUST be at least one character long [MQTT-4.7.3-1]
see All Topic Names and Topic Filters MUST be at least one character long [MQTT-4.7.3-1] p 58

andsel:
Fixed with 4dff5b51ebef14f091e20bfdb9c0d0fbc31f7173

ckrey:
I re-tested and found the new check is performed only in QoS1, not for QoS0 or QoS2

PR #416, opened 2018-06-23
Fix issue #412,  Zero-length topic messages accepting

hiro-gh27:
Depending on the implementation of the MQTT client, it seems that when receiving "" (<-zero length) message, it detects the protocol error and terminates the connection from the client side.

For subscriptions of "#", PUBLISH(TOPIC:"") shouldn't judge to be applicable in the broker side. So, I reviewed the topic's Match and Empty condition. As a result, messages of PUBLISH (TOPIC: "") will not be sent to SUBSCRIBE (TOPIC: "#").

### DIFF
--- a/broker/src/main/java/io/moquette/spi/impl/subscriptions/Topic.java
+++ b/broker/src/main/java/io/moquette/spi/impl/subscriptions/Topic.java
@@ -128,7 +128,10 @@ public class Topic implements Serializable {
 
     public boolean isEmpty() {
         final List<Token> tokens = getTokens();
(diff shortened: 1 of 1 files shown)


andsel:
Hi @hiro-gh27 , Travis says that the PR broke something, so you don't fixed the tests, and then you speak about zero length messages (that's permissible by spec [MQTT-3.3.1-10] and is used to clear the existing retained message published on a topic) while your change is about topic matching. Could you explain better, and how to reproduce.

andsel:
I solved this raising an error in Topic parsing during the processing of Publish.
Thanks

PR #419, opened 2018-07-02
Refactor metrics management

sbouchex:
Expose metrics to embedded server


andsel:
I don't think that exposing a method to poll continuously to get fresh metrics is a good design; probably it's better to have some form of subscription

PR #422, opened 2018-07-09
Adds client remote address getter to ConnectionDescriptor class.

jcroig:
Being able to retrieve client remote address would be very useful to implement ip black list or to use it in the user interceptors  for logging / debug or any specific bespoke behaviour.


andsel:
If you want to blacklist a client probably you have to close the connection before creating the ConnectionDescriptor, but the IP could be useful in logging

PR #428, opened 2018-08-10
Update MQTTMessageLogger.java

ricardojlrufino:
reduce the level of logging of less significant things


ricardojlrufino:
The log category is not in default pattern, correct must be:
log4j.category.io.moquette.server.netty.metrics=WARN, messagelog
Not: log4j.category.messageLogger=WARN, messagelog

PR #431, opened 2018-09-26
Fixes NPE when client disconnects from the broker

jcroig:
There's NPE polluting the logs some times when client disconnects abruptly because there's no active session and so ProtocolProcessor::348 throws NPE.


andsel:
No more applicable to version v0.12, due to heavy restructure

PR #432, opened 2018-10-19
LoggerFactory.getLogger: Wrong logging class

davidgraeff:
Replace "MessageLogger" with full class name, as probably intended.

Signed-of-By: David Graeff <david.graeff@web.de>


andsel:
"messageLogger" is the name of the logger to have separate log file fore messages. This name is  used into the log4j definition.

davidgraeff:
Yes but in a complex multiple bundle application the name is just "MessageLogger" and normally as per unwritten convention you use the full class name. All your other classes use the logger class correctly.

andsel:
You are right, checked in with 2416347beadf1c571a621a950f7bfb1406d928b1

PR #449, opened 2019-01-28
Reimplemented `getSubscriptions()` function.

SpeckiJ:
`getSubscriptions()` was commented out during refactorings between v0.11 and 0.12.

Resolved TODO and reimplemented `getSubscriptions()` function.


andsel:
This is not acceptable PR because, ISubscriptionRepository is intend to be used internally by the broker, especially from ISubscirptionDirectory, so at least this method should contact ISubscriptionDirectory.

I commented it to re-implement in something smarter, because it could potentially load many subscriptions. I prefere a listener approach for the client library, the client library should register no be notified about subscription changes, I don't have a valid use case to expose the full list of subcriptions

PR #450, opened 2019-01-28
Enhanced exception handling / logging for invalid messages

mscheibler:
Just a small enhancement when it comes to handling invalid messages.
The current implementation suppresses the root cause of the invalid message, which make problem tracking difficult.


andsel:
Hi @mscheibler  seems reasonable, thanks

PR #459, opened 2019-03-20
Refactors config implementation: Use java.util.Properties directly now

flonix8:
I simplified the config implementation the following way:
* Removing `ConfigurationParser` and replacing it with good old `java.util.Properties` as I couldn't find a reason to use the custom implementation (if there is, please let me now :) )
* Moving property access methods of `*Config` classes up to `IConfig`

I'd love to hear some feedback on that.


andsel:
@flonix8 I don't understand why did you remove the config file parser? That parser was intended to use the same config synthax of Mosquitto that's not a properties files, it's like having a project configurable by YAML, properties or JSON config files, it's just a config grammar

flonix8:
I might have a slight misunderstanding here. 

Let my explain myself:
I'm using mosquette for a university project and I need multi-line config values. It was mainly a readability thing, so I quickly wanted to add it but then I noticed that the existing `ConfigurationParser` does nothing that the `Properties` read method wouldn't also do. The config keys and values end up in a `Properties` object anyway, so I thought I might just remove the custom read implementation.

As I see it the `ConfigurationParser` is not compatible to the Mosquitto config format. For instance, it can't handle some of the multi-value config options in Mosquitto config files, at least not without any further processing. The behavior of `Properties` should be exactly the same in that regard...

andsel:
Moquette coudl offer both configuration formats, properties or conf file.

flonix8:
I'm sorry, but I really don't know what you mean.
The current config file format (as implicitly defined by the `ConfigurationParser`) is already a perfectly valid Properties file. I'd say it's a subset of a Properties file, because the `Properties` parser supports a lot more features (multiline values for instance).
So there really is no change to the file format needed here. :smile:

andsel:
Currently Moquette support (more or less) the mosquitto.conf file format.
You wrote an implementation for Java's properties files as a replacement,
what I mean is that we could think to support both format as valid config
file formats, it the file is a properties then we use that if it's conf
type we can parse and use that, abstracting from the file format. This was
the reason why I created IConfig, to use various implemnetations.

flonix8:
It would certainly be cool if different file formats were possible, as long as the right values end up attached to the right keys.
What would you say differentiates the existing `ConfigurationParser.parse()` from `Properties.read()`?
Because that's the part that is really puzzling me. I know that `Properties.read()` supports more features, but it at least supports all of the syntax of `ConfigurationParser.parse()`, doesn't it?

andsel:
You are essentially right, `ConfigurationParser.parse()` produces a `Properties` instance, than that property intance is used by a subaclass of IConfig, simply wrapping the access to the properties instance with typed method (i mean  the 2 method `intProp` and `booProp`.

The intention of the classes is:
- `ConfigurationParser` should be just a parser that produces an instance of `Properties` that's consumed by a subclass of  `IConfig`. This means that you could drop the ConfigParser adn provide an implementation of Properties that load the props from a Mosquitto config synthax file
 - `IConfig` defines the typed accessor methods to the underling Properties instance, the subclasses (`ResourceLoaderConfig` and `MemoryConfig` are intended to load the configuration from a resouce file or from an in memory configuration)
- `IResourceLoader` and the 2 subclasses `ClasspathResourceLoader` and `FileResourceLoader` are responsible to provide the content of the config file to load.


So recap, we have a ResourceLoaderConfig that loads the text content of the config file from somewhere (could be local filesystem or classpath) and pass this to a ConfigParser to create and wrap the Properties.
The other path is:  MemoryConfig wraps a Properties instancs to provide typed access.

I've to admit that the design it's a little bit convoluted, but some concept could be kept in a redesign:
- separate the parsing phases to plug other types config (for example yml files)
- provide typed accessor methods (to avoid parsing and casting from `Properties.getProperties()`
- provide an in-memory configuration rappresentation.

flonix8:
Sorry for not getting back to you earlier!
Sadly, I don't have any time to spare right now (due to my ongoing thesis), so I won't be able to assist until autumn...
If you don't feel that my changes are in line for what you have in mind for this project, please go ahead and discard them, I don't mind. Maybe keep the thread for future reference :)

PR #462, opened 2019-03-25
Added possibility to configure additional Netty channel handlers

aemaem:
Added possibility to configure additional Netty channel handlers in order to extend Moquette configuration possibility.

Example to add CORS configuration in Kotlin:
```kotlin
val pipelineConfigurer: INettyChannelPipelineConfigurer = object : INettyChannelPipelineConfigurer {
    override fun configure(pipeline: ChannelPipeline?) {
        if (pipeline == null) throw RuntimeException("Unable to configure pipeline")
        pipeline.addBefore("aggregator", "cors", CorsHandler(CorsConfigBuilder.forAnyOrigin().build()))
    }
}
Server().startServer(null, null, null, null, null, pipelineConfigurer)
```

### DIFF
--- a/broker/src/main/java/io/moquette/broker/NewNettyAcceptor.java
+++ b/broker/src/main/java/io/moquette/broker/NewNettyAcceptor.java
@@ -18,6 +18,7 @@ package io.moquette.broker;
 
 import io.moquette.BrokerConstants;
 import io.moquette.broker.config.IConfig;
+import io.moquette.broker.config.INettyChannelPipelineConfigurer;
 import io.moquette.broker.metrics.*;
 import io.netty.bootstrap.ServerBootstrap;
 import io.netty.buffer.ByteBuf;
@@ -95,7 +96,6 @@ class NewNettyAcceptor {
         abstract void init(SocketChannel channel) throws Exception;
     }
 
-
     private class LocalPortReaderFutureListener implements ChannelFutureListener {
         private String transportName;
 
@@ -126,6 +126,7 @@ class NewNettyAcceptor {
     private MessageMetricsCollector metricsCollector = new MessageMetricsCollector();
     private Optional<? extends ChannelInboundHandler> metrics;
     private Optional<? extends ChannelInboundHandler> errorsCather;
+    private final INettyChannelPipelineConfigurer pipelineConfigurer;
 
     private int nettySoBacklog;
     private boolean nettySoReuseaddr;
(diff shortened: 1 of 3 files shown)


andsel:
Hi @ aemaem   what do you mean for CORS (http's CORS?) If yes which is ht reason or the intended behaviour to manage CORS after the MQTT?

aemaem:
Hey @andsel, yes HTTP's CORS. Actually CORS has nothing to do with the pull request itself. The changes I made in code let you configure additional channel handlers, where [CorsHandler](https://netty.io/4.1/api/io/netty/handler/codec/http/cors/CorsHandler.html) is just a use case.

broker/src/main/java/io/moquette/broker/NewNettyAcceptor.java
```diff
@@ -136,6 +137,10 @@ public void operationComplete(ChannelFuture future) throws Exception {
 
     private Class<? extends ServerSocketChannel> channelClass;
 
+    NewNettyAcceptor(INettyChannelPipelineConfigurer pipelineConfigurer) {
+        this.pipelineConfigurer = pipelineConfigurer == null ? new INettyChannelPipelineConfigurer() {} : pipelineConfigurer;
```
andsel, line 141: Instead of an inner anonymous implementation is better to create a class for example EmptyPipelineCOnfigured implments INettyChannelPipelineConfigurer.
Beeing the pipeline configurer used in initialize method, probably is better to pass the reference there instead that in the constructor

broker/src/main/java/io/moquette/broker/config/INettyChannelPipelineConfigurer.java
```diff
@@ -0,0 +1,9 @@
+package io.moquette.broker.config;
+
+import io.netty.channel.ChannelPipeline;
+
+public interface INettyChannelPipelineConfigurer {
+
+    default void configure(ChannelPipeline pipeline) {
```
andsel, line 7: Why an empty default method, I mean the implementation of this interface should implement the INettyChannelPipelineConfigurer to customized the pipeline

andsel (commented):
Hi @aemaem  the idea of the pipeline configurer could be interesting, let the client code that use and integrate with Moquette to put some configurations. Instead of manipulate the pipeline on which Moquette decoder rely on it's best to have a callback mechanism like the one already present in Moquette based on Interceptors. Which is the intention of you PR?

aemaem:
As stated in the title of this PR, it is intended to configure additional Netty channel handlers. Please take the idea behind this PR and implement it the way you think it is best for your project. I won't provide a new PR for this.

andsel:
Ok thanks.

PR #468, opened 2019-04-08
Make the number of Netty worker threads number configurable

komamitsu:
I noticed Moquette users can't configure the number of Netty worker threads in current Moquette. Increasing Netty worker threads number is important in our use case that handles a lot of MQTT requests.

This PR makes it configurable.


andsel:
This is not necessary. Netty used asunch IO an Moquette is essentially IO
bound so adding thread only adds threads that do IO waiting. The broker
starts and Co figure the number of threads equals to the number of cores -
1, let this configurable to put for example 1000 threads it's a no sense,
just adds weight to the scheduler and stole memory for each thread's stack.

Il lun 8 apr 2019, 07:24 Mitsunori Komatsu <notifications@github.com> ha
scritto:

PR #478, opened 2019-05-21
remove unnecessary quotes which fixes execution in windows 10

joestr:
removed unnecessary quotes on line 34 and  line 86; these quotes make an execution in Windows 10 impossible;


andsel:
Thanks a lot @joestr

PR #479, opened 2019-06-02
fix trie match bug and bytesMetrics bug

ilylia:
fix trie match bug when topic contains root;
fix bytesMetrics wroteBytes always 0;
fix startup shell can not run outside bin;

### DIFF
--- a/broker/src/main/java/io/moquette/broker/subscriptions/CTrie.java
+++ b/broker/src/main/java/io/moquette/broker/subscriptions/CTrie.java
@@ -14,7 +14,7 @@ public class CTrie {
         T getResult();
     }
 
-    private static final Token ROOT = new Token("root");
+    private static final Token ROOT = new Token("+root");
     private static final INode NO_PARENT = null;
 
     private enum Action {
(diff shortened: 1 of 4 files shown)


andsel:
Sorry @ilylia but this PR miss an important part, at least one test that manifest the problem you are trying to solve.

ilylia:
Hi @andsel , there is a test case for the trie match bug in the new pr, please check.

broker/src/main/java/io/moquette/broker/subscriptions/CTrie.java
```diff
@@ -14,7 +14,7 @@
         T getResult();
     }
 
-    private static final Token ROOT = new Token("root");
+    private static final Token ROOT = new Token("+root");
```
andsel, line 17: @ilylia  did you put a + sign only because to be able to create an internal token marker? I understand but could sound little bit confusing, did you consider other possible tokens?

ilylia: because only `+` and `#` are the reserved key which can not appear in the tokens parsed from topic. other keys may conflict with normal tokens.

Issue #488, opened 2019-07-16
Retained repository retainedOnTopic method calls Topic.match incorrectly


davesters:
From what I can tell, the call to `Topic.match` is being done incorrectly in the `retainedOnTopic` methods of `MemoryRetainedRepository`, and maybe also the other implementations of `IRetainedRepository`.

For example, in `MemoryRetainedRepository` line 60 it has this code.

```
if (searchTopic.match(scanTopic)) {
    matchingMessages.add(entry.getValue());
}
```

From what I can tell, it needs to be like this.

```
if (scanTopic.match(searchTopic)) {
    matchingMessages.add(entry.getValue());
}
```

`searchTopic` is what the client subscribed to, and `scanTopic` is the topic of the retained message. Right now, it does a match as if `scanTopic` is what the client subscribed to. I think this should be a pretty easy fix to just swap them, but would need some tests around it.

Topic unit tests looks good, but no specific unit tests exist for this scenario that I could find.

### Expected behavior
When I have retained messages for a topic like `myhome/kitchen/light`, and a client subscribes to `myhome/kitchen/#` they receive the retained message.

### Actual behavior
Retained messages are not received unless the client subscribes to exact topic `myhome/kitchen/light`.

### Steps to reproduce
Call `internalPublish` with a message to the topic listed above. Connect with a client and subscribe to a wildcard version of that topic.

### Moquette MQTT version
v0.12.1

### JVM version (e.g. `java -version`)
Java version "1.8.0_144"
Java(TM) SE Runtime Environment (build 1.8.0_144-b01)
Java HotSpot(TM) 64-Bit Server VM (build 25.144-b01, mixed mode)

### OS version (e.g. `uname -a`)
Darwin Davids-MBP.lan 18.5.0 Darwin Kernel Version 18.5.0: Mon Mar 11 20:40:32 PDT 2019; root:xnu-4903.251.3~3/RELEASE_X86_64 x86_64

davesters:
I am going to make this change and see if I starts working for me as I expect. If I am able to verify it works, I'll try to add some unit tests around the `MemoryRetainedRepository` class to test it.

Thanks for the awesome library BTW!

PR #489, opened 2019-07-17
Fix retained messages with wildcard subscriptions

davesters:
This PR is a fix for #488 

I changed the conditional for `retainedOnTopic` method of `MemoryRetainedRepository` and added some tests around it.

This same change probably also needs to happen to `H2RetainedRepository`, but I did not want to mess with that as I was unsure how to add tests for it, etc.

This also uncovered another bug around retained messages. When you subscribe to a wildcard topic that has any number of retained messages that match it, those retained messages get sent with the same topic as you subscribed. This is incorrect, and the retained messages should also retain their respective topics. I added a topic field to the `RetainedMessage` class also and then used it in the `publishRetainedMessagesForSubscriptions` method of the `PostOffice` class so they are sent with the correct topics. This probably also needs tests, but I was unsure how to add them as they are a bit complex.

### DIFF
--- a/broker/src/main/java/io/moquette/broker/MemoryRetainedRepository.java
+++ b/broker/src/main/java/io/moquette/broker/MemoryRetainedRepository.java
@@ -42,7 +42,7 @@ final class MemoryRetainedRepository implements IRetainedRepository {
         final ByteBuf payload = msg.content();
         byte[] rawPayload = new byte[payload.readableBytes()];
         payload.getBytes(0, rawPayload);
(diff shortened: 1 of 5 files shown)


martinvw:
Hello, thank you very much for fixing it! Is there already a plan to build a new release which includes this fix?

davesters:
@martinvw I ended up building my own new version and pushing it to our company's private maven registry. I am unsure of when this would make it into a new release, but I am looking forward to it.

PR #498, opened 2019-08-17
Java 11 compatibility

qbast:
Gradle 4.6 does not work with Java 11.
Update Gradle to newest version (5.6) along with errorprone and spotbugs in order to make moquette buildable with JDK 11. Compatibility with Java 8 is retained.


hylkevds:
This PR is no longer relevant, since the revert to Maven.

Issue #531, opened 2020-07-23
New configuration parameter : setting the HandshakeTimeout on SslHandler


dsoulayrol:
Hi,

I am working with embedded devices which are very limited in resources, and the server certificate verification takes a very long time. The server closes the connection after 10 seconds, far before the client has completed its computation to ask for the key exchange.

This timeout is set by `io.netty.handler.ssl.SslHandler`. This handler is created in `NewNettyAcceptor#createSslHandler`. (I am using version 0.12.1, but I see nothing changed in master). When I call `SslHandler#setHandshakeTimeoutMillis` here, my problem is solved.

So I'd like to have the capability to provide this timeout value as a property for moquette. I believe it would be an interesting feature because MQTT is easily used on embedded devices, and SSL support is somewhat heavy, so my problem could be seen elsewhere.

I am ready to create a Pull Request if necessary, but I would be happy to hear your comments first on the subject.

dsoulayrol:
Hi.

I have to freeze and package a product in a few days which needs this feature. I need to know if I have a chance to see this request soon merged so as to rely on the main repository of moquette, or if I have to generate my own version.
Could someone post a comment about this?

Thanks.

PR #547, opened 2020-11-05
Added configuration item for handshake timeout on SSL channel (#531)

dsoulayrol:


broker/pom.xml
```diff
@@ -5,7 +5,7 @@
         <relativePath>../</relativePath>
         <artifactId>moquette-parent</artifactId>
         <groupId>io.moquette</groupId>
-        <version>0.14-SNAPSHOT</version>
+        <version>0.14-issue-531</version>
```
andsel, line 8: @dsoulayrol I don't understand the change from SNAPSHOT version

dsoulayrol: Sorry, this independent commit is only there for my own use of the fork. I did not think it would come with the pull request. Are there any other problems in the proposed feature ? Should I create another pull request ?

andsel: Nothing I can see, but before merging:
- document `netty.channel_handshake_timeout.seconds` in https://github.com/moquette-io/moquette/blob/master/distribution/src/main/resources/moquette.conf
- squash all the commits

Then I could merge it

PR #551, opened 2020-11-13
Added configuration item for handshake timeout on SSL channel (#531)

dsoulayrol:
A cleaner version with the requested parameter description in `moquette.conf`.


hylkevds (approved):
Looks fine.

PR #552, opened 2020-11-16
GitHub actions

HanSolo:
Added a github action that runs mvn verify on each push on Ubuntu 18.04 using JDK 11, 12 and 15


andsel:
@HanSolo thank's a lot for your contribution, I've renamed it to `maven_build` because I've the intention to also create a workflow for bintray deploy on tag creation. Let me know if the changes I've done are in line with your original intent

PR #572, opened 2021-02-02
Fixes #571: ConcurrentModificationException in MemorySubscriptionsRepository

### LINKED ISSUE #571: 0.14: ConcurrentModificationException in MemorySubscriptionsRepository
The MemorySubscriptionsRepository implementation does not synchronise write access to the subscriptions list. As a result, Iterations over the list will break when another Thread modifies the list.

I'm not sure what the best way to solve this is:
- the easy way is to use a CopyOnWriteArrayList to protect iterations, and to synchronise add and remove operations. This would not be good for add/remove performance.
- Another option is to use a ConcurrentSkipListSet that has better add/remove performance, but that would mean the interface of the `public List<Subscription> listAllSubscriptions()` method would need to change to return a `Set`.

What do you think?
I'll make a PR for the change to ConcurrentSkipListSet. The other one is pretty trivial.

### Expected behavior
No exceptions

### Actual behavior
ConcurrentModificationExceptions in MemorySubscriptionsRepository

### Steps to reproduce
Have many clients subscribe and unsubscribe often, while publishes are in progress.

- https://github.com/hylkevds/moquetteTests/
- In MoquetteTest
  - Set CLIENT_LIVE_MILLIS = 100;
  - Set threadCountListen = 200;
  - Comment-out line 213 that sets the PERSISTENT_STORE_PROPERTY_NAME, so that the MemorySubscriptionsRepository is used

### Moquette MQTT version
0.14

OpenJDK 11

Linux

hylkevds:
The MemorySubscriptionsRepository implementation does not synchronise
write access to the subscriptions list. As a result, Iterations over the
list will break when another Thread modifies the list.
This changes the ArrayList to a ConcurrentSkipListSet and changes the
interface of the public List<Subscription> listAllSubscriptions() method
to return a Set.

### DIFF
--- a/broker/src/main/java/io/moquette/persistence/MemorySubscriptionsRepository.java
+++ b/broker/src/main/java/io/moquette/persistence/MemorySubscriptionsRepository.java
@@ -18,17 +18,23 @@ package io.moquette.persistence;
 import io.moquette.broker.ISubscriptionsRepository;
 import io.moquette.broker.subscriptions.Subscription;
 
-import java.util.ArrayList;
 import java.util.Collections;
-import java.util.List;
+import java.util.Set;
+import java.util.concurrent.ConcurrentSkipListSet;
 
 public class MemorySubscriptionsRepository implements ISubscriptionsRepository {
 
-    private final List<Subscription> subscriptions = new ArrayList<>();
+    private final Set<Subscription> subscriptions = new ConcurrentSkipListSet<>((o1, o2) -> {
+        int compare = o1.getClientId().compareTo(o2.getClientId());
+        if (compare != 0) {
+            return compare;
+        }
+        return o1.getTopicFilter().toString().compareTo(o2.getTopicFilter().toString());
+    });
 
     @Override
-    public List<Subscription> listAllSubscriptions() {
-        return Collections.unmodifiableList(subscriptions);
+    public Set<Subscription> listAllSubscriptions() {
+        return Collections.unmodifiableSet(subscriptions);
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


broker/src/main/java/io/moquette/persistence/MemorySubscriptionsRepository.java
```diff
@@ -18,17 +18,23 @@
...
-import java.util.ArrayList;
 import java.util.Collections;
-import java.util.List;
+import java.util.Set;
+import java.util.concurrent.ConcurrentSkipListSet;
 
 public class MemorySubscriptionsRepository implements ISubscriptionsRepository {
 
-    private final List<Subscription> subscriptions = new ArrayList<>();
+    private final Set<Subscription> subscriptions = new ConcurrentSkipListSet<>((o1, o2) -> {
```
andsel, line 27: Instead of providing an explicit comparator is better to implements the Comparable interface, PR for the change is  #595

andsel (changes requested):
Hi @hylkevds I've create #595 to switch to Topic and Subscription to be comparable.
Check if you want that PR goes before on `master` on embed into this one

broker/src/main/java/io/moquette/persistence/MemorySubscriptionsRepository.java
hylkevds: Yeah, totally agree. Only reason I didn't was to minimise the impact of the PR.
After merging #595, I can either update this PR, or you merge it and then fix the constructor to not supply a Comparator.

Issue #506, opened 2019-10-09
BrokerInterceptor thread pool never closing on serverStop


bobrowskim:
### Expected behavior

When using moquette-io as embedded broker in application when invoking method `stopServer` from class `Server` it would be expected that all related threads are closing.

### Actual behavior

Invoking `stopServer` from class `Server` does not call to stop `BrokerInterceptor`  thus the threads pool it created is never shutdown and some thread is kept in awaiting state. This leads to application hang running.

### Steps to reproduce

1. Start `Server`
2. Connect client to server, publish and subscribe to some topics
3. Stop server `Server.stopServer`
  - Thread pool created by `BrokerInterceptor` is still in waiting state
4. Application never stops.

### Minimal yet complete reproducer code (or URL to code) or complete log file
[Here is sample code](https://github.com/bobrowskim/moquette-closing-issue)

Run class `NotClosingApplication`. 
 
### Moquette MQTT version
Tested on version: `0.12.1`

### JVM version (e.g. `java -version`)
Tested with: `1.8_162`, `11.0.1` and `13`

### OS version (e.g. `uname -a`)

`Windows 10` and `OSx 10.11.6`

mjacoby:
Any progress on this?
I just stumbled across the exact same issue and could reproduce it with Moquette version 0.11, 0.13 and 0.14.

Actually, my sample scenario is even a bit simpler: start server, do publish, stop server; which results in the application never closing as there is still a thread-pool thread alive. Just starting and stopping a server without doing a publish works just fine, however adding a call to publish(...) seems to spawn that thread that is never closed.

Here's my smaple code:

```
package MoquetteTest;

import io.moquette.BrokerConstants;
import io.moquette.broker.Server;
import io.moquette.broker.config.IConfig;
import io.moquette.broker.config.MemoryConfig;
import java.io.IOException;
import java.util.Properties;
import org.eclipse.paho.client.mqttv3.MqttClient;
import org.eclipse.paho.client.mqttv3.MqttConnectOptions;
import org.eclipse.paho.client.mqttv3.MqttException;
import org.eclipse.paho.client.mqttv3.MqttMessage;
import org.eclipse.paho.client.mqttv3.persist.MemoryPersistence;

public class Main {

    private static final String HOST = "localhost";
    private static final int PORT = 1883;
    private static Server server;

    public static void main(String[] args) throws IOException, MqttException, InterruptedException {
        startServer();
        publish();
        Thread.sleep(2000);
        stopServer();
[...]

PR #575, opened 2021-02-10
#506 Close BrokerInterceptor with Server

bobrowskim:
Fix to the issue #506


andsel:
Thank's @bobrowskim nice catch

PR #576, opened 2021-02-17
Fixed #573 ByteBuf reference counting

### LINKED ISSUE #573: 0.14: IllegalReferenceCountException: refCnt: 0
This one is probably hard to track down. I've only seen it happen once in the last few days.
It did cause the connection to the client to be dropped.
I'm currently using the in-memory persistence store.

### Expected behavior
No Exceptions

### Actual behavior
```
10:02:45.602 [nioEventLoopGroup-3-1] ERROR i.m.broker.NewNettyMQTTHandler - Unexpected exception while processing MQTT message. Closing Netty channel. CId=Aggregator
io.netty.util.IllegalReferenceCountException: refCnt: 0
	at io.netty.buffer.AbstractByteBuf.ensureAccessible(AbstractByteBuf.java:1490)
	at io.netty.buffer.AbstractByteBuf.duplicate(AbstractByteBuf.java:1201)
	at io.netty.buffer.AbstractByteBuf.retainedDuplicate(AbstractByteBuf.java:1207)
	at io.moquette.broker.Session.resendInflightNotAcked(Session.java:308)
	at io.moquette.broker.MQTTConnection.resendNotAckedPublishes(MQTTConnection.java:502)
	at io.moquette.broker.NewNettyMQTTHandler.userEventTriggered(NewNettyMQTTHandler.java:112)
	at io.netty.channel.AbstractChannelHandlerContext.invokeUserEventTriggered(AbstractChannelHandlerContext.java:346)
	at io.netty.channel.AbstractChannelHandlerContext.invokeUserEventTriggered(AbstractChannelHandlerContext.java:332)
	at io.netty.channel.AbstractChannelHandlerContext.fireUserEventTriggered(AbstractChannelHandlerContext.java:324)
	at io.netty.channel.ChannelInboundHandlerAdapter.userEventTriggered(ChannelInboundHandlerAdapter.java:117)
	at io.netty.handler.codec.ByteToMessageDecoder.userEventTriggered(ByteToMessageDecoder.java:365)
	at io.netty.channel.AbstractChannelHandlerContext.invokeUserEventTriggered(AbstractChannelHandlerContext.java:346)
	at io.netty.channel.AbstractChannelHandlerContext.invokeUserEventTriggered(AbstractChannelHandlerContext.java:332)
	at io.netty.channel.AbstractChannelHandlerContext.fireUserEventTriggered(AbstractChannelHandlerContext.java:324)
	at io.netty.channel.ChannelInboundHandlerAdapter.userEventTriggered(ChannelInboundHandlerAdapter.java:117)
	at io.moquette.broker.MoquetteIdleTimeoutHandler.userEventTriggered(MoquetteIdleTimeoutHandler.java:48)
	at io.netty.channel.AbstractChannelHandlerContext.invokeUserEventTriggered(AbstractChannelHandlerContext.java:346)
	at io.netty.channel.AbstractChannelHandlerContext.invokeUserEventTriggered(AbstractChannelHandlerContext.java:332)
	at io.netty.channel.AbstractChannelHandlerContext.fireUserEventTriggered(AbstractChannelHandlerContext.java:324)
	at io.moquette.broker.InflightResender.resendNotAcked(InflightResender.java:160)
	at io.moquette.broker.InflightResender.access$100(InflightResender.java:32)
	at io.moquette.broker.InflightResender$WriterIdleTimeoutTask.run(InflightResender.java:58)
	at io.netty.util.concurrent.PromiseTask.runTask(PromiseTask.java:98)
	at io.netty.util.concurrent.ScheduledFutureTask.run(ScheduledFutureTask.java:170)
	at io.netty.util.concurrent.AbstractEventExecutor.safeExecute(AbstractEventExecutor.java:164)
	at io.netty.util.concurrent.SingleThreadEventExecutor.runAllTasks(SingleThreadEventExecutor.java:472)
	at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:500)
	at io.netty.util.concurrent.SingleThreadEventExecutor$4.run(SingleThreadEventExecutor.java:989)
	at io.netty.util.internal.ThreadExecutorMap$2.run(ThreadExecutorMap.java:74)
	at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
	at java.base/java.lang.Thread.run(Thread.java:834)
```

### Steps to reproduce
See below

None yet

### Moquette MQTT version
0.14 (with the patches from #568, #570, #572)

OpenJDK 11

Linux / Docker openjdk:11

hylkevds:
The main place where ByteBufs were not released was the inflightWindow.
Also improves two postoffice methods that passed the payload and retain
information twice, once direct and once in the message.

### DIFF
--- a/broker/src/main/java/io/moquette/broker/MQTTConnection.java
+++ b/broker/src/main/java/io/moquette/broker/MQTTConnection.java
@@ -18,6 +18,7 @@ package io.moquette.broker;
 import io.moquette.broker.subscriptions.Topic;
 import io.moquette.broker.security.IAuthenticator;
 import io.netty.buffer.ByteBuf;
+import io.netty.buffer.ByteBufHolder;
 import io.netty.channel.Channel;
 import io.netty.channel.ChannelFuture;
 import io.netty.channel.ChannelFutureListener;
@@ -376,8 +377,9 @@ final class MQTTConnection {
             }
             case EXACTLY_ONCE: {
                 bindedSession.receivedPublishQos2(messageID, msg);
+                // Second pass-on, retain
+                msg.payload().retain();
                 postOffice.receivedPublishQos2(this, msg, username);
-//                msg.release();
                 break;
             }
             default:
@@ -426,6 +428,11 @@ final class MQTTConnection {
                 channelFuture = channel.write(msg);
             }
             channelFuture.addListener(FIRE_EXCEPTION_ON_FAILURE);
+        } else {
+            // msg not passed on, release.
+            if (msg instanceof ByteBufHolder) {
+                ((ByteBufHolder) msg).release();
+            }
         }
     }
 
--- a/broker/src/main/java/io/moquette/broker/Server.java
+++ b/broker/src/main/java/io/moquette/broker/Server.java
@@ -308,6 +308,7 @@ public class Server {
         }
         LOG.trace("Internal publishing message CId: {}, messageId: {}", clientId, messageID);
         dispatcher.internalPublish(msg);
+        msg.payload().release();
     }
 
     public void stopServer() {
--- a/broker/src/main/java/io/moquette/broker/Session.java
+++ b/broker/src/main/java/io/moquette/broker/Session.java
@@ -183,7 +183,12 @@ class Session {
     }
 
     public void processPubRec(int packetId) {
-        inflightWindow.remove(packetId);
+        // Message discarded, make sure any buffers in it are released
+        SessionRegistry.EnqueuedMessage removed = inflightWindow.remove(packetId);
+        if (removed != null) {
+            removed.release();
+        }
+
         inflightSlots.incrementAndGet();
         if (canSkipQueue()) {
             inflightSlots.decrementAndGet();
@@ -200,7 +205,12 @@ class Session {
     }
 
     public void processPubComp(int messageID) {
-        inflightWindow.remove(messageID);
+        // Message discarded, make sure any buffers in it are released
+        SessionRegistry.EnqueuedMessage removed = inflightWindow.remove(messageID);
+        if (removed != null) {
+            removed.release();
+        }
+
         inflightSlots.incrementAndGet();
 
         drainQueueToConnection();
@@ -216,6 +226,9 @@ class Session {
             case AT_MOST_ONCE:
                 if (connected()) {
                     mqttConnection.sendPublishNotRetainedQos0(topic, qos, payload);
+                } else {
+                    // buffer not passed on, release it.
+                    payload.release();
                 }
                 break;
             case AT_LEAST_ONCE:
@@ -226,12 +239,16 @@ class Session {
                 break;
             case FAILURE:
                 LOG.error("Not admissible");
+                // buffer not passed on, release it.
+                payload.release();
         }
     }
 
     private void sendPublishQos1(Topic topic, MqttQoS qos, ByteBuf payload) {
         if (!connected() && isClean()) {
             //pushing messages to disconnected not clean session
+            //buffer not passed on, release it.
+            payload.release();
             return;
         }
 
@@ -240,6 +257,9 @@ class Session {
             int packetId = mqttConnection.nextPacketId();
             inflightWindow.put(packetId, new SessionRegistry.PublishedMessage(topic, qos, payload));
             inflightTimeouts.add(new InFlightPacket(packetId, FLIGHT_BEFORE_RESEND_MS));
+            
+            // second pass-on, add retain
+            payload.retain();
             MqttPublishMessage publishMsg = MQTTConnection.notRetainedPublishWithMessageId(topic.toString(), qos,
                                                                                            payload, packetId);
             mqttConnection.sendPublish(publishMsg);
@@ -257,6 +277,9 @@ class Session {
             int packetId = mqttConnection.nextPacketId();
             inflightWindow.put(packetId, new SessionRegistry.PublishedMessage(topic, qos, payload));
             inflightTimeouts.add(new InFlightPacket(packetId, FLIGHT_BEFORE_RESEND_MS));
+
+            // second pass-on, add retain
+            payload.retain();
             MqttPublishMessage publishMsg = MQTTConnection.notRetainedPublishWithMessageId(topic.toString(), qos,
                                                                                            payload, packetId);
             mqttConnection.sendPublish(publishMsg);
@@ -283,7 +306,11 @@ class Session {
 
     void pubAckReceived(int ackPacketId) {
         // TODO remain to invoke in somehow m_interceptor.notifyMessageAcknowledged
-        inflightWindow.remove(ackPacketId);
+        SessionRegistry.EnqueuedMessage removed = inflightWindow.remove(ackPacketId);
+        if (removed != null) {
+            removed.release();
+        }
+
         inflightSlots.incrementAndGet();
         drainQueueToConnection();
     }
@@ -347,6 +374,8 @@ class Session {
                 mqttConnection.sendIfWritableElseDrop(pubRel);
             } else {
                 final SessionRegistry.PublishedMessage msgPub = (SessionRegistry.PublishedMessage) msg;
+                // Second pass-on.
+                msgPub.payload.retain();
                 MqttPublishMessage publishMsg = MQTTConnection.notRetainedPublishWithMessageId(msgPub.topic.toString(),
                     msgPub.publishingQos,
                     msgPub.payload, sendPacketId);
--- a/broker/src/main/java/io/moquette/broker/PostOffice.java
+++ b/broker/src/main/java/io/moquette/broker/PostOffice.java
@@ -237,6 +237,8 @@ class PostOffice {
         final String clientId = connection.getClientId();
         if (!authorizator.canWrite(topic, username, clientId)) {
             LOG.error("MQTT client is not authorized to publish on topic: {}", topic);
+            // msg not passed on, release payload.
+            payload.release();
             return;
         }
 
@@ -254,6 +256,8 @@ class PostOffice {
 
         String clientID = connection.getClientId();
         interceptor.notifyTopicPublished(mqttPublishMessage, clientID, username);
+        // none of the methods above released the payload, do it now.
+        payload.release();
     }
 
     static MqttQoS lowerQosToTheSubscriptionDesired(Subscription sub, MqttQoS qos) {
--- a/broker/src/main/java/io/moquette/broker/SessionRegistry.java
+++ b/broker/src/main/java/io/moquette/broker/SessionRegistry.java
@@ -38,6 +38,12 @@ import java.util.stream.Collectors;
 public class SessionRegistry {
 
     public abstract static class EnqueuedMessage {
+
+        /**
+         * Releases any held resources. Must be called when the EnqueuedMessage is no
+         * longer needed.
+         */
+        public void release() {}
     }
 
     public static class PublishedMessage extends EnqueuedMessage {
@@ -62,6 +68,11 @@ public class SessionRegistry {
 
         public ByteBuf getPayload() {
             return payload;
+        }
+
+        @Override
+        public void release() {
+            payload.release();
         }
     }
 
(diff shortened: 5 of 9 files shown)


andsel:
Hi @hylkevds to simplify reasoning about buffer allocation could you:
- avoid in this PR the improvements in postoffice methods
- fix the conflicts.
this should help grasp better the PR, thanks

hylkevds:
I'll update the PR, and split out the postoffice bits in a separate one.

broker/src/main/java/io/moquette/broker/MQTTConnection.java
```diff
@@ -376,8 +377,9 @@ void processPublish(MqttPublishMessage msg) {
             }
             case EXACTLY_ONCE: {
                 bindedSession.receivedPublishQos2(messageID, msg);
+                // Second pass-on, retain
```
andsel, line 380: I think that this payload retain is better to stay close to the retain of the full message, so I would move it into the `Session.receivedPublishQos2`

broker/src/main/java/io/moquette/broker/Server.java
```diff
@@ -308,6 +308,7 @@ public void internalPublish(MqttPublishMessage msg, final String clientId) {
         }
         LOG.trace("Internal publishing message CId: {}, messageId: {}", clientId, messageID);
         dispatcher.internalPublish(msg);
+        msg.payload().release();
```
andsel, line 311: This `release` call impose the client code that uses `Server.internalPublish` to do a `retain` upfront (like in https://github.com/moquette-io/moquette/pull/576/files#diff-3b795bc663c1934e2316bbeee5566fd65aa5c2e2a2804e2bdc6f812e54f4a53aR150) and respect this not explicit behavior:
```
payload.retain
server.internalPublish
```

I think in this case is responsibility of the client code to decide if retain/release the payload or not

broker/src/main/java/io/moquette/broker/Session.java
```diff
@@ -183,7 +183,12 @@ boolean isClean() {
     }
 
     public void processPubRec(int packetId) {
-        inflightWindow.remove(packetId);
+        // Message discarded, make sure any buffers in it are released
+        SessionRegistry.EnqueuedMessage removed = inflightWindow.remove(packetId);
+        if (removed != null) {
```
andsel, line 188: This case happen only if the broker receives a PUBREC (from a client) with  not matching packetID, in which case the broker should ignore the message and don't do any processing

broker/src/main/java/io/moquette/broker/Session.java
```diff
@@ -200,7 +205,12 @@ public void processPubRec(int packetId) {
     }
 
     public void processPubComp(int messageID) {
-        inflightWindow.remove(messageID);
+        // Message discarded, make sure any buffers in it are released
+        SessionRegistry.EnqueuedMessage removed = inflightWindow.remove(messageID);
```
andsel, line 209: Same concerns as PUBREC above

broker/src/main/java/io/moquette/broker/Session.java
```diff
@@ -283,7 +306,11 @@ private boolean inflighHasSlotsAndConnectionIsUp() {
 
     void pubAckReceived(int ackPacketId) {
         // TODO remain to invoke in somehow m_interceptor.notifyMessageAcknowledged
-        inflightWindow.remove(ackPacketId);
+        SessionRegistry.EnqueuedMessage removed = inflightWindow.remove(ackPacketId);
```
andsel, line 309: Same concern as PUBREC and PUBREL with invalid packetID

andsel (changes requested):
Thanks @hylkevds for the contributions, I've some concerns I tried to fix with PR #599.

broker/src/main/java/io/moquette/broker/Server.java
hylkevds: If the client code wants to use the payload further, it should always do a retain. Passing the buffer to moquette means giving moquette the control over the payload, with the expectation that moquette will release it when it's done with the buffer. This is the same as what Netty does when it passes a buffer to moquette, or when moquette passes a buffer back to Netty.
No one expects having to release a buffer after passing said buffer on.

broker/src/main/java/io/moquette/broker/Session.java
hylkevds: Moved the discussion to #599

broker/src/main/java/io/moquette/broker/Session.java
hylkevds: Same as above. If something is removed from the inflightWindow, the buffer must be released. This was the main cause for the memory leaks.

broker/src/main/java/io/moquette/broker/MQTTConnection.java
hylkevds: That works, as long as Session.receivedPublishQos2 is not called from other places.
Though I don't think it would make the code easier to read. `MQTTConnectio.processPublish()` is making the two passes, so it is also responsible for retaining the buffer. Otherwise `Session.receivedPublishQos2()` has two retains, even though it only passes the buffer one extra time.

The easiest way to get the buffer handling consistent is to stick to the simple rule:
If a method makes or receives a buffer, it must
- retain it for every call that passes it on
- release it before returning

By sticking to this rule for every method that does anything with buffers, you never have to check what the receiving methods do with the buffer to decide if you need to add a retain or release, making memory leaks a lot easier to find.
In practice, one retain cancels out the release, so for methods that pass the buffer only once, nothing needs to be done.

andsel: One thing to reason about is that `msg.retain()` done in `Session.receivedPublishQos2` (the first step) is essentially another retain of the payload, so this is mainly a double retain.
 
I don't think that a method that receives a buffer has retain it for every method call where it passes it, and release before the return. I think that every method that need to work with a buffer has to retain it. 
The responsibility of retaining a buffer is of the method that works on it, not on the caller's hierarchy

broker/src/main/java/io/moquette/broker/Server.java
andsel: when the buffer is passed to moquette, the responsibility passes to moquette, so in this case it' moquette code that needs to `retain` and `release`, it needs to do it. In this the `retain` and `release` of the buffer should be responsibility of the `internalPublish`, this means the code become something like:

```java
public void internalPublish(MqttPublishMessage msg, final String clientId) {
        final int messageID = msg.variableHeader().packetId();
        if (!initialized) {
            LOG.error("Moquette is not started, internal message cannot be published. CId: {}, messageId: {}", clientId,
                      messageID);
            throw new IllegalStateException("Can't publish on a integration is not yet started");
        }
        LOG.trace("Internal publishing message CId: {}, messageId: {}", clientId, messageID);
   ---> msg.payload().retain();
        dispatcher.internalPublish(msg);
        msg.payload().release();
    }
```

`dispatcher.internalPublish(msg);` manages its `retains`, so this method simply does 
```
refCnt +1
dispatcher.internalPublish(msg);
refCnt - 1 
```

which is equivalent to don't do anything, and we could remove completely the retain and release

andsel (commented):
Hi @hylkevds thanks for the effort you are putting in this.
I have some comments on your suggestions, in the intent to simplify the management of retain

hylkevds:
You are correct that:
```
refCnt +1
dispatcher.internalPublish(msg);
refCnt - 1 
```
is the same as not doing any retains/releases. Hence my statement _In practice, one retain cancels out the release, so for methods that pass the buffer only once, nothing needs to be done._ 

In the end it comes down to maintainability. How easy is it to see if a method, viewed on its own, handles buffers correctly. It is very hard to see from just the name of a message call, if that call will consume the buffer or not, since the name does not tell the viewer if the call just passes the buffer on within moquette, or if it is actually a call to an external library. As a result, when reviewing code, the reviewer has to check for every method call where it goes to see if a retain is needed.

andsel:
Correct, I'm thing to best ways to avoid instructions that are not strictly correlated to a method logic, such those retain before passing a buffer to a method call, and release after that method returned, because I think it make the code little bit hard to understand.

I thought that if a method `consumes` a buffer we could also impose that before touching it, it  has to `retain` and `release` after, before the return of the method. In this way only the outer "shell method" that received the buffer from Netty (or created it) has the duty to call the `release` to make it clean.

hylkevds:
Yes, it is also possible to only do those retains/releases at the "edges" of Moquette.
Lets see, that would mean:

Release at the end of every method that
- receives data from Netty or other external sources. It's important that those methods are never called from within Moquette. Maybe explicitly naming them would help there. Adding a javadoc comment about it would also help. Renaming `internalPublish` to `directPublish` (or adding a directPublish that releases) may be a good idea to make this clearer.
- creates a buffer.
- calls `remove()` on a queue/buffer. (also, `put()` can remove something!)
- calls 'get()' on a queue/buffer, unless:
  - the value is never passed on, then also no `retain()`
  - it's certain that another thread can not release() the value before the method is done

Retain when
- Handing to external libraries
  - `channel.write()`
  - `channel.writeAndFlush()`
  - others?
- `put()` on a queue/buffer
- `get()` on a queue/buffer, unless
  -  the value is never passed on, then also no `release()`
  - it's certain that another thread can not release() the value before the method is done

I think that's all the cases...

andsel:
> Adding a javadoc comment about it would also help. 

This is good in general, in Javadoc cite the fact that the ByteBuf the method is receiving is going to be released.

> Renaming `internalPublish` to `directPublish`

Yes, I'm in favor of this, maybe it's a separate PR

> calls `remove()` on a queue

In this case depends if the method is passing the `ByteBuf` to a Netty codec, in which case it's Netty that invoke `release()` on it. In the other case if the buffer is processed an no more usable it could be dropped, so it should invoke `release()`

> (also, put() can remove something!)

I think in this case you mean a `put` on an hash map that overwrite the existing value,is it correct?

> (also, put() can remove something!)

Do you  mean a `pop` operation that also remove the Buffer from the head of the queue?

hylkevds:
I forgot the pop(), right.
I did indeed mean put() on a Map that returns a possible previous entry with the same key.
Is there a different check somewhere to see if a client is re-using a message ID?

I've given it a try, based on the branch that has my set of PRs, and it seems quite feasible, the commit is smaller than expected:
https://github.com/FraunhoferIOSB/moquette/commit/c9cbffd66217330469a6d878a66f25f0a7f27cec?branch=c9cbffd66217330469a6d878a66f25f0a7f27cec&diff=split

I did notice that just doing a retain() before sending to Netty is not enough, it actually has to be a call to `retainedDuplicate()`, which is a bit tricky, since the method that does the sending has a generic MqttMessage, which may or may not be a ByteBufHolder...
Currently testing, no obvious memory leaks after 5e6 messages...

By the way, what formatting rules, for which IDE do you use?

andsel:
> Is there a different check somewhere to see if a client is re-using a message ID?

No and actually Moquette doesn't check for duplication of packet identifier, indeed reading
http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718025 a second message with same packet ID is admissible because could be a re-send. In this case also the DUP flag should be true, and actually Moquette doesn't check for uniqueness of packet ID nor it checks for dup flag in case of re-send.

> By the way, what formatting rules, for which IDE do you use?

I use IntelliJIDEA, but never forced styled rules. I tend to follow Sun style guide, but we can enforce a style with a plugin in another PR.

----

This PR is shaping good in the commit https://github.com/FraunhoferIOSB/moquette/commit/c9cbffd66217330469a6d878a66f25f0a7f27cec?branch=c9cbffd66217330469a6d878a66f25f0a7f27cec&diff=split

I'll though if we can add some verification of buffer counting in the existing unit tests, so to don't have a future regression on this.

I think also that a brief description of the policy we are introducing here about the usage of buffers and ref counting could be described in the README.md or a DEVELOPER.md

hylkevds:
I've made a new PR that moves the ByteBuf reference counting to the outer interfaces of Moquette: #600 
That means this PR can be closed.

PR #577, opened 2021-02-18
Fixes the relativePath pointing to the parent pom.

hylkevds:
The relativePath in the parent section of the pom of sub projects should point to the pom, not the directory of the pom.
http://maven.apache.org/ref/3.3.9/maven-model/maven.html#class_parent


andsel:
Looking at the link you posted, it says:
```
The relative path of the parent pom.xml file within the check out. If not specified, it defaults to ../pom.xml
```
so we can cleanly remove the `relativePath` since by default it is `../pom.xml`

hylkevds:
Yes, completely removing it also work.
Personally I prefer explicitly setting it, for clarity, but either way works.

PR #578, opened 2021-02-18
Fixes the exception on DROP TABLE, by adding IF EXISTS

hylkevds:
The DBAuthenticatorTest throws an ugly exception when it tries to drop a non-existing ACCOUNT table.
This can simply be fixed by adding "IF EXISTS" to the query.


andsel (approved):
LGTM

PR #580, opened 2021-02-18
Fixed ServerLowLevel tests not running

hylkevds:
JUnit only runs tests in classes that have a name that ends in Test,
the class in question had a name that ended in Tests, with an added s.
As a result, the tests in this class were never executed. This renames
the class to not have the added s.

### DIFF
--- a/broker/src/test/java/io/moquette/integration/ServerLowlevelMessagesIntegrationTest.java
+++ b/broker/src/test/java/io/moquette/integration/ServerLowlevelMessagesIntegrationTest.java
@@ -40,9 +40,9 @@ import static io.netty.handler.codec.mqtt.MqttConnectReturnCode.*;
 import static java.nio.charset.StandardCharsets.UTF_8;
 import static org.junit.jupiter.api.Assertions.*;
 
-public class ServerLowlevelMessagesIntegrationTests {
+public class ServerLowlevelMessagesIntegrationTest {
 
-    private static final Logger LOG = LoggerFactory.getLogger(ServerLowlevelMessagesIntegrationTests.class);
+    private static final Logger LOG = LoggerFactory.getLogger(ServerLowlevelMessagesIntegrationTest.class);
     static MqttClientPersistence s_dataStore;
     Server m_server;
     Client m_client;
--- a/broker/src/test/java/io/moquette/integration/ServerLowlevelMessagesIntegrationTest.java
+++ b/broker/src/test/java/io/moquette/integration/ServerLowlevelMessagesIntegrationTest.java
@@ -16,6 +16,7 @@
 
 package io.moquette.integration;
 
+import static io.moquette.BrokerConstants.FLIGHT_BEFORE_RESEND_MS;
 import io.moquette.broker.Server;
 import io.moquette.broker.config.IConfig;
 import io.moquette.broker.config.MemoryConfig;
@@ -38,6 +39,7 @@ import java.util.Properties;
 import java.util.concurrent.TimeUnit;
 import static io.netty.handler.codec.mqtt.MqttConnectReturnCode.*;
 import static java.nio.charset.StandardCharsets.UTF_8;
+import java.util.concurrent.atomic.AtomicBoolean;
 import static org.junit.jupiter.api.Assertions.*;
 
 public class ServerLowlevelMessagesIntegrationTest {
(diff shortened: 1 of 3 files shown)


hylkevds:
The first commit fixes the tests not running, the second commit adds a test for #573.
So the builds should all fail after the second commit, and be fixed again by PR #576.

broker/src/test/java/io/moquette/integration/ServerLowlevelMessagesIntegrationTest.java
```diff
@@ -171,4 +173,38 @@ public void testWillMessageIsPublishedOnClientBadDisconnection() throws Interrup
...
+        try {
+            subscriber.connect();
+            publisher.connect();
+
+            AtomicBoolean isFirst = new AtomicBoolean(true);
+            CountDownLatch countDownLatch = new CountDownLatch(2);
+            subscriber.subscribe(topic, 1, (String topic1, org.eclipse.paho.client.mqttv3.MqttMessage message) -> {
+                if (isFirst.getAndSet(false)) {
+                    // wait to trigger resending PUBLISH
+                    TimeUnit.SECONDS.sleep(12);
```
andsel: 12 seconds is more then 2 times of resend interval (https://github.com/moquette-io/moquette/blob/f6f3f6eedbb6b50128a25cb9c031662dd5ad043d/broker/src/main/java/io/moquette/broker/Session.java#L37) this direct linkage with this interval should be more explicit, so:
- move from `private` to `package private` or even `public` that constant and reuse here as `FLIGHT_BEFORE_RESEND_MS * 2`

broker/src/test/java/io/moquette/integration/ServerLowlevelMessagesIntegrationTest.java
```diff
@@ -171,4 +173,38 @@ public void testWillMessageIsPublishedOnClientBadDisconnection() throws Interrup
...
+
+        MqttClient subscriber = new MqttClient("tcp://localhost:1883", "Subscriber");
+        MqttClient publisher = new MqttClient("tcp://localhost:1883", "Publisher");
+
+        try {
+            subscriber.connect();
+            publisher.connect();
+
+            AtomicBoolean isFirst = new AtomicBoolean(true);
+            CountDownLatch countDownLatch = new CountDownLatch(2);
```
andsel: The countdown latch initialized to 2 is redundant. It should be initialized to 1 and `countDown()` only when it reaches the second publish.

It could be done even more straightforward with Awaitility, for example:
```java
AtomicBoolean isFirst = new AtomicBoolean(true);
AtomicBoolean receivedPublish = new AtomicBoolean(false);
subscriber.subscribe(topic, 1, (String topic1, org.eclipse.paho.client.mqttv3.MqttMessage message) -> {
  if (isFirst.getAndSet(false)) {
    TimeUnit.MILLISECONDS.sleep(FLIGHT_BEFORE_RESEND_MS * 2);
  } else {
    receivedPublish.set(true); 
  }
});
publisher.publish(topic, "hello".getBytes(), 1, false);

Awaitility.await()
    .atMost(FLIGHT_BEFORE_RESEND_MS * 4, TimeUnit.MILLISECONDS)
    .pollInterval(FLIGHT_BEFORE_RESEND_MS, TimeUnit.MILLISECONDS)
    .untill(receivedPublish::get, equalTo(true));
   
```

andsel (changes requested):
Hi @hylkevds the test is usefull, I would suggest some changes to be more explicit why some magic numbers are the values they are.

broker/src/test/java/io/moquette/integration/ServerLowlevelMessagesIntegrationTest.java
hylkevds: I agree using the constant is much nicer, but Session would also have to be made public for that to work...
Another option is to make a separate class for constants.
What would you prefer?

broker/src/test/java/io/moquette/integration/ServerLowlevelMessagesIntegrationTest.java
andsel: Is better is we move those two constants into `broker/src/main/java/io/moquette/BrokerConstants.java` wihtou changing visibility of `Session` class

hylkevds: Ok, I'll do that, and clean up the test.

andsel (approved):
LGTM

PR #581, opened 2021-02-22
Use static netty-tcnative

hylkevds:
Ubuntu 20.04 no longer has openssl1.0.0, and thus the dynamically linked netty-tcnative no longer works.


andsel (approved):
Thanks to have pointed it out

PR #586, opened 2021-02-25
Fixes #585 MessageId not limited to 65535

### LINKED ISSUE #585: 0.14 MessageId not limited to 65535
### Expected behavior
Generated message IDs are, as per the MQTT specification, always in the 1 - 65535 range.

### Actual behavior
Message IDs go over 65535.

Once the messages tick over, the ID in the packet loops back to 0, but the ID expected in ACKs is still > 65535, resulting in the messages never being ACKed. The only reason this was not noticeable was the messages were only re-sent once (causing memory leaks)

### Steps to reproduce
Have a stream of QOS 1 or 2 messages go to a client, wait for 65535 messages to pass.

https://github.com/hylkevds/moquetteTests/

### Moquette MQTT version
0.14

hylkevds:
Message ID should always be in the range 1 - 65535. This adds a test, and resets the counter to 1 when needed.

### DIFF
--- a/broker/src/main/java/io/moquette/broker/MQTTConnection.java
+++ b/broker/src/main/java/io/moquette/broker/MQTTConnection.java
@@ -503,7 +503,7 @@ final class MQTTConnection {
     }
 
     int nextPacketId() {
-        return lastPacketId.incrementAndGet();
+        return lastPacketId.updateAndGet(v -> v == 65535 ? 1 : v + 1);
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


broker/src/main/java/io/moquette/broker/MQTTConnection.java
```diff
@@ -503,7 +503,22 @@ public void resendNotAckedPublishes() {
     }
 
     int nextPacketId() {
-        return lastPacketId.incrementAndGet();
+        final int next = lastPacketId.incrementAndGet();
```
andsel: I think we can simplify with 
```java
return lastPacketId.updateAndGet(v -> v == 65_536 ? 1 : v + 1);
```

andsel (commented):
I think we could avoid the `synchronized` and leverage the `updateAndGet` insteaed of `incrementAndGet '

broker/src/main/java/io/moquette/broker/MQTTConnection.java
hylkevds: Yep, good point.
Much nicer :)

hylkevds: Should compare to 65_535, since 65_536 is already invalid...

hylkevds:
Simplified now :)

PR #598, opened 2021-05-08
Removed "duplicate" parameters from receivedPublishQosX methods.

hylkevds:
Removed "duplicate" parameters from receivedPublishQosX methods.

These two methods took both a message and payload + retain parameters, but the payload and retain information is also in the message. This essentially duplicates these fields which is confusing. Several tests passed conflicting values for these parameters.


andsel (approved):
LGTM

PR #599, opened 2021-05-16
Fix/suggestions to PR #576

andsel:
This is a draft PR to share some ideas to be applied in PR #576

### DIFF
--- a/broker/src/main/java/io/moquette/broker/MQTTConnection.java
+++ b/broker/src/main/java/io/moquette/broker/MQTTConnection.java
@@ -18,6 +18,7 @@ package io.moquette.broker;
 import io.moquette.broker.subscriptions.Topic;
 import io.moquette.broker.security.IAuthenticator;
 import io.netty.buffer.ByteBuf;
+import io.netty.buffer.ByteBufHolder;
 import io.netty.channel.Channel;
 import io.netty.channel.ChannelFuture;
 import io.netty.channel.ChannelFutureListener;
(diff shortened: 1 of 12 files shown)


hylkevds:
Just had another look at the MQTT message flow, and I think there is yet another potential problem with the PUB_REC methods. For the PUB_ACK/COMP the fix you just made would be fine.

- If we receive a PUB_REC, we drop the PUBLISH message from the inflightWindow, send a PUB_REL _with the same messageId_, and put it on the inflightWindow.
- If we then receive a second PUB_REC, we drop the PUB_REL, message that we just put in the inflightWindow, and send a new PUB_REL.

Shouldn't we check if the thing in the inflightWindow is actually a PUBLISH message before we drop it in response to a PUB_REC?
On the other hand, this shouldn't happen much, and would only slightly reset the message flow. In response to the first PUB_REL, the client will already have released the message and sent the PUB_COMP. Unless that PUB_COMP happens to arrive right in between us dropping the PUB_REL and generating a new PUB_REL...?

hylkevds:
Thinking about the PUB_REC method a bit more, simply using the method you suggested here is probably the most efficient overall. 

- Adding a fetch + check + remove would mean hitting the map twice for every message, which would be bad.
- Adding remove + check + re-insert is worth considering, but more complexity for something that shouldn't happen in the first place. Are messages on the same connection always handled by the same thread?

andsel:
>  Are messages on the same connection always handled by the same thread?

yes this is a grant from Netty, each Channel is managed always by the same thread, however the same thread manage multiple Channels (it's a Non blocking IO)

regard to your observation on double the reception of multiple PUBREC with same packetID, is possible, and happens when for some reason the PUBREL is slow to reach the client, and the client trigger a timeout so it decides to resend the PUBREC.

Your observation on the deletion check, to avoid the deletion from `inflighWindow` of `PubRelMarker` that's waiting for a PUBCOMP, is correct. The broker should check this and avoid a remove+insert of the `PubRelMarker` marker message.

I'll fix this

PR #600, opened 2021-05-24
Fixed #573 ByteBuf reference counting

hylkevds:
The reworks most of the ByteBuf reference counting by moving all buffer release/retain calls to the outside interfaces of moquette.
This supersedes #576 and is based on the discussion there.

Close #576
Close #599
Fix #573

PR #602, opened 2021-06-05
Test to reproduce buffer leak

andsel:


andsel:
Close because integrated into #600

Issue #589, opened 2021-03-01
https://www.infoq.com/news/2021/02/jfrog-jcenter-bintray-closure/


slavino:
Some other repo will be needed for releases

andsel:
Fixed with commit https://github.com/moquette-io/moquette/commit/3f54b59688d237896145ddfd00b1aaf3afbf0489

Issue #528, opened 2020-05-13
Publish javadoc & sources of 0.13


hylkevds:
It seems something went wrong during the publication of 0.13 to the bintray repository. The sources jar is missing:
https://jcenter.bintray.com/io/moquette/moquette-broker/0.13/

Though I just noticed they are also missing for 0.12.
0.11 does have sources.

andsel:
Fixed with commit https://github.com/moquette-io/moquette/commit/3f54b59688d237896145ddfd00b1aaf3afbf0489

PR #606, opened 2021-06-11
Maven Central publish

andsel:
Add configurations to package/sign and deploy artifacts to Maven Central repository


hylkevds:
Closes #589 
This probably also Closes #528, since Maven Central requires sources and javadoc.

PR #582, opened 2021-02-24
Fix #423 resent msg not added to timeout queue

hylkevds:
Messages that are re-sent were not added to the timeout queue, as a result, if the re-sent failed as well, the message was not resent again, instead remaining on the inflightWindow for ever. By re-adding it to the inflightTimeouts, the message is "kept alive" until sending succeeds.

`
inflightTimeouts.add(new InFlightPacket(notAckPacketId.packetId, FLIGHT_BEFORE_RESEND_MS));
`

The second commit is a slight performance optimisation, doing a get + null check instead of a contains + get

PR #607, opened 2021-06-29
Added a test to verify regressions for not ack-ed resends

andsel:


andsel:
Close because #582 was aligned with this

PR #609, opened 2021-07-20
Added missing release() for message removed from inflightWindow

hylkevds:
When stress testing QoS2 I got another complaint of Netty about leaky buffers, and found a (hopefully) last remaining missing `release()`.

### DIFF
--- a/broker/src/main/java/io/moquette/broker/Session.java
+++ b/broker/src/main/java/io/moquette/broker/Session.java
@@ -191,6 +191,7 @@ class Session {
             LOG.warn("Received a PUBREC with not matching packetId");
             return;
         }
+        removed.release();
         if (removed instanceof SessionRegistry.PubRelMarker) {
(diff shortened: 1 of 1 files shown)


hylkevds:
This should also be added to the (as of yet unreleased) 0.15 branch. Do you cherry-pick the commit, or would you like me to make a PR for that?

On a side-note, when stress testing QoS2, I do see all of the notes/warnings passing by about the out-of-order packets, so we did not add that code for nothing! :)
```
INFO      io.moquette.broker.Session - Received a PUBREC for packetId that was already moved in second step of Qos2
WARN      io.moquette.broker.Session - Received a PUBCOMP with not matching packetId
WARN      io.moquette.broker.Session - Received a PUBREC with not matching packetId
```

andsel (approved):
LGTM, good catch!
Thanks @hylkevds.

I'll merge and create a backport PR, because I need to track it in the Changelog file

PR #612, opened 2021-07-21
Fixed #610: Subscriptions not removed from session on unsubscribe

### LINKED ISSUE #610: Leaking Token/Subscription/Topic instances
Now that the big memory leaks from ByteBuffers seem to be gone, I'm looking into the smaller fish.
There seems to be a leak involving Token/Subscription/Topic instances. 
Below heap dump is after running my stress-test for on hour with:
- 5 publishers,
  - each publishing 500 messages per second,
  - to 50 topics with pattern Datastreams(x)/Observations
- 5 clients,
  - each listening to the same 50 topics
  - every 20 seconds unsubscribing, disconnecting, sleeping 10ms, reconnecting, resubscribing

The heap dump was taken after first stopping the publishers, then stopping the listeners and waiting for about 15 seconds.
The heap dump also only shows io.moquette classes. The non-moquette classes were just the usual suspects.

MemLeak-Token

All in all not a big issue given that it took 10e6 messages and 920 reconnects to get to this point, but I'm still investigating...

hylkevds:
On unsubscribe, subscriptions were not removed from Sessions and duplicate subscriptions did not replace the old subscription in the Session.

By changing Session.subscription from a List to a Set, duplicate subscriptions automatically replace old ones, and removing subscriptions becomes easier.
Also adds a Test.

### DIFF
--- a/broker/src/main/java/io/moquette/broker/PostOffice.java
+++ b/broker/src/main/java/io/moquette/broker/PostOffice.java
@@ -134,6 +134,7 @@ class PostOffice {
 
     public void unsubscribe(List<String> topics, MQTTConnection mqttConnection, int messageId) {
         final String clientID = mqttConnection.getClientId();
+        final Session session = sessionRegistry.retrieve(clientID);
         for (String t : topics) {
             Topic topic = new Topic(t);
             boolean validTopic = topic.isValid();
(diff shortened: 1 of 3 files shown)


andsel (approved):
LGTM

PR #613, opened 2021-07-22
feat: expose netty enabled tls protocols option

jbutler:
By exposing this configurations, customers can choose to disable older
or insecure versions of TLS. For example, with this change, it is
possible to force clients to use TLSv1.2

This was tested by using mosquitto client tools to force a specific TLS version and confirm that older protocols were, indeed, disabled. Though it would be nice to have an explicit test to this. I can look at adding one.

### DIFF
--- a/broker/src/main/java/io/moquette/broker/DefaultMoquetteSslContextCreator.java
+++ b/broker/src/main/java/io/moquette/broker/DefaultMoquetteSslContextCreator.java
@@ -89,6 +89,14 @@ class DefaultMoquetteSslContextCreator implements ISslContextCreator {
             if (Boolean.valueOf(sNeedsClientAuth)) {
                 addClientAuthentication(ks, contextBuilder);
             }
+
+            // if enabled tls protocols are not provided, we use the default
+            String enabledTLSProtocols = props.getProperty(BrokerConstants.NETTY_ENABLED_TLS_PROTOCOLS_PROPERTY_NAME);
+            if (enabledTLSProtocols != null) {
+                LOG.info(String.format("Enabled TLS Protocols: {%s}", enabledTLSProtocols));
+                contextBuilder.protocols(enabledTLSProtocols.split(";"));
+            }
+
             contextBuilder.sslProvider(sslProvider);
             SslContext sslContext = contextBuilder.build();
             LOG.info("The SSL context has been initialized successfully.");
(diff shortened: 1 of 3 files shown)


jbutler:
@andsel any thoughts on this? Our security team requires disabling older TLS versions, and I imaging this will be useful for others as well

broker/src/main/java/io/moquette/broker/DefaultMoquetteSslContextCreator.java
```diff
@@ -89,6 +89,14 @@ public SslContext initSSLContext() {
             if (Boolean.valueOf(sNeedsClientAuth)) {
                 addClientAuthentication(ks, contextBuilder);
             }
+
+            // if enabled tls protocols are not provided, we use the default
+            String enabledTLSProtocols = props.getProperty(BrokerConstants.NETTY_ENABLED_TLS_PROTOCOLS_PROPERTY_NAME);
+            if (enabledTLSProtocols != null) {
+                LOG.info(String.format("Enabled TLS Protocols: {%s}", enabledTLSProtocols));
+                contextBuilder.protocols(enabledTLSProtocols.split(";"));
```
andsel, line 97: I think that since this parameter is passed down to Netty, we should perform some sort of parsing.
For example if the only admitted values are `TLSv1.2` and `TLSv1.3`, we could check the strings written by the user are effectively those two, and if's not warn the user with a log line.
The risk is to make Netty crashing with ugly exception.

andsel (changes requested):
Thanks @jbutler to ping me, I lost it.
The idea seems good I would only add some parameter validation, since it's free text that the user has to type, and typos could happen.

broker/src/main/java/io/moquette/broker/DefaultMoquetteSslContextCreator.java
andsel: @jbutler do you think it's feasible to cover it with an integration test? 
I mean that once `TLSv1.2` is configured a client that tries to connect with `TLSv1.1` receives an error on connetion.

jbutler: Absolutely. I can definitely make those changes. Thanks for taking a look

jbutler:
After doing a little bit of research on allowable TLS protocol strings, I've found that JDK8+ (at least sun and openjdk) disable older / insecure TLS protocols by default. Furthermore, this is also controllable directly via Java system properties.

So, this probably doesn't need to be exposed via Moquette config. I'll go ahead and close this PR for now.

andsel:
Maybe I should add a section on the [guide](https://moquette-io.github.io/moquette/documentation.html) about this

Issue #608, opened 2021-07-20
Dropping session queues leaks netty ByteBufs, resulting in OOM


jbutler:
### Expected behavior
Steps described below should not result in Moquette OOM

### Actual behavior
After a couple hours, Moquette is unable to allocate memory for incoming PUBLISH messages.

```
[nioEventLoopGroup-3-6] ERROR NewNettyMQTTHandler exceptionCaught 93  - Unexpected exception while processing MQTT message. Closing Netty channel. CId=pubclient
io.netty.util.internal.OutOfDirectMemoryError: failed to allocate 16777216 byte(s) of direct memory (used: 4294967296, max: 4294967296)
	at io.netty.util.internal.PlatformDependent.incrementMemoryCounter(PlatformDependent.java:776)
	at io.netty.util.internal.PlatformDependent.allocateDirectNoCleaner(PlatformDependent.java:731)
	at io.netty.buffer.PoolArena$DirectArena.allocateDirect(PoolArena.java:645)
	at io.netty.buffer.PoolArena$DirectArena.newChunk(PoolArena.java:621)
	at io.netty.buffer.PoolArena.allocateNormal(PoolArena.java:204)
	at io.netty.buffer.PoolArena.tcacheAllocateNormal(PoolArena.java:188)
	at io.netty.buffer.PoolArena.allocate(PoolArena.java:138)
	at io.netty.buffer.PoolArena.allocate(PoolArena.java:128)
	at io.netty.buffer.PooledByteBufAllocator.newDirectBuffer(PooledByteBufAllocator.java:378)
	at io.netty.buffer.AbstractByteBufAllocator.directBuffer(AbstractByteBufAllocator.java:187)
	at io.netty.buffer.AbstractByteBufAllocator.directBuffer(AbstractByteBufAllocator.java:178)
	at io.netty.buffer.AbstractByteBufAllocator.ioBuffer(AbstractByteBufAllocator.java:139)
	at io.netty.channel.DefaultMaxMessagesRecvByteBufAllocator$MaxMessageHandle.allocate(DefaultMaxMessagesRecvByteBufAllocator.java:114)
	at io.netty.channel.nio.AbstractNioByteChannel$NioByteUnsafe.read(AbstractNioByteChannel.java:150)
	at io.netty.channel.nio.NioEventLoop.processSelectedKey(NioEventLoop.java:719)
	at io.netty.channel.nio.NioEventLoop.processSelectedKeysOptimized(NioEventLoop.java:655)
[...]

PR #620, opened 2021-08-02
fix: re-factor Session creation logic and fix Session mem leaks

jbutler:
There are a lot of changes in this PR. I can can probably try to tease them apart and submit separate PRs, but everything is related so it may be easier to review with full context.

There are several issues with how the SessionRegistry manages Sessions, and Sessions themselves are not properly cleaned up. See the discussion in #608 

This attempts to solve these issues by refactoring Session management logic. If a Session is being resumed, then the same Session object is used (rather than copying state into a new one). Conversely, clean sessions always result in a brand new Session. This eliminates current issues where resumed Sessions actually drop messages that are in the inflightQueue, and also where clean sessions can still receive messages queued in the old Session.

This change also properly "terminates" a Session so that netty ByteBuf's aren't leaked when Session objects are GC'd. I've added reference counting to the Session object to prevent messages from being queued while a Session is being terminated.

### DIFF
--- a/broker/src/main/java/io/moquette/broker/SessionRegistry.java
+++ b/broker/src/main/java/io/moquette/broker/SessionRegistry.java
@@ -187,7 +187,7 @@ public class SessionRegistry {
 
     private void terminateSession(Session session) {
         unsubscribe(session);
-        // TODO - terminate session
+        session.terminateSession();
     }
 
     private void reactivateSubscriptions(Session session, String username) {
@@ -237,6 +237,7 @@ public class SessionRegistry {
     }
 
     public void remove(Session session) {
+        session.terminateSession();
         pool.remove(session.getClientID(), session);
     }
 
--- a/broker/src/main/java/io/moquette/broker/SessionRegistry.java
+++ b/broker/src/main/java/io/moquette/broker/SessionRegistry.java
@@ -187,7 +187,7 @@ public class SessionRegistry {
 
     private void terminateSession(Session session) {
         unsubscribe(session);
-        session.terminateSession();
+        session.release();
     }
 
     private void reactivateSubscriptions(Session session, String username) {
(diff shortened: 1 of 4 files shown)


broker/src/main/java/io/moquette/broker/SessionRegistry.java
```diff
@@ -114,110 +114,91 @@ public SessionCreationResult(Session session, CreationModeEnum mode, boolean alr
...
-
-                LOG.trace("case 3, oldSession with same CId {} disconnected", clientId);
-                creationResult = new SessionCreationResult(oldSession, CreationModeEnum.REOPEN_EXISTING, true);
+            final boolean connecting = oldSession.markConnecting();
+            if (!connecting) {
+                throw new SessionCorruptedException("old session moved in connected state by other thread");
             }
+            // case 2
+            LOG.trace("case 2, oldSession with same CId {} disconnected", clientId);
+            oldSession.bind(mqttConnection);
```
jbutler, line 184: @hylkevds here's a manifestation of the issue you mentioned. I need to fix this. Two sessions can bind here, but only one will win. Though I think the bug is a little different. Instead of having one "zombie" connection, you'd actually just have two connections acting on the same Session. The two connections could disconnect each other. One could eventually become a zombie.

hylkevds: Yeah, that should also be in a synchronised block if it happens on an existing session.

On a new session it does not have to be synchronised, but only if it happens before the session is registered, since once it's registered another Thread/Connection could intervene.

jbutler: That's a good call out. Only `reopenExistingSession` needs to be synchronized so the penalty is only paid by client resuming Sessions. It makes sense that there would be a tad more overhead there.

In practice, the penalty may even be dwarfed by the time needed to perform the TLS handshake anyway (assuming user has configured that). I may poke around and see if I can find some MQTT CONNECT benchmark that I can get set up quickly. It's possible this is a non-issue.

PR #630, opened 2021-09-15
Fix #629: Adding subscriptions is very slow

### LINKED ISSUE #629: 62w subscriptions cost about 85min
### Expected behavior

We hope load all the data in 2 minutes.

### Actual behavior

62w subscriptions cost about 85min.

### Steps to reproduce

1. Start moquetto 
2. Start the client and subscribe to 62w topic
3. Restart moquetto

image

image

image

image

### Moquette MQTT version

0.15

openjdk version "1.8.0_292"
OpenJDK Runtime Environment (build 1.8.0_292-b10)
OpenJDK 64-Bit Server VM (build 25.292-b10, mixed mode)

Linux MyServerName 3.10.0-1062.18.1.el7.x86_64 #1 SMP Tue Mar 17 23:49:17 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux

hylkevds:
- Sort CNode.children by token
  - made Token comparable
  - made CNode.token final
  - Changed all searches to use Collections.binarySearch()
  - Changed inserts to respect sorting order
- Optimised away all instances that loop over all children
- Optimised away double lookup caused by "contains then get"

### DIFF
--- a/broker/src/main/java/io/moquette/broker/subscriptions/CNode.java
+++ b/broker/src/main/java/io/moquette/broker/subscriptions/CNode.java
@@ -40,22 +40,15 @@ class CNode implements Comparable<CNode> {
         return token;
     }
 
-    boolean anyChildrenMatch(Token token) {
-        int idx = Collections.binarySearch(children, token, (Object node, Object token1) -> ((INode) node).mainNode().token.compareTo((Token) token1));
-        return idx >= 0;
-    }
-
     List<INode> allChildren() {
         return new ArrayList<>(this.children);
     }
 
     INode childOf(Token token) {
         int idx = Collections.binarySearch(children, token, (Object node, Object token1) -> ((INode) node).mainNode().token.compareTo((Token) token1));
-        INode child = children.get(idx);
-        if (child != null) {
-            return child;
-        }
-        throw new IllegalArgumentException("Asked for a token that doesn't exists in any child [" + token + "]");
+        if (idx < 0)
+            return null;
+        return children.get(idx);
     }
 
     private boolean equalsToken(Token token) {
--- a/broker/src/main/java/io/moquette/broker/subscriptions/CNode.java
+++ b/broker/src/main/java/io/moquette/broker/subscriptions/CNode.java
@@ -17,15 +17,16 @@ package io.moquette.broker.subscriptions;
 
 import java.util.*;
 
-class CNode {
+class CNode implements Comparable<CNode> {
 
-    private Token token;
-    private List<INode> children;
+    private final Token token;
+    private final List<INode> children;
     Set<Subscription> subscriptions;
 
-    CNode() {
+    CNode(Token token) {
         this.children = new ArrayList<>();
         this.subscriptions = new HashSet<>();
+        this.token = token;
     }
 
     //Copy constructor
@@ -39,30 +40,20 @@ class CNode {
         return token;
     }
 
-    public void setToken(Token token) {
-        this.token = token;
-    }
-
     boolean anyChildrenMatch(Token token) {
-        for (INode iNode : children) {
-            final CNode child = iNode.mainNode();
-            if (child.equalsToken(token)) {
-                return true;
-            }
-        }
-        return false;
+        int idx = Collections.binarySearch(children, token, (Object node, Object token1) -> ((INode) node).mainNode().token.compareTo((Token) token1));
+        return idx >= 0;
     }
 
     List<INode> allChildren() {
-        return this.children;
+        return new ArrayList<>(this.children);
     }
 
     INode childOf(Token token) {
-        for (INode iNode : children) {
-            final CNode child = iNode.mainNode();
-            if (child.equalsToken(token)) {
-                return iNode;
-            }
+        int idx = Collections.binarySearch(children, token, (Object node, Object token1) -> ((INode) node).mainNode().token.compareTo((Token) token1));
+        INode child = children.get(idx);
+        if (child != null) {
+            return child;
         }
         throw new IllegalArgumentException("Asked for a token that doesn't exists in any child [" + token + "]");
     }
@@ -81,11 +72,17 @@ class CNode {
     }
 
     public void add(INode newINode) {
-        this.children.add(newINode);
+        int idx = Collections.binarySearch(children, token, (Object node, Object token1) -> ((INode) node).mainNode().token.compareTo((Token) token1));
+        if (idx < 0) {
+            children.add(-1 - idx, newINode);
+        } else {
+            children.add(idx, newINode);
+        }
     }
 
     public void remove(INode node) {
-        this.children.remove(node);
+        int idx = Collections.binarySearch(children, node.mainNode().token, (Object node1, Object token1) -> ((INode) node1).mainNode().token.compareTo((Token) token1));
+        this.children.remove(idx);
     }
 
     CNode addSubscription(Subscription newSubscription) {
@@ -135,5 +132,10 @@ class CNode {
             }
         }
         this.subscriptions.removeAll(toRemove);
+    }
+
+    @Override
+    public int compareTo(CNode o) {
+        return token.compareTo(o.token);
     }
 }
--- a/broker/src/main/java/io/moquette/broker/subscriptions/CNode.java
+++ b/broker/src/main/java/io/moquette/broker/subscriptions/CNode.java
@@ -72,7 +72,7 @@ class CNode implements Comparable<CNode> {
     }
 
     public void add(INode newINode) {
-        int idx = Collections.binarySearch(children, token, (Object node, Object token1) -> ((INode) node).mainNode().token.compareTo((Token) token1));
+        int idx = Collections.binarySearch(children, newINode.mainNode().token, (Object node, Object token1) -> ((INode) node).mainNode().token.compareTo((Token) token1));
         if (idx < 0) {
             children.add(-1 - idx, newINode);
         } else {
--- a/broker/src/main/java/io/moquette/broker/subscriptions/CNode.java
+++ b/broker/src/main/java/io/moquette/broker/subscriptions/CNode.java
@@ -45,10 +45,16 @@ class CNode implements Comparable<CNode> {
     }
 
     INode childOf(Token token) {
-        int idx = Collections.binarySearch(children, token, (Object node, Object token1) -> ((INode) node).mainNode().token.compareTo((Token) token1));
-        if (idx < 0)
+        int idx = findIndexForToken(token);
+        if (idx < 0) {
             return null;
+        }
         return children.get(idx);
+    }
+
+    private int findIndexForToken(Token token) {
+        final INode tempTokenNode = new INode(new CNode(token));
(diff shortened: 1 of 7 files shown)


hylkevds:
This not only massively speeds up adding subscriptions, also matching publishes to subscriptions is much more efficient when there are many topics with subscriptions.

broker/src/test/java/io/moquette/integration/PublishToManySubscribersUseCaseTest.java
```diff
@@ -162,7 +162,7 @@ void onePublishTriggerManySubscriptionsNotifications() throws MqttException, Int
     }
 
     private void segmentedParallelSubscriptions(BiConsumer<IMqttAsyncClient, IMqttActionListener> biConsumer) throws InterruptedException {
-        int openSlotCount = COMMAND_QUEUE_SIZE;
```
andsel, line 165: I don't understand the reason why of this

broker/src/main/java/io/moquette/broker/subscriptions/CNode.java
```diff
@@ -39,32 +40,15 @@ public Token getToken() {
...
 
     INode childOf(Token token) {
-        for (INode iNode : children) {
-            final CNode child = iNode.mainNode();
-            if (child.equalsToken(token)) {
-                return iNode;
-            }
-        }
-        throw new IllegalArgumentException("Asked for a token that doesn't exists in any child [" + token + "]");
+        int idx = Collections.binarySearch(children, token, (Object node, Object token1) -> ((INode) node).mainNode().token.compareTo((Token) token1));
```
andsel: All these `binarySearch` uses, forces casting, passing from Object, because the `token` is not of the right type.
It's better to pass through the compiler check to avoid runtime casting error at the price of creating a couple of instances, with something like:
```java
final INode tempTokenNode = new INode(new CNode(token));
        int idx = Collections.binarySearch(children, tempTokenNode, (INode node, INode tokenNode) -> node.mainNode().token.compareTo(tokenNode.mainNode().getToken()));
```

broker/src/main/java/io/moquette/broker/subscriptions/CNode.java
```diff
@@ -39,32 +40,15 @@ public Token getToken() {
...
+        if (idx < 0)
+            return null;
```
andsel: In previous code when `childOf` wasn't able to find an INode associated  to a Token it raised an IllegalArgumentException, I don't understand the reason to move to a null return.

broker/src/main/java/io/moquette/broker/subscriptions/CNode.java
```diff
@@ -81,11 +65,17 @@ CNode copy() {
     }
 
     public void add(INode newINode) {
-        this.children.add(newINode);
+        int idx = Collections.binarySearch(children, newINode.mainNode().token, (Object node, Object token1) -> ((INode) node).mainNode().token.compareTo((Token) token1));
```
andsel: All instances of the binarySearch, maybe could be extracted in a helper method, to avoid code duplication.

andsel (changes requested):
I think the idea to switch to binarySearch is good, however, due to the facts that some optimizations are not easy to grasp, maybe each one could be separated in its own PS, describing well the optimization and why.

I would appreciate if the code could be accompanied with a test that with a huge set of sibling subscriptions shows the slowness. For example an insert or a search that fails after 5 minutes. 
In this way just switching the linear scan to binary search shows that time is reduced a lot.

hylkevds:
> I think the idea to switch to binarySearch is good, however, due to the facts that some optimizations are not easy to grasp, maybe each one could be separated in its own PS, describing well the optimization and why.

That's what the separate commits are for. Separating those into their own PRs makes no sense, since they don't work separately. Just go through the separate commits. You can even check them out individually with GIT to try them out locally.

> I would appreciate if the code could be accompanied with a test that with a huge set of sibling subscriptions shows the slowness. For example an insert or a search that fails after 5 minutes. In this way just switching the linear scan to binary search shows that time is reduced a lot.

That should be possible :)

broker/src/test/java/io/moquette/integration/PublishToManySubscribersUseCaseTest.java
hylkevds: Timing...
Because the code is now faster, it can flood the queue since each action generates more than one command on the Queue.
It used to work because the slowness of the insert delayed the creation of commands enough for the worker threads to catch up.

It surprised me too, but you should try it out for yourself :)

broker/src/main/java/io/moquette/broker/subscriptions/CNode.java
hylkevds: That's part of the "optimised away double lookup" commit. Often the code did a `anyChildrenMatch()` followed by a `childOf()` meaning it did the expensive lookup twice. That's really inefficient of course. So to remove this superfluous `anyChildrenMatch()` call, we make `childOf()` return null if there is no such child. We could do an `Optional`, but since it's not a public interface that seemed redundant.

hylkevds:
> I would appreciate if the code could be accompanied with a test that with a huge set of sibling subscriptions shows the slowness. For example an insert or a search that fails after 5 minutes.

I've added two tests that together run in ~4 seconds with the new code on my machine, but time out after 5 minutes with the old code.

broker/src/main/java/io/moquette/broker/subscriptions/CNode.java
hylkevds: Good point. I've created a method for the binarySearch and cleaned up the types.

broker/src/main/java/io/moquette/broker/subscriptions/CNode.java
hylkevds: Good idea. Done :)

broker/src/main/java/io/moquette/broker/subscriptions/CNode.java
andsel: Going step by step is much easier and now I can see the motivation of the changes. Maybe a more descriptive comment for the "Optimised away all instances that loop over all children" case. Something like:
```
Not all next level children has to be verified recursively, but only the multi level any (#) the single level one (+) or the only matching node with the next head token part.
```

broker/src/main/java/io/moquette/broker/subscriptions/CNode.java
```diff
@@ -39,32 +40,21 @@ public Token getToken() {
...
 
     INode childOf(Token token) {
-        for (INode iNode : children) {
-            final CNode child = iNode.mainNode();
-            if (child.equalsToken(token)) {
-                return iNode;
-            }
+        int idx = findIndexForToken(token);
+        if (idx < 0) {
+            return null;
```
andsel: I think would better to be more explicit. `null`is speedy but could drive us to the NPE. I think would be better to explicitly return an `Optional`, it's designed for cases such this.

broker/src/test/java/io/moquette/broker/subscriptions/CTrieSpeedTest.java
```diff
@@ -0,0 +1,118 @@
+/*
+ * Copyright (c) 2012-2018 The original author or authors
```
andsel:
```suggestion
 * Copyright (c) 2012-2023 The original author or authors
```

broker/src/test/java/io/moquette/broker/subscriptions/CTrieSpeedTest.java
```diff
@@ -0,0 +1,118 @@
...
+    public List<Subscription> prepareSubscriptionsFlat() {
+        List<Subscription> results = new ArrayList<>(TOTAL_SUBSCRIPTIONS);
+        int count = 0;
+        long start = System.currentTimeMillis();
+        for (int b = 0; b < TOTAL_SUBSCRIPTIONS / 10; b++) {
+            for (int a = 0; a < 10; a++) {
+                count++;
+                results.add(clientSubOnTopic("Client-" + a, "mainTopic-" + b));
+            }
+        }
```
andsel:
```suggestion
        for (int v = 0; topicIdx < TOTAL_SUBSCRIPTIONS / 10; topicIdx++) {
            for (int clientIdx = 0; clientIdx < 10; clientIdx++) {
                count++;
                results.add(clientSubOnTopic("Client-" + clientIdx, "mainTopic-" + topicIdx));
            }
        }
```

Instead of `a` and `b`, using something more expressive would help readability

broker/src/test/java/io/moquette/broker/subscriptions/CTrieSpeedTest.java
```diff
@@ -0,0 +1,118 @@
...
+                    for (int d = 0; d < countPerLevel; d++) {
+                        count++;
+                        results.add(clientSubOnTopic("Client-" + a, "mainTopic-" + b + "/subTopic-" + c + "/subSubTopic" + d));
+                        if (count >= TOTAL_SUBSCRIPTIONS) {
+                            break outerloop;
+                        }
+                    }
+                }
+            }
+        }
```
andsel:
```suggestion
        for (int clientIdx = 0; clientIdx < countPerLevel; clientIdx++) {
            for (int rootLayerIdx = 0; rootLayerIdx < countPerLevel; rootLayerIdx++) {
                for (int firstLayerIdx = 0; firstLayerIdx < countPerLevel; firstLayerIdx++) {
                    for (int secondLayerIdx = 0; secondLayerIdx < countPerLevel; secondLayerIdx++) {
                        count++;
                        results.add(clientSubOnTopic("Client-" + clientIdx, "mainTopic-" + rootLayerIdx + "/subTopic-" + firstLayerIdx + "/subSubTopic" + secondLayerIdx));
                        if (count >= TOTAL_SUBSCRIPTIONS) {
                            break outerloop;
                        }
                    }
                }
            }
        }
```

The same spirit could be applied here, with better naming could increase readability

andsel (changes requested):
Thank's @hylkevds  for your patience.  The PR is pretty good, just a couple of final touches.

broker/src/main/java/io/moquette/broker/subscriptions/CNode.java
hylkevds: Sure, I can update the commit message with a longer explanation.

broker/src/test/java/io/moquette/broker/subscriptions/CTrieSpeedTest.java
hylkevds: Good point, I've updated it.

In FROST-Server we're now using spotless-maven-plugin to automatically check the licence header and code formatting. When something isn't how it should be it can also automatically apply the corrections.

broker/src/main/java/io/moquette/broker/subscriptions/CNode.java
hylkevds: No Problem, updated!

broker/src/test/java/io/moquette/broker/subscriptions/CTrieSpeedTest.java
hylkevds: Sure, no problem. Also for the other method.

andsel (approved):
Great work LGTM!

PR #631, opened 2021-09-15
Adds session events processors and queues of commands

andsel:
## Release notes
Fix: introduces sessions event processors to segregate changes to a session in one single thread, simplifying concurrency and code

## What does this PR do?
Introduces the concept of sticky sessions, all the change operations to a session is done always by the same thread, so this simplifies the concurrency access, segregating the access to only one thread. The same thread can handle multiple sessions.

Every session event processor is an event loop, that peeks commands out of a queue and executes them.
This means that there is an isolation queue between the IO threads that executes the lower layer of protocol  and the event loops that mutate and managed the upper logic, mainly bound to Session's related structures.

The length of those queue can be configured with option `session_queue_size`, by default is has 1024 slots, but it could be that if there is an heavy publisher, the queue becomes full and this case the only thing to do is log an error. So it could happen that for a SUB command no SUBACK is sent (the same is valid for PUB) if a queue is full. In MQTT 3.1.1 there is no timeout specification for missed SUBACK, so the client will not send a second SUB (dup=1) and we potentially could loose SUBscriptions (and PUBlishes).

This commit also implement a couple of optimizations on this event loop side:
- if a command processed by an EventLoop generated the sending of other commands to the same EventLoop, then those commands are inlined and doesn't pass through the queue.
- when a single PUB command involves many subscribers, then all the subsequent PUB that goes to the same queue are compressed on only one command on the queue, instead of insert many single PUB commands

## Related issues

- Relates  #608

### DIFF
--- a/broker/src/main/java/io/moquette/broker/PostOffice.java
+++ b/broker/src/main/java/io/moquette/broker/PostOffice.java
@@ -365,8 +365,8 @@ class PostOffice {
     }
 
     private static class SubQosCombo {
-        public final Subscription sub;
-        public final MqttQoS qos;
+        final Subscription sub;
+        final MqttQoS qos;
 
         public SubQosCombo(Subscription sub, MqttQoS qos) {
             this.sub = sub;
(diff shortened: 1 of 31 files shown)


broker/src/main/java/io/moquette/broker/PostOffice.java
```diff
@@ -158,102 +292,225 @@ public void unsubscribe(List<String> topics, MQTTConnection mqttConnection, int
...
             LOG.error("MQTT client: {} is not authorized to publish on topic: {}", clientId, topic);
-            return;
+            return CompletableFuture.completedFuture(null);
         }
 
         ByteBuf payload = msg.payload();
-        publish2Subscribers(payload, topic, AT_LEAST_ONCE);
+        final CompletableFuture<RoutingResults> publishFuture;
+        if (msg.fixedHeader().isDup()) {
+            final Set<String> failedClients = failedPublishes.listFailed(clientId, messageID);
```
hylkevds, line 342: I think this can return an empty Set if a client sends a dup on an empty failedPublishes. In that case all clients get the dup message. Is that correct behavior?

andsel: You are right an it's not the correct behavior.  Fixed with b2c08ddbf6c18d4cee038b936d63308947507503 where the cases  "no filter to apply" and "no failed publish present" is distinguished by a constant instance of a Set.
I thought about using an Optional parameter, but seems more convoluted than the self explaining constant

PR #635, opened 2021-11-29
Moquette broker refactor

kazetsukaimiko:
h2-mvstore on the classpath caused me some issues running the broker from Quarkus. Workaround:
```xml
    <dependency>
      <artifactId>moquette-broker</artifactId>
      <version>0.15</version>
      <exclusions>
        <!-- to avoid a CHECK2 NoSuchFieldError -->
        <exclusion>
          <groupId>com.h2database</groupId>
          <artifactId>h2-mvstore</artifactId>
        </exclusion>
      </exclusions>
    </dependency>
```

So I dove into the moquette sources a bit. Initially it seemed like bumping the h2 version would resolve this, but on deeper look there are some pretty nice interfaces present in moquette (IQueueRepository, ISubscriptionsRepository, IRetainedRepository) that seem like they could enable moquette to make a clean divide between the broker and the H2 implementations:

This is a large body of work I don't expect to be merged, but I wanted to get an idea of the maintainers' openness to such a move, and any alternative approaches they would take to get there. I'll also note that this work is also incomplete- the goal would be the removal of all H2 dependencies from `moquette-broker`, moving the current H2 persistence backends to their own module. 

```xml
    <dependency>
      <artifactId>moquette-broker</artifactId>
      <version>0.15</version>
    </dependency>

    <!-- H2 implementation classes -->
    <dependency>
      <artifactId>moquette-repository-h2</artifactId>
      <version>0.15</version>
    </dependency>

    <!-- Redis implementation classes -->
    <dependency>
      <artifactId>moquette-repository-redis</artifactId>
      <version>0.15</version>
    </dependency>

    <!-- JSR-107 implementation classes -->
    <dependency>
      <artifactId>moquette-repository-jcache</artifactId>
      <version>0.15</version>
    </dependency>
```
There are numerous advantages to this from an embedder's perspective, particularly in an EE context.

Branch builds and tests pass, although ErrorProne is disabled- moquette does not build with vanilla JDK11 / mvn clean package with that present. The README is outdated, referring to gradle wrappers that seemingly no longer exist.


andsel:
Thanks @kazetsukaimiko to throw the ball for discussion.
Those interfaces was created because Moquette initially used `MapDB` as local KV storage, then switched to `H2 MVStore` because H2 is more widely used and probably the code was more tested and maintained.
So those interfaces was created to simplify the fix and the life when there were both implementation available, was never meant as a real API to be externally implemented.
More on this, I plan to rewrite the queue storage using memory mapped files and also the storage of subscriptions and other metadata, so that Moquette is self contained in terms of dependencies, or at least is very minimal.

However is legit to provide an SPI (service provider interface) to offer extension hooks for some of the functionalities, but those has to be carefully crafted because "once you publish an interface you have to support it for the rest of the project lifetime" and this could potentially put some constraints into the Moquette evolution

kazetsukaimiko:
@andsel Everything you've said is essentially correct/I agree with.

The overall theme of this pull is, I do think it could be nice if one could swap out the storage backend for an alternative. Some ideas:

Its possible you could externalize some of your concerns around providing an SPI by piggybacking on existing known interfaces, something like the JSR-107 API or just `java.lang.Map` in this case perhaps? Seems like you'd want something that can spawn maps/caches by name and handle events around map keys.

My opinion on SPI stability... as long as you keep any API breaking changes to major versions and document the steps needed to migrate, I think you've done your part to keep things stable. Many projects seem to confuse stability with stagnation this way, refusing to break or change anything, ever. SPIs can change- I wouldn't want to use Hibernate 3.x or 4.x interfaces today, the modern equivalents have significant improvements in simplicity and function (they've externalized most concerns to JPA as well). I've done 3->4 and 3->5 migrations before, there's always a level of pain.

Nitpick: If the interfaces you mention are not meant to be used externally, I'd suggest reducing their visibility to the maximum extent possible. Anything that is an interface + is public is a hint to a developer "this is what you should be using".

Thanks again for taking a look at this.

andsel:
> Its possible you could externalize some of your concerns around providing an SPI by piggybacking on existing known interfaces, something like the JSR-107 API or just java.lang.Map in this case perhaps? 

The problem is not providing a set interfaces, the problem id design those interfaces in a way that's future proof, at least for a couple of releases.
Using existing public interfaces (like Map or JCache JSR-107) is not a good thing, because those interrfaces exposes the concept of  an associative table, and probably that abstraction doesn't fit well in the shape needed by the broker.
I'm not saying it's difficult, but only need time to  define an curve it. 
Maybe you have an idea for this?

kazetsukaimiko:
> The problem is not providing a set interfaces, the problem id design those interfaces in a way that's future proof, at least for a couple of releases. ...
> Maybe you have an idea for this?

@andsel Sorry about the late reply, I've only recently been able to come back to the project implementing moquette (found a new bug).

A blind guess is a set of interfaces much like `InterceptHandler`.... a `PersistenceHandler` of sorts, where you have a bunch of `InterceptMessage`-likes that all extend / implement `PersistenceTarget` or something similar is going to be close.

To give you a better answer maybe you could tell me what things need persistence under what conditions, and how they need to be looked up. Either way this PR should get closed.

kazetsukaimiko:
Closing this.

PR #637, opened 2022-01-01
Feature/intro session events processors fixes

hylkevds:
After quite a bit more testing I found some loose threads. Most I've fixed prototypically in this PR:

* Disconnects must be handled on the sessionExecutors Threads, otherwise the session create logic breaks.
* Queues were never removed from QueueRepositories, not even in case of cleanSession.
* Queues in the Repositories were always reused, even in case of cleanSession.
* Queues were in some cases removed from the local queue cache, while in-use.
* Messages on queues and in Sessions were not released when discarded.
* Session.mqttConnection can be set to null with messages still in the queue.
* Processor threads were not stopped on server exit.

With the fixes in this PR memory use is quite constant, and Netty doesn't complain. The Unit Tests do cause Netty to complain about unreleased buffers, but I've not found the cause yet.

I'm not sure the failedPublishes collection is a good way to solve command queues being full. It moves the problem from a fixed size queue to a limitless queue. Especially since the only reason for a failure at this point is a full command queue, and putting the message in the FailedPublishCollection does not reduce the load (it actually increases load). Blocking until the command queue has space is probably more effective. We just have to make sure no messages are added to the command queues by the sessionExecutors Threads.

One last important ToDo is Persistent Sessions. They are never removed, gathering messages, even when the client never reconnects. We'll need to figure a way to add a configurable timeout for persistent sessions...

### DIFF
--- a/broker/src/main/java/io/moquette/broker/PostOffice.java
+++ b/broker/src/main/java/io/moquette/broker/PostOffice.java
@@ -365,8 +365,8 @@ class PostOffice {
     }
 
     private static class SubQosCombo {
-        public final Subscription sub;
-        public final MqttQoS qos;
+        final Subscription sub;
+        final MqttQoS qos;
 
         public SubQosCombo(Subscription sub, MqttQoS qos) {
             this.sub = sub;
@@ -374,17 +374,17 @@ class PostOffice {
         }
     }
 
-    private static class PublishListCmd implements  Callable<String> {
+    private static class BatchPublishCommand implements Callable<String> {
 
-        public final String eventLoopNr;
-        public final ByteBuf payload;
-        public final Topic topic;
-        public final List<SubQosCombo> subscriptions = new ArrayList<>();
+        final String firstClientId;
+        final ByteBuf payload;
+        final Topic topic;
+        final List<SubQosCombo> subscriptions = new ArrayList<>();
         private final PostOffice postOffice;
 
-        public PublishListCmd(PostOffice postOffice, String eventLoopNr, ByteBuf payload, Topic topic) {
+        public BatchPublishCommand(PostOffice postOffice, String firstClientId, ByteBuf payload, Topic topic) {
             this.postOffice = postOffice;
-            this.eventLoopNr = eventLoopNr;
+            this.firstClientId = firstClientId;
             this.payload = payload;
             this.topic = topic;
         }
@@ -398,7 +398,7 @@ class PostOffice {
             for (SubQosCombo subQos:subscriptions) {
                 postOffice.publishToSession(payload, topic, subQos.sub, subQos.qos);
             }
-            return eventLoopNr;
+            return firstClientId;
         }
     }
 
@@ -407,7 +407,7 @@ class PostOffice {
         Set<Subscription> topicMatchingSubscriptions = subscriptions.matchQosSharpening(topic);
         List<CompletableFuture<RouteResult>> publishFutures = new ArrayList<>(topicMatchingSubscriptions.size());
 
-        final PublishListCmd[] subCommands = new PublishListCmd[eventLoops];
+        final BatchPublishCommand[] subCommands = new BatchPublishCommand[eventLoops];
 
         for (final Subscription sub : topicMatchingSubscriptions) {
             MqttQoS qos = lowerQosToTheSubscriptionDesired(sub, publishingQos);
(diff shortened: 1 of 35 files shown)


broker/src/main/java/io/moquette/broker/Session.java
```diff
@@ -215,7 +215,10 @@ public void processPubRec(int pubRecPacketId) {
         inflightWindow.put(pubRecPacketId, new SessionRegistry.PubRelMarker());
         inflightTimeouts.add(new InFlightPacket(pubRecPacketId, FLIGHT_BEFORE_RESEND_MS));
         MqttMessage pubRel = MQTTConnection.pubrel(pubRecPacketId);
-        mqttConnection.sendIfWritableElseDrop(pubRel);
+        final MQTTConnection myMqttConnection = mqttConnection;
```
andsel: It's not clean why copy the field in local variable before the check.
Also it's not clear to me why it becomes null. The `null` assignment is done at  https://github.com/moquette-io/moquette/blob/1b1bac76fa9be9e9bb0b28be3fs c627afee684d73/broker/src/main/java/io/moquette/broker/Session.java#L181 during `disconnect` of the `Session`.  
So the NPE here could happen only if a client disconnected and the send PUB QoS2 which is a weird behavior that shouldn't happen because the DISCONNECT and the PUB are handled serially by the same thread.

broker/src/main/java/io/moquette/broker/Session.java
```diff
@@ -261,9 +264,10 @@ private void sendPublishQos1(Topic topic, MqttQoS qos, ByteBuf payload) {
             return;
         }
 
-        if (canSkipQueue()) {
+        final MQTTConnection myMqttConnection = mqttConnection;
+        if (myMqttConnection != null && canSkipQueue()) {
```
andsel: Here the null check is correct because the PUB is processed by a different thread that could have executed the DISCONNECT.
I would avoid to name the local reference to the mqttConnection as `myMqttConnection`, something like `localMqttConnectionReference` or similar (have not better ideas).
The same for the `sendPublishQos2` method

andsel (changes requested):
Hi @hylkevds left some notes as discussed at https://github.com/moquette-io/moquette/issues/608#issuecomment-1004198215

broker/src/main/java/io/moquette/broker/Session.java
hylkevds: Indeed, I removed this instance of the check.

broker/src/main/java/io/moquette/broker/Session.java
hylkevds: Ok, I've renamed the local variables.

broker/src/main/java/io/moquette/broker/PostOffice.java
```diff
@@ -365,34 +364,85 @@ private void manageRetain(Topic topic, MqttPublishMessage msg) {
...
+                if (subCommands[targetQueueId] == null) {
+                    subCommands[targetQueueId] = new PublishListCmd(this, Integer.toString(targetQueueId), payload, topic);
+                }
+                subCommands[targetQueueId].add(new SubQosCombo(sub, qos));
+            }
+        }
+
+        for (int i = 0; i < eventLoops; i++) {
+            if (subCommands[i] != null) {
+                publishFutures.add(routeCommand(subCommands[i].eventLoopNr, subCommands[i]));
```
andsel: I think that here there is the risk to send the bulk publish command to an unrelated session queue.

The target queueId is determined by 
```java
Math.abs(sub.getClientId().hashCode()) % this.eventLoops
```
but if we used that id, convert to string and route it again I don't thing that a second execution of the formula gives same result.

Support the `clientId` is `subscriber_0` the `targetQueueId` after first application of the modulo is `2`.
If we convert the integer to to the string `"2"` and re-apply the formula on it, we could have 1 as result.

So  it goes to be processed by the incorrect Session Event Loop.

andsel (commented):
@hylkevds thanks for sahring this, I brought this commit to PR #631 with some little changes

broker/src/main/java/io/moquette/broker/PostOffice.java
hylkevds: Yeah, that's not quite correct yet. The reason for passing the targetQueue number is to make it easier to find out which of the queues was the one that caused a failure if one of the futures returned a failure.
Instead we should just pass the first clientId that we encountered for that targetQueue... Or add a routeComand that accepts a queueId instead of a clientId?

andsel: I used the first encountered clientId.
To identify which queue has the problem, given the clientId it's always possible to recalculate it.

andsel:
#631 has been merged, so I think this PR could be closed.

hylkevds:
This PR is no longer relevant.

PR #638, opened 2022-01-03
Implements concurrent segmented queues

andsel:
## Release notes
Introduce new persistent queue implementation based on concurrent access to segmented queues.

## What does this PR do?

All queues are saved in segments, each segment in stored in a memory mapped region named page where each page can contains multiple segments of different queues. The access to the segment is based on CAS operator on such named `SegmentPointer` objects, when the CAS is not possible a retry is repeated until new `Segment` access is required, in which case the access is managed by a usual lock guarded section

### DIFF
--- a/.github/workflows/maven_central_release_version.yml
+++ b/.github/workflows/maven_central_release_version.yml
@@ -8,11 +8,7 @@ name: Build and ship to Maven Central
 #            - "v*"
 on:
     release:
-<<<<<<< HEAD
         types: [released]
-=======
-        types: [created]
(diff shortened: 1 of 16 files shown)


hylkevds:
Is there a way to discard/close a queue? For example when the client reconnects with cleanSession=true?

andsel:
@hylkevds good to remind it, I haven't yet thought about :-) 
I'm working on your suggestion to reuse already consumed segments in pages.
I think that the close/cleanup of a queue it's easy, simply consider the segments that composed the queue as "available for reuse". Whatever is not listed in the checkpoint files is garbage.

andsel:
Superseeded by #691

Issue #641, opened 2022-01-18
Moquette broker kills parent application if embedded and broker port already used


mkokoszka-ms:
### Expected behavior
When moquette is embedded into  other application and defined broker port (e.g. 1833) is already used, server should not start and log message should be generated.
### Actual behavior
When port is already used by something else, moquette broker generates log message "io.moquette.broker.NewNettyAcceptor initFactory Cannot bind to port: 1883". Parent application which uses moquetted is killed by System.exit(1);
Problematic System.exit(1) is here: io.moquette.broker.NewNettyAcceptor.initFactory()
### Steps to reproduce
Start any MQTT Broker on port 1833 and then start moquette broker on the same port.

### Moquette MQTT version
moquette-broker-0.15
### JVM version (e.g. `java -version`)
not relevant
### OS version (e.g. `uname -a`)
not relevant

sbaranov-iqmessenger:
yeah, I encountered the same issue. It would be better to remove `System.exit()` from `NewNettyAcceptor`

PR #646, opened 2022-02-01
Moquette broker kills parent application if embedded and broker port already used #641

sbaranov-iqmessenger:
## Release notes
Avoid Server.startServer to exit the process it port is bound, just raise an error (#646)

[#641](https://github.com/moquette-io/moquette/issues/641)
just there was removed System.exit(1)

### DIFF
--- a/broker/src/main/java/io/moquette/broker/NewNettyAcceptor.java
+++ b/broker/src/main/java/io/moquette/broker/NewNettyAcceptor.java
@@ -223,7 +223,7 @@ class NewNettyAcceptor {
         } catch (Exception ex) {
             if (ex instanceof BindException) {
                LOG.error("Cannot bind to port: " + port, ex);
-               System.exit(1);
+                throw new RuntimeException(ex);
             } else {
                 LOG.error("An interruptedException was caught while initializing integration. Protocol={}", protocol, ex);
                 throw new RuntimeException(ex);
--- a/broker/src/main/java/io/moquette/broker/NewNettyAcceptor.java
+++ b/broker/src/main/java/io/moquette/broker/NewNettyAcceptor.java
@@ -223,7 +223,7 @@ class NewNettyAcceptor {
         } catch (Exception ex) {
             if (ex instanceof BindException) {
                LOG.error("Cannot bind to port: " + port, ex);
-                throw new RuntimeException(ex);
+               throw new RuntimeException("Cannot bind to port: " + port, ex);
             } else {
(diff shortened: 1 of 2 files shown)


andsel:
The `io.moquette.broker.Server` class is also used in `bat` and `sh` scripts used by the tar.gz distribution of Moquette. To complete this PR we should create another class (maybe `io.moquette.broker.Launcher` that simply wrap `Server` and catch for the exceptions, because in case of bind error, it should exit with error code without ugly traces in the console.

sbaranov-iqmessenger:
> The `io.moquette.broker.Server` class is also used in `bat` and `sh` scripts used by the tar.gz distribution of Moquette. To complete this PR we should create another class (maybe `io.moquette.broker.Launcher` that simply wrap `Server` and catch for the exceptions, because in case of bind error, it should exit with error code without ugly traces in the console.

I didn't change current exception "culture" and don't throw BindException to not add additional exceptions in method signatures. What do you think ?

broker/src/main/java/io/moquette/broker/NewNettyAcceptor.java
```diff
@@ -223,7 +223,7 @@ public void initChannel(SocketChannel ch) throws Exception {
         } catch (Exception ex) {
             if (ex instanceof BindException) {
                LOG.error("Cannot bind to port: " + port, ex);
-               System.exit(1);
+                throw new RuntimeException(ex);
```
andsel:
```suggestion
               throw new RuntimeException("Cannot bind to port: " + port, ex);
```

andsel (changes requested):
We are almost there, please integrate the suggestion and we are good to merge.

sbaranov-iqmessenger:
> We are almost there, please integrate the suggestion and we are good to merge.

I've committed what you suggested but is there any benefit of containing human readable "Cannot bind..." because in `Server.main` I don't log exception. I just call System.exit(1) without any logging

andsel:
> is there any benefit of containing human readable "Cannot bind..."

I think it's useful in the code that use that method, incorporating it as a library, maybe it want to log the cause of the error, if we don't throw a specific error excpetion.

sbaranov-iqmessenger:
> > is there any benefit of containing human readable "Cannot bind..."
> 
> I think it's useful in the code that use that method, incorporating it as a library, maybe it want to log the cause of the error, if we don't throw a specific error excpetion.

oh, yeah, you are right. I got it

andsel:
Thanks a lot @sbaranov-iqmessenger for contributing! Warm welcome

PR #648, opened 2022-02-02
Fix session handling

hylkevds:
See discussion in #608.

- Add a remove method to IQueueRepository, so queues can actually be removed when no longer needed.
- Make the Session the only place where the Queue lives outside of the repository by removing the queues field from SessionRegistry.
- Tie a Queue and a Session in a 1-to-1 way.
- Remove and clean the old Session and Queue before creating a new pair for a ClientId.
- Clean Session and Queue when they are no longer needed.
- In case a new Connection takes over a Session from an existing, connected connection, ensure the old session can not influence the Session any more after the Session was taken over.

### DIFF
--- a/broker/src/main/java/io/moquette/broker/SessionRegistry.java
+++ b/broker/src/main/java/io/moquette/broker/SessionRegistry.java
@@ -27,6 +27,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.net.InetSocketAddress;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Map;
 import java.util.Optional;
@@ -137,24 +138,25 @@ public class SessionRegistry {
             }
         }
         if (!queues.isEmpty()) {
-            // This indicates a bug...
-            LOG.error("Recreating sessions left {} unused queues.", queues.size());
+            LOG.error("Recreating sessions left {} unused queues. This is probably bug. Session IDs: {}", queues.size(), Arrays.toString(queues.keySet().toArray()));
         }
     }
 
     SessionCreationResult createOrReopenSession(MqttConnectMessage msg, String clientId, String username) {
         SessionCreationResult postConnectAction;
-        //final Session newSession = createNewSession(msg, clientId);
         final Session oldSession = retrieve(clientId);
         if (oldSession == null) {
-            // case 1 no old session.
+            // case 1, no existing session with given clientId.
             final Session newSession = createNewSession(msg, clientId);
             postConnectAction = new SessionCreationResult(newSession, CreationModeEnum.CREATED_CLEAN_NEW, false);
 
             // publish the session
             final Session previous = pool.put(clientId, newSession);
             if (previous != null) {
-                LOG.error("Another Thread added a Session for our clientId {}, this is a bug!", clientId);
+                // if this happens mean that another Session Event Loop thread processed a CONNECT message
+                // with the same clientId. This is a bug because all messages for the same clientId should
+                // be handled by the same event loop thread.
+                LOG.error("Another thread added a Session for our clientId {}, this is a bug!", clientId);
             }
 
             LOG.trace("case 1, not existing session with CId {}", clientId);
@@ -179,7 +181,7 @@ public class SessionRegistry {
                 throw new SessionCorruptedException("old session has already changed state");
             }
 
-            // case 2
+            // case 2, reopening existing session but with cleanSession true
             // publish new session
             unsubscribe(oldSession);
             remove(clientId);
@@ -193,7 +195,7 @@ public class SessionRegistry {
             if (!connecting) {
                 throw new SessionCorruptedException("old session moved in connected state by other thread");
             }
-            // case 3
+            // case 3, reopening existing session without cleanSession, so keep the existing subscriptions
             copySessionConfig(msg, oldSession);
             reactivateSubscriptions(oldSession, username);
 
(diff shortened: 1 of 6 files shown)


broker/src/main/java/io/moquette/broker/SessionRegistry.java
```diff
@@ -116,106 +116,89 @@ public SessionCreationResult(Session session, CreationModeEnum mode, boolean alr
...
-            final Queue<EnqueuedMessage> persistentQueue = queues.get(clientId);
+            final Queue<EnqueuedMessage> persistentQueue = queues.remove(clientId);
             if (persistentQueue != null) {
                 Session rehydrated = new Session(clientId, false, persistentQueue);
                 pool.put(clientId, rehydrated);
             }
         }
+        if (!queues.isEmpty()) {
+            // This indicates a bug...
+            LOG.error("Recreating sessions left {} unused queues.", queues.size());
```
andsel: I would remove the comment and update the comment to indicate it's a bug:
```
Recreating sessions left {} unused queues. This indicates a BUG!
```

It should also list all the clientIds left in the `queues` Map.

broker/src/main/java/io/moquette/broker/SessionRegistry.java
```diff
@@ -116,106 +116,89 @@ public SessionCreationResult(Session session, CreationModeEnum mode, boolean alr
...
+        }
     }
 
     SessionCreationResult createOrReopenSession(MqttConnectMessage msg, String clientId, String username) {
         SessionCreationResult postConnectAction;
-        final Session newSession = createNewSession(msg, clientId);
-        final Session oldSession = pool.get(clientId);
+        //final Session newSession = createNewSession(msg, clientId);
```
andsel:
```suggestion
```

broker/src/main/java/io/moquette/broker/SessionRegistry.java
```diff
@@ -116,106 +116,89 @@ public SessionCreationResult(Session session, CreationModeEnum mode, boolean alr
...
-            final Session previous = pool.putIfAbsent(clientId, newSession);
-            final boolean success = previous == null;
-
-            if (success) {
-                LOG.trace("case 1, not existing session with CId {}", clientId);
-            } else {
-                postConnectAction = reopenExistingSession(msg, clientId, previous, newSession, username);
+            final Session previous = pool.put(clientId, newSession);
+            if (previous != null) {
+                LOG.error("Another Thread added a Session for our clientId {}, this is a bug!", clientId);
```
andsel:
```suggestion
                // if this happens mean that another Session Event Loop thread processed a CONNECT message 
                // with the same clientId. This is a bug because all messages for the same clientId is serialized on the
                // event loop thread.
                LOG.error("Another thread added a Session for our clientId {}, this is a bug!", clientId);
```

broker/src/main/java/io/moquette/broker/SessionRegistry.java
```diff
@@ -116,106 +116,89 @@ public SessionCreationResult(Session session, CreationModeEnum mode, boolean alr
...
 
     SessionCreationResult createOrReopenSession(MqttConnectMessage msg, String clientId, String username) {
         SessionCreationResult postConnectAction;
-        final Session newSession = createNewSession(msg, clientId);
-        final Session oldSession = pool.get(clientId);
+        //final Session newSession = createNewSession(msg, clientId);
+        final Session oldSession = retrieve(clientId);
         if (oldSession == null) {
-            // case 1
+            // case 1 no old session.
```
andsel:
```suggestion
            // case 1, no existing session with given clientId.
```

broker/src/main/java/io/moquette/broker/SessionRegistry.java
```diff
@@ -116,106 +116,89 @@ public SessionCreationResult(Session session, CreationModeEnum mode, boolean alr
...
-                //the other client was disconnecting and removed it's own session
-                pool.put(clientId, newSession);
+
+        if (newIsClean) {
+            boolean result = oldSession.assignState(SessionStatus.DISCONNECTED, SessionStatus.DESTROYED);
+            if (!result) {
+                throw new SessionCorruptedException("old session has already changed state");
             }
+
+            // case 2
```
andsel:
```suggestion
            // case 2, reopening existing session but with cleanSession true
```

broker/src/main/java/io/moquette/broker/SessionRegistry.java
```diff
@@ -116,106 +116,89 @@ public SessionCreationResult(Session session, CreationModeEnum mode, boolean alr
...
+            creationResult = new SessionCreationResult(newSession, CreationModeEnum.CREATED_CLEAN_NEW, true);
         } else {
-            LOG.debug("Replace session of client with same id");
-            if (!pool.replace(clientId, oldSession, oldSession)) {
-                throw new SessionCorruptedException("old session was already removed");
+            final boolean connecting = oldSession.assignState(SessionStatus.DISCONNECTED, SessionStatus.CONNECTING);
+            if (!connecting) {
+                throw new SessionCorruptedException("old session moved in connected state by other thread");
             }
+            // case 3
```
andsel:
```suggestion
            // case 3, reopening existing session without cleanSession, so keep the existing subscriptions
```

andsel (changes requested):
@hylkevds we are almost there, just a couple of nitpick and the PR then is ready.

hylkevds:
> @hylkevds we are almost there, just a couple of nitpick and the PR then is ready.

Excellent. They should all be fixed now.

andsel (approved):
LGTM

PR #654, opened 2022-02-12
Move the execution of receivePublish for qos0 and qos1 into the Session onEventLoop

andsel:
This PR wraps the execution of methods to handle the receive of PUB at qos0 and qos1 into the SessionEventLoop. While this processing didn't present any problem in being execute by the Netty's IO thread, to be adherent to the chosen architecture, is best to be in a session event loop thread.


andsel:
cc @hylkevds

hylkevds:
Makes sense :+1:

PR #657, opened 2022-02-18
Updates the command enqueuing to Session Event Loop to return the status of the operation

andsel:
## Release notes
[rn:skip]

## What does this PR do?

Fix a design problem in the enqueuing of commands to Session Event loops.
In the first version the of the process publish methods, the `receivedPublish*` operations returned a `CompletableFuture` which wrapped the logic of the message processing. In case of validation error, was returned an already completed future.
For example the `receivedPublishQos1` returned  a future that was obtained as composition of the action to publish to subscribers, concatenated with the logic to send the PUBACK to the publisher. So it happened that the PUBACK was sent when [publish was forwarded to the matching subscribers](https://github.com/moquette-io/moquette/blob/15d5494975cf8f173a829733d60a0add93d51c33/broker/src/main/java/io/moquette/broker/PostOffice.java#L349).
The problem is that the operation to enqueue could be successful or fail, and in case of success there is a `CompleteableFuture` that rappresents the part of the forwarding to subscribers.
The PUBACK message should be sent when the message is taken into account from the broker and not when the publish has completely forwarded to all subscribers.

This PR changes the `receivedPublishQos0` `receivedPublishQos1` `receivedPublishQos2` to return a `RouteResult` object that contains the status of the enqueuing operation plus the future of the command task.

It touches a lot of parts of the code due to signature changes, but the core changes are all in `PostOffice`.

The final user is not impated by this, the developer has a more straightforward and clean process logic

## Checklist

- ~~[ ] I have made corresponding changes to the documentation~~
- ~~[ ] I have made corresponding change to the default configuration files (and/or docker env variables)~~

## Related issues

- Related #631

### DIFF
--- a/broker/src/main/java/io/moquette/broker/MQTTConnection.java
+++ b/broker/src/main/java/io/moquette/broker/MQTTConnection.java
@@ -429,14 +429,14 @@ final class MQTTConnection {
                         return null;
                     postOffice.receivedPublishQos0(topic, username, clientId, msg);
                     return null;
-                });
+                }).ifFailed(msg::release);
             case AT_LEAST_ONCE:
                 return postOffice.routeCommand(clientId, "PUB QoS1", () -> {
                     if (!isBoundToSession())
                         return null;
                     postOffice.receivedPublishQos1(this, topic, username, messageID, msg);
                     return null;
-                });
+                }).ifFailed(msg::release);
             case EXACTLY_ONCE: {
                 final PostOffice.RouteResult firstStepResult = postOffice.routeCommand(clientId, "PUB QoS2", () -> {
                     if (!isBoundToSession())
@@ -444,6 +444,11 @@ final class MQTTConnection {
                     bindedSession.receivedPublishQos2(messageID, msg);
                     return null;
                 });
+                if (!firstStepResult.isSuccess()) {
+                    msg.release();
+                    LOG.trace("Failed to enqueue PUB QoS2 to session loop for  {}", clientId);
+                    return firstStepResult;
+                }
                 firstStepResult.completableFuture().thenRun(() ->
                     postOffice.receivedPublishQos2(this, msg, username).completableFuture()
                 );
--- a/broker/src/main/java/io/moquette/broker/MQTTConnection.java
+++ b/broker/src/main/java/io/moquette/broker/MQTTConnection.java
@@ -25,14 +25,12 @@ import io.netty.channel.ChannelFutureListener;
 import io.netty.channel.ChannelPipeline;
 import io.netty.handler.codec.mqtt.*;
 import io.netty.handler.timeout.IdleStateHandler;
-import io.netty.util.ReferenceCountUtil;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.net.InetSocketAddress;
 import java.util.*;
 import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 
@@ -110,7 +108,7 @@ final class MQTTConnection {
 
     private void processPubComp(MqttMessage msg) {
         final int messageID = ((MqttMessageIdVariableHeader) msg.variableHeader()).messageId();
-        this.postOffice.routeCommand(bindedSession.getClientID(), () -> {
+        this.postOffice.routeCommand(bindedSession.getClientID(), "PUBCOMP", () -> {
             bindedSession.processPubComp(messageID);
             return bindedSession.getClientID();
         });
@@ -118,7 +116,7 @@ final class MQTTConnection {
 
     private void processPubRec(MqttMessage msg) {
         final int messageID = ((MqttMessageIdVariableHeader) msg.variableHeader()).messageId();
-        this.postOffice.routeCommand(bindedSession.getClientID(), () -> {
+        this.postOffice.routeCommand(bindedSession.getClientID(), "PUBREC", () -> {
             bindedSession.processPubRec(messageID);
             return null;
         });
@@ -132,13 +130,13 @@ final class MQTTConnection {
     private void processPubAck(MqttMessage msg) {
         final int messageID = ((MqttMessageIdVariableHeader) msg.variableHeader()).messageId();
         final String clientId = getClientId();
-        this.postOffice.routeCommand(clientId, () -> {
+        this.postOffice.routeCommand(clientId, "PUB ACK", () -> {
             bindedSession.pubAckReceived(messageID);
             return null;
         });
     }
 
-    Future<PostOffice.RouteResult> processConnect(MqttConnectMessage msg) {
+    PostOffice.RouteResult processConnect(MqttConnectMessage msg) {
         MqttConnectPayload payload = msg.payload();
         String clientId = payload.clientIdentifier();
         final String username = payload.userName();
@@ -147,20 +145,20 @@ final class MQTTConnection {
         if (isNotProtocolVersion(msg, MqttVersion.MQTT_3_1) && isNotProtocolVersion(msg, MqttVersion.MQTT_3_1_1)) {
             LOG.warn("MQTT protocol version is not valid. CId: {}", clientId);
             abortConnection(CONNECTION_REFUSED_UNACCEPTABLE_PROTOCOL_VERSION);
-            return CompletableFuture.completedFuture(null);
+            return PostOffice.RouteResult.failed(clientId);
         }
         final boolean cleanSession = msg.variableHeader().isCleanSession();
         if (clientId == null || clientId.length() == 0) {
             if (!brokerConfig.isAllowZeroByteClientId()) {
                 LOG.info("Broker doesn't permit MQTT empty client ID. Username: {}", username);
                 abortConnection(CONNECTION_REFUSED_IDENTIFIER_REJECTED);
-                return CompletableFuture.completedFuture(null);
+                return PostOffice.RouteResult.failed(clientId);
             }
 
             if (!cleanSession) {
                 LOG.info("MQTT client ID cannot be empty for persistent session. Username: {}", username);
                 abortConnection(CONNECTION_REFUSED_IDENTIFIER_REJECTED);
-                return CompletableFuture.completedFuture(null);
+                return PostOffice.RouteResult.failed(clientId);
             }
 
             // Generating client id.
@@ -171,11 +169,11 @@ final class MQTTConnection {
         if (!login(msg, clientId)) {
             abortConnection(CONNECTION_REFUSED_BAD_USER_NAME_OR_PASSWORD);
             channel.close().addListener(CLOSE_ON_FAILURE);
-            return CompletableFuture.completedFuture(null);
+            return PostOffice.RouteResult.failed(clientId);
         }
 
         final String sessionId = clientId;
-        return this.postOffice.routeCommand(clientId, () -> {
+        return postOffice.routeCommand(clientId, "CONN", () -> {
             executeConnect(msg, sessionId);
             return null;
         });
@@ -305,7 +303,7 @@ final class MQTTConnection {
         }
         // this must not be done on the netty thread
         LOG.debug("Notifying connection lost event");
-        postOffice.routeCommand(clientID, () -> {
+        postOffice.routeCommand(clientID, "CONN LOST", () -> {
             if (isBoundToSession() || isSessionUnbound()) {
                 LOG.debug("Cleaning {}", clientID);
                 processConnectionLost(clientID);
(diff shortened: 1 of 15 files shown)


hylkevds:
:+1:  That makes things a lot easier to follow!

andsel:
Hi @hylkevds may I ask you for a review of this PR?
I know it's big, but the core part of the changes is in `PostOffice` class. Let me know :-)

hylkevds:
> Hi @hylkevds may I ask you for a review of this PR?
> I know it's big, but the core part of the changes is in `PostOffice` class. Let me know :-)

Of course, no problem.

hylkevds:
(Review in progress...)
Found one leak, when a publish fails because a Session command queue is full (at least for QoS 2). I think it is either because
- `MQTTConnection.processPublish` does a retain, but when the queue is full the release in `PostOffice.receivedPublishQos2` never runs.
- `PostOffice.publish2Subscribers:499` does a retain for each relevant queue, but the command lambda never executes in `PostOffice.publish2Subscribers:503`

Or maybe it's both reasons.

Maybe we can change `Server.internalPublish` for embedded applications to return the `RoutingResults` too, that way the embedded application can check if the publish was OK, and re-try if it wants to.

broker/src/main/java/io/moquette/broker/MQTTConnection.java
```diff
@@ -415,27 +413,28 @@ void sendUnsubAckMessage(List<String> topics, String clientID, int messageID) {
         msg.retain();
```
hylkevds, line 424: This retain is released in one of the `postOffice.receivedPublisQosX` methods. But only of the routeCommand succeeds.
We should check the RouteResult before returning it, and call `msg.release()` if the RouteResult is failed.

broker/src/main/java/io/moquette/broker/PostOffice.java
```diff
@@ -322,45 +361,47 @@ public void unsubscribe(List<String> topics, MQTTConnection mqttConnection, int
...
-    CompletableFuture<RoutingResults> receivedPublishQos1(MQTTConnection connection, Topic topic, String username, int messageID,
+    RoutingResults receivedPublishQos1(MQTTConnection connection, Topic topic, String username, int messageID,
                                                 MqttPublishMessage msg) {
         // verify if topic can be written
         topic.getTokens();
         if (!topic.isValid()) {
             LOG.warn("Invalid topic format, force close the connection");
             connection.dropConnection();
-            return CompletableFuture.completedFuture(null);
+            return RoutingResults.preroutingError();
```
hylkevds, line 389: When this happens, `ReferenceCountUtil.release(msg);` is not called.

broker/src/main/java/io/moquette/broker/PostOffice.java
```diff
@@ -322,45 +361,47 @@ public void unsubscribe(List<String> topics, MQTTConnection mqttConnection, int
...
         }
         final String clientId = connection.getClientId();
         if (!authorizator.canWrite(topic, username, clientId)) {
             LOG.error("MQTT client: {} is not authorized to publish on topic: {}", clientId, topic);
-            return CompletableFuture.completedFuture(null);
+            return RoutingResults.preroutingError();
```
hylkevds, line 395: When this happens, `ReferenceCountUtil.release(msg);` is not called.

broker/src/main/java/io/moquette/broker/PostOffice.java
```diff
@@ -322,45 +361,47 @@ public void unsubscribe(List<String> topics, MQTTConnection mqttConnection, int
...
+            // QoS1 message was enqueued successfully to every event loop
+            connection.sendPubAck(messageID);
+            manageRetain(topic, msg);
+            interceptor.notifyTopicPublished(msg, clientId, username);
         } else {
-            publishFuture = publish2Subscribers(payload, topic, AT_LEAST_ONCE);
+            // some session event loop enqueue raised a problem
+            failedPublishes.insertAll(messageID, clientId, routes.failedRoutings);
         }
+        ReferenceCountUtil.release(msg);
```
hylkevds, line 418: This is the release that should also be mirrored for the two return statements above.

broker/src/main/java/io/moquette/broker/PostOffice.java
```diff
@@ -309,9 +348,9 @@ public void unsubscribe(List<String> topics, MQTTConnection mqttConnection, int
             LOG.error("client is not authorized to publish on topic: {}", topic);
             return CompletableFuture.completedFuture(null);
```
hylkevds, line 361: When this happens, `ReferenceCountUtil.release(msg);` is not called.

broker/src/main/java/io/moquette/broker/PostOffice.java
```diff
@@ -309,9 +348,9 @@ public void unsubscribe(List<String> topics, MQTTConnection mqttConnection, int
...
         }
-        final CompletableFuture<RoutingResults> publishFuture = publish2Subscribers(msg.payload(), topic, AT_MOST_ONCE);
+        final RoutingResults publishResult = publish2Subscribers(msg.payload(), topic, AT_MOST_ONCE);
```
hylkevds, line 363: When the routing fails, `ReferenceCountUtil.release(msg);` is not called.

broker/src/main/java/io/moquette/broker/PostOffice.java
```diff
@@ -439,26 +496,29 @@ private int subscriberEventLoop(String clientId) {
...
+            .filter(RouteResult::isSuccess)
+            .map(RouteResult::completableFuture).toArray(CompletableFuture[]::new);
+        final CompletableFuture<Void> publishes = CompletableFuture.allOf(publishFutures);
+
+        final List<String> failedRoutings = new ArrayList<>();
+        final List<String> successedRoutings = new ArrayList<>();
+        for (RouteResult rr : publishResults) {
+            Collection<String> subscibersIds = collector.subscriberIdsByEventLoop(rr.clientId);
+            if (rr.status == RouteResult.Status.FAIL) {
+                failedRoutings.addAll(subscibersIds);
```
hylkevds, line 535: For failed batches, `payload.release();` should be called here.

broker/src/main/java/io/moquette/broker/PostOffice.java
```diff
@@ -439,26 +496,29 @@ private int subscriberEventLoop(String clientId) {
                 collector.add(sub);
             }
         }
+        payload.retain(collector.countBatches());
 
-        List<CompletableFuture<RouteResult>> publishFutures = collector.routeBatchedPublishes((batch) -> {
+        List<RouteResult> publishResults = collector.routeBatchedPublishes((batch) -> {
             publishToSession(payload, topic, batch, publishingQos);
+            payload.release();
```
hylkevds, line 522: This `payload.release();` only happens when the routing succeeds.

broker/src/main/java/io/moquette/broker/PostOffice.java
```diff
@@ -489,41 +549,41 @@ private void publishToSession(ByteBuf payload, Topic topic, Subscription sub, Mq
...
         LOG.trace("Processing PUB QoS2 message on connection: {}", connection);
         final Topic topic = new Topic(msg.variableHeader().topicName());
         final ByteBuf payload = msg.payload();
 
         final String clientId = connection.getClientId();
         if (!authorizator.canWrite(topic, username, clientId)) {
             LOG.error("MQTT client is not authorized to publish on topic: {}", topic);
-            return CompletableFuture.completedFuture(null);
+            // WARN this is a special case failed is empty, but this result is to be considered as error.
+            return RoutingResults.preroutingError();
```
hylkevds, line 582: Before returning, `ReferenceCountUtil.release(msg);` should be called.

broker/src/main/java/io/moquette/broker/PostOffice.java
```diff
@@ -583,33 +644,38 @@ void dispatchConnectionLost(String clientId,String userName) {
...
         if (Thread.currentThread() == sessionExecutors[targetQueueId]) {
             SessionEventLoop.executeTask(task);
-            return cmd.completableFuture().thenApply(RouteResult::success);
+            return RouteResult.success(clientId, cmd.completableFuture());
         }
         if (this.sessionQueues[targetQueueId].offer(task)) {
-            return cmd.completableFuture().thenApply(RouteResult::success);
+            return RouteResult.success(clientId, cmd.completableFuture());
         } else {
             LOG.warn("Session command queue {} is full", targetQueueId);
```
hylkevds: Maybe add `actionDescription` to the warning output too? Would help when debugging.

broker/src/main/java/io/moquette/broker/Server.java
```diff
@@ -303,16 +302,16 @@ private void initInterceptors(IConfig props, List<? extends InterceptHandler> em
...
         final int messageID = msg.variableHeader().packetId();
         if (!initialized) {
             LOG.error("Moquette is not started, internal message cannot be published. CId: {}, messageId: {}", clientId,
                       messageID);
             throw new IllegalStateException("Can't publish on a integration is not yet started");
         }
         LOG.trace("Internal publishing message CId: {}, messageId: {}", clientId, messageID);
-        return dispatcher.internalPublish(msg)
-            .thenRun(() -> msg.payload().release());
+        dispatcher.internalPublish(msg);
```
hylkevds: If we return the result of the `dispatcher.internalPublish(msg);`, then the embedding application can check if the publish succeeded, and decide if it wants to do a resend.

hylkevds (commented):
Looks great, makes the flow a lot easier to follow!
Did find some leaks though. They're easier to spot now that the code is clearer :)

broker/src/main/java/io/moquette/broker/MQTTConnection.java
andsel: Fixed with 01ac4e37d857f53be28e8e9646779f7b120b3326

broker/src/main/java/io/moquette/broker/PostOffice.java
andsel: Fixed with f8bfb9c68b4164a686ef2eee2b17127a35aad378

broker/src/main/java/io/moquette/broker/PostOffice.java
andsel: Fixed with f8bfb9c68b4164a686ef2eee2b17127a35aad378

broker/src/main/java/io/moquette/broker/PostOffice.java
andsel: Fixed with f8bfb9c68b4164a686ef2eee2b17127a35aad378

broker/src/main/java/io/moquette/broker/PostOffice.java
andsel: Fixed by 6ca447490a1d96f7311d81fd212939a1592eb080

broker/src/main/java/io/moquette/broker/PostOffice.java
andsel: Fixed  by de34525dcd728c9d5b57d498ee39c09f862b65bc

broker/src/main/java/io/moquette/broker/PostOffice.java
andsel: Fixed by d2d5b498258f49383abeb621d226b13472613ed4

broker/src/main/java/io/moquette/broker/Server.java
andsel: Fixed by 92f341f94779a7a3ffb35e3359600fb9e08f336c

broker/src/main/java/io/moquette/broker/PostOffice.java
andsel: `publishResult` it's not called if and only if _all_ the sub-routings failed. If only one of these successed then the composed Future is invoked and frees the buffer

andsel: Fixed by 95a69a0a4b0f10383eb38f788aee44ec9ec89351

broker/src/main/java/io/moquette/broker/PostOffice.java
andsel: Fixed by 6ca447490a1d96f7311d81fd212939a1592eb080

andsel:
Hi @hylkevds thanks a ton to have spotted all those missed `release`.
Have fixed and aligned to `main`, now thi s PR is ready for a second run of review :-)

andsel:
@hylkevds wait before review, because the test `testClientDoesntRemainSubscribedAfterASubscriptionAndServerRestart` started to hangup

broker/src/main/java/io/moquette/broker/PostOffice.java
```diff
@@ -135,11 +135,11 @@ public RoutingResults(List<String> successedRoutings, List<String> failedRouting
         }
 
         public boolean isAllSuccess() {
-            return failedRoutings.isEmpty();
+            return failedRoutings.isEmpty() && !successedRoutings.isEmpty();
```
hylkevds: That doesn't work.
Now, if a Pub message has no subscribers, both `failedRoutings` and `successedRoutings` are empty, resulting in `isAllSuccess` returning false, and the QoS 1 and 2  messages never being ACKed.
I would say that if both lists are empty then the routing succeeded.

This is also why the tests are hanging.

hylkevds: Reverting that one line fixes the problem, and then this PR looks ready to go.

andsel: Nor
```
failedRoutings.isEmpty() && successedRoutings.isEmpty()
```
 is correct because, when there is no failed routing and only successed `isAllSuccess` has to value to 
`true`.

hylkevds (approved):
Looks good!

PR #658, opened 2022-02-19
Fix 653 retained messages publishing

### LINKED ISSUE #653: QoS 1/2 retained message publishing doesn't work correctly
### Expected behavior
Sending QoS 1/2 retained messages to a subscriber who is initially not connected should work without error.

### Actual behavior
QoS 1 and 2 retained messages do not setup the necessary state management in Moquette which leads to failures, most specifically for QoS 2. When sending retained high-QoS messages, the inflightWindow map isn't used to keep the publish state as it would be when there is an existing session. Instead if uses this code path: https://github.com/moquette-io/moquette/blob/9ed6ca9d373e5315a58053ace88107b005c47de7/broker/src/main/java/io/moquette/broker/MQTTConnection.java#L456 which will publish the message to the client, but it doesn't setup the session structures correctly to accept the PUBACK, PUBREC, etc. We therefore get the error `Received a PUBACK with not matching packetId` and `Received a PUBREC with not matching packetId`

### Steps to reproduce

https://github.com/eclipse/paho.mqtt.testing/blob/8f5ff2d47d6099f1624958299e782034632772d2/interoperability/client_test.py#L160-L164

### Moquette MQTT version
Tip of main (8e2a0b830b9c4e5110caa859253a1d458ab3b1d0)

hylkevds:
This PR adds a test for, and fixes #653.

Retained messages used a difference code-path that skipped the Session, and thus did not properly set up in-flight status. 
To fix this, this PR:
- Adds a test for retained messages at different QoS levels
- Makes the retained messages use the same code path as non-retained messages
- Removes the retained-specific code.
- Adds a `retained` flag to this code-path
- Adds a `retained` flag to `SessionRegistry.PublishedMessage` so the flag is stored for queued messages

### DIFF
--- a/broker/src/main/java/io/moquette/broker/MQTTConnection.java
+++ b/broker/src/main/java/io/moquette/broker/MQTTConnection.java
@@ -535,17 +535,30 @@ final class MQTTConnection {
 
     // TODO move this method in Session
     void sendPublishQos0(Topic topic, MqttQoS qos, ByteBuf payload, boolean retained) {
-        MqttPublishMessage publishMsg = createPublishMessage(topic.toString(), qos, payload, retained);
+        MqttPublishMessage publishMsg = createPublishMessage(topic.toString(), qos, payload, 0, retained);
         sendPublish(publishMsg);
     }
 
-    static MqttPublishMessage createPublishMessage(String topic, MqttQoS qos, ByteBuf message,
-                                                   boolean retained) {
-        return createPublishMessage(topic, qos, message, 0, retained);
+    static MqttPublishMessage createRetainedPublishMessage(String topic, MqttQoS qos, ByteBuf message) {
+        return createPublishMessage(topic, qos, message, 0, true);
     }
 
-    static MqttPublishMessage createPublishMessage(String topic, MqttQoS qos, ByteBuf message,
-                                                   int messageId, boolean retained) {
+    static MqttPublishMessage createNonRetainedPublishMessage(String topic, MqttQoS qos, ByteBuf message) {
+        return createPublishMessage(topic, qos, message, 0, false);
+    }
+
+    static MqttPublishMessage createRetainedPublishMessage(String topic, MqttQoS qos, ByteBuf message,
+                                                           int messageId) {
+        return createPublishMessage(topic, qos, message, messageId, true);
+    }
+
+    static MqttPublishMessage createNonRetainedPublishMessage(String topic, MqttQoS qos, ByteBuf message,
+                                                              int messageId) {
+        return createPublishMessage(topic, qos, message, messageId, false);
+    }
+
+    private static MqttPublishMessage createPublishMessage(String topic, MqttQoS qos, ByteBuf message,
+                                                           int messageId, boolean retained) {
         MqttFixedHeader fixedHeader = new MqttFixedHeader(MqttMessageType.PUBLISH, false, qos, retained, 0);
         MqttPublishVariableHeader varHeader = new MqttPublishVariableHeader(topic, messageId);
         return new MqttPublishMessage(fixedHeader, varHeader, message);
--- a/broker/src/main/java/io/moquette/broker/MQTTConnection.java
+++ b/broker/src/main/java/io/moquette/broker/MQTTConnection.java
@@ -527,41 +527,26 @@ final class MQTTConnection {
         return NettyUtils.userName(channel);
     }
 
-    public void sendPublishRetainedQos0(Topic topic, MqttQoS qos, ByteBuf payload) {
-        MqttPublishMessage publishMsg = retainedPublish(topic.toString(), qos, payload);
-        sendPublish(publishMsg);
-    }
-
-    public void sendPublishRetainedWithPacketId(Topic topic, MqttQoS qos, ByteBuf payload) {
+    public void sendPublishWithPacketId(Topic topic, MqttQoS qos, ByteBuf payload, boolean retained) {
         final int packetId = nextPacketId();
-        MqttPublishMessage publishMsg = retainedPublishWithMessageId(topic.toString(), qos, payload, packetId);
+        MqttPublishMessage publishMsg = createPublishMessage(topic.toString(), qos, payload, packetId, retained);
         sendPublish(publishMsg);
-    }
-
-    private static MqttPublishMessage retainedPublish(String topic, MqttQoS qos, ByteBuf message) {
-        return retainedPublishWithMessageId(topic, qos, message, 0);
-    }
-
-    private static MqttPublishMessage retainedPublishWithMessageId(String topic, MqttQoS qos, ByteBuf message,
-                                                                   int messageId) {
-        MqttFixedHeader fixedHeader = new MqttFixedHeader(MqttMessageType.PUBLISH, false, qos, true, 0);
-        MqttPublishVariableHeader varHeader = new MqttPublishVariableHeader(topic, messageId);
-        return new MqttPublishMessage(fixedHeader, varHeader, message);
     }
 
     // TODO move this method in Session
-    void sendPublishNotRetainedQos0(Topic topic, MqttQoS qos, ByteBuf payload) {
-        MqttPublishMessage publishMsg = notRetainedPublish(topic.toString(), qos, payload);
+    void sendPublishQos0(Topic topic, MqttQoS qos, ByteBuf payload, boolean retained) {
+        MqttPublishMessage publishMsg = createPublishMessage(topic.toString(), qos, payload, retained);
         sendPublish(publishMsg);
     }
 
-    static MqttPublishMessage notRetainedPublish(String topic, MqttQoS qos, ByteBuf message) {
-        return notRetainedPublishWithMessageId(topic, qos, message, 0);
+    static MqttPublishMessage createPublishMessage(String topic, MqttQoS qos, ByteBuf message,
+                                                   boolean retained) {
+        return createPublishMessage(topic, qos, message, 0, retained);
     }
 
-    static MqttPublishMessage notRetainedPublishWithMessageId(String topic, MqttQoS qos, ByteBuf message,
-                                                              int messageId) {
-        MqttFixedHeader fixedHeader = new MqttFixedHeader(MqttMessageType.PUBLISH, false, qos, false, 0);
+    static MqttPublishMessage createPublishMessage(String topic, MqttQoS qos, ByteBuf message,
+                                                   int messageId, boolean retained) {
+        MqttFixedHeader fixedHeader = new MqttFixedHeader(MqttMessageType.PUBLISH, false, qos, retained, 0);
         MqttPublishVariableHeader varHeader = new MqttPublishVariableHeader(topic, messageId);
         return new MqttPublishMessage(fixedHeader, varHeader, message);
     }
--- a/broker/src/main/java/io/moquette/broker/MQTTConnection.java
+++ b/broker/src/main/java/io/moquette/broker/MQTTConnection.java
@@ -552,7 +552,7 @@ final class MQTTConnection {
         return createPublishMessage(topic, qos, message, messageId, true);
     }
 
-    static MqttPublishMessage createNonRetainedPublishMessage(String topic, MqttQoS qos, ByteBuf message,
+    static MqttPublishMessage createNotRetainedPublishMessage(String topic, MqttQoS qos, ByteBuf message,
                                                               int messageId) {
         return createPublishMessage(topic, qos, message, messageId, false);
     }
--- a/broker/src/main/java/io/moquette/broker/PostOffice.java
+++ b/broker/src/main/java/io/moquette/broker/PostOffice.java
@@ -475,7 +475,7 @@ class PostOffice {
         if (isSessionPresent) {
             LOG.debug("Sending PUBLISH message to active subscriber CId: {}, topicFilter: {}, qos: {}",
                       sub.getClientId(), sub.getTopicFilter(), qos);
-            targetSession.sendNonRetainedPublishOnSessionAtQos(topic, qos, payload);
+            targetSession.sendNotRetainedPublishOnSessionAtQos(topic, qos, payload);
         } else {
             // If we are, the subscriber disconnected after the subscriptions tree selected that session as a
             // destination.
--- a/broker/src/main/java/io/moquette/broker/PostOffice.java
+++ b/broker/src/main/java/io/moquette/broker/PostOffice.java
@@ -249,7 +249,7 @@ class PostOffice {
                 MqttQoS qos = lowerQosToTheSubscriptionDesired(subscription, retainedQos);
 
                 final ByteBuf payloadBuf = Unpooled.wrappedBuffer(retainedMsg.getPayload());
-                targetSession.sendPublishOnSessionAtQos(retainedMsg.getTopic(), qos, payloadBuf, true);
+                targetSession.sendRetainedPublishOnSessionAtQos(retainedMsg.getTopic(), qos, payloadBuf);
                 // We made the buffer, we must release it.
                 payloadBuf.release();
             }
@@ -475,7 +475,7 @@ class PostOffice {
         if (isSessionPresent) {
             LOG.debug("Sending PUBLISH message to active subscriber CId: {}, topicFilter: {}, qos: {}",
                       sub.getClientId(), sub.getTopicFilter(), qos);
-            targetSession.sendPublishOnSessionAtQos(topic, qos, payload, false);
+            targetSession.sendNonRetainedPublishOnSessionAtQos(topic, qos, payload);
         } else {
             // If we are, the subscriber disconnected after the subscriptions tree selected that session as a
             // destination.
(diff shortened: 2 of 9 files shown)


hylkevds:
@andsel: Ready for review.
The build issue is fixed, on the build-server the initial retained publish was processed slow enough for the subscription to finish first, causing the initial publish to be sent to the subscriber, and not the retained message. The test is more robust now.

broker/src/main/java/io/moquette/broker/PostOffice.java
```diff
@@ -249,7 +249,7 @@ private void publishRetainedMessagesForSubscriptions(String clientID, List<Subsc
                 MqttQoS qos = lowerQosToTheSubscriptionDesired(subscription, retainedQos);
 
                 final ByteBuf payloadBuf = Unpooled.wrappedBuffer(retainedMsg.getPayload());
-                targetSession.sendRetainedPublishOnSessionAtQos(retainedMsg.getTopic(), qos, payloadBuf);
+                targetSession.sendPublishOnSessionAtQos(retainedMsg.getTopic(), qos, payloadBuf, true);
```
andsel: Method like this was introduced to make explicit what it does, and avoid boolean flag that needed to investigate every time what that boolean flag was. So for example 
```java
targetSession.sendPublishOnSessionAtQos(retainedMsg.getTopic(), qos, payloadBuf, true);
``` 
becomes more explicit, specifying the intention with the method name itself
```java
targetSession.sendRetainedPublishOnSessionAtQos(retainedMsg.getTopic(), qos, payloadBuf);
``` 
 I wouldn't like remove, because the client code becomes more readable, and it's more clear its intention.
Every time we use a method with explicitly passed value of `true` or `false` to programatically pass the retained behavior, we could introduce a method with the `retained` suffix and the other with `notRetained`.

broker/src/test/java/io/moquette/integration/ServerIntegrationRetainTest.java
```diff
@@ -0,0 +1,190 @@
+/*
+ * Copyright (c) 2012-2018 The original author or authors
```
andsel:
```suggestion
 * Copyright (c) 2012-2022 The original author or authors
```

broker/src/test/java/io/moquette/integration/ServerIntegrationRetainTest.java
```diff
@@ -0,0 +1,190 @@
...
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.Assertions;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class ServerIntegrationRetainTest {
+
+    private static final Logger LOG = LoggerFactory.getLogger(ServerIntegrationRetainTest.class);
+
+    private static IConfig m_config;
+    private static Server m_server;
```
andsel: I would avoid `m_` prefixed fields in new code base.

broker/src/test/java/io/moquette/integration/ServerIntegrationRetainTest.java
```diff
@@ -0,0 +1,190 @@
...
+            assertEquals(Math.min(qosPub, qosSub), message.getQos());
+        } else {
+            Assertions.assertFalse(shouldReceive, "QoS " + qosPub + " Messages should be retained.");
+        }
+    }
+
+    @Test
+    public void checkSubscriberQoS0ReceiveQoS0Retained() throws Exception {
+        LOG.info("*** checkSubscriberQoS0ReceiveQoS0Retained ***");
+        checkRetained(0, 0, false);
```
andsel: Using a parameter to make the check code customizable complicates the test code because needs conditionals.
Only for qos0 the messages could be not received, because it's a best effort. So we could create two methods that share some common logic:

```java
public void publishRetained(int qosPub, int qosSub) throws Exception {
        final String topic = "/topic" + qosPub + qosSub;
        final String messageString = "Hello world MQTT " + qosPub + " " + qosSub;
        clientPublisher.subscribe(topic);
        clientPublisher.publish(topic, messageString.getBytes(UTF_8), qosPub, true);
        // Wait for the publish to finish
        Awaitility.await().until(() -> callbackPublisher.isMessageReceived());
        callbackPublisher.getMessageImmediate();

        clientSubscriber.subscribe(topic, qosSub);
}

public void verifyShouldReceive(int qosPub, int qosSub) {
    final boolean messageReceived = callbackSubscriber.isMessageReceived();
    if (messageReceived) {
      verifyReceive(qosPub, qosSub);
    }
}

public void verifyMustReceive(int qosPub, int qosSub) {
    assertTrue(callbackSubscriber.isMessageReceived(), "Expected message to be received");
    verifyReceive(qosPub, qosSub);
}


public void verifyReceive(int qosPub, int qosSub) {
    MqttMessage message = callbackSubscriber.retrieveMessage();
    assertEquals(messageString, message.toString());
    assertEquals(Math.min(qosPub, qosSub), message.getQos());
}


@Test
public void checkSubscriberQoS0ReceiveQoS0Retained() throws Exception {
    publishRetained(0, 0);
    verifyShouldReceive(0, 0);
}    
...
@Test
public void checkSubscriberQoS0ReceiveQoS1Retained() throws Exception {
    publishRetained(1, 0);
    verifyReceive(1, 0);

}

```

andsel (changes requested):
Thank's @hylkevds for your work,
I've left some comments for changes, to try to minimize the code change, but more importantly to clean the intentions in some code paths.

andsel:
@hylkevds another quick note, these tests seems more suitable to be parameterized test: https://junit.org/junit5/docs/current/user-guide/#writing-tests-parameterized-tests

broker/src/main/java/io/moquette/broker/PostOffice.java
hylkevds: I could do:

```
    public void sendRetainedPublishOnSessionAtQos(Topic topic, MqttQoS qos, ByteBuf payload) {
        sendPublishOnSessionAtQos(topic, qos, payload, true);
    }

    public void sendNonRetainedPublishOnSessionAtQos(Topic topic, MqttQoS qos, ByteBuf payload) {
        sendPublishOnSessionAtQos(topic, qos, payload, true);
    }

    private void sendPublishOnSessionAtQos(Topic topic, MqttQoS qos, ByteBuf payload, boolean retained) { ... }
```

That still minimises code duplication.

broker/src/test/java/io/moquette/integration/ServerIntegrationRetainTest.java
hylkevds: No problem.
What was the reason you used that prefix originally?

hylkevds:
> @hylkevds another quick note, these tests seems more suitable to be parameterized test: https://junit.org/junit5/docs/current/user-guide/#writing-tests-parameterized-tests

Yep, I copied the test from `ServerIntegrationQoSValidationTest`, we should convert that one too at some point.

andsel:
Was stolen from cpp practice, m stay for member, it dates when Moquette was
started and the same practice was used by Apache Felix, so I stool the
idea. But the time proven, it's useless, also the 'this.' with ide like
idea are just more text to read.

Il dom 20 feb 2022, 19:58 Hylke van der Schaaf ***@***.***>
ha scritto:

broker/src/test/java/io/moquette/integration/ServerIntegrationRetainTest.java
hylkevds: Actually, for QoS0 we currently do not retain at all, we only clean any previous retain:
https://github.com/moquette-io/moquette/blob/main/broker/src/main/java/io/moquette/broker/PostOffice.java#L314-L322
Do we want to change that? If so, probably better to do that in a separate Issue/PR...

broker/src/main/java/io/moquette/broker/PostOffice.java
hylkevds: Ok, updated in 10b6d637fe3ad09720fe560153231a3ef7b079e5

broker/src/test/java/io/moquette/integration/ServerIntegrationRetainTest.java
hylkevds: I've parameterised the test. That turns out quite nice!

Since the current broker does not retain QoS0 messages, I've kept the explicit checks for this, and made these methods:
```
private void validateMustReceive(int qosPub, int qosSub)
private void validateMustNotReceive(int qosPub)
```

I've also extended the test to test if the retain is cleaned up when sensing an empty retained message.

broker/src/main/java/io/moquette/broker/MQTTConnection.java
```diff
@@ -527,41 +527,39 @@ String getUsername() {
...
-        MqttPublishMessage publishMsg = notRetainedPublish(topic.toString(), qos, payload);
-        sendPublish(publishMsg);
+    static MqttPublishMessage createRetainedPublishMessage(String topic, MqttQoS qos, ByteBuf message,
+                                                           int messageId) {
+        return createPublishMessage(topic, qos, message, messageId, true);
     }
 
-    static MqttPublishMessage notRetainedPublish(String topic, MqttQoS qos, ByteBuf message) {
-        return notRetainedPublishWithMessageId(topic, qos, message, 0);
+    static MqttPublishMessage createNonRetainedPublishMessage(String topic, MqttQoS qos, ByteBuf message,
```
andsel:
```suggestion
    static MqttPublishMessage createNotRetainedPublishMessage(String topic, MqttQoS qos, ByteBuf message,
```

broker/src/main/java/io/moquette/broker/Session.java
```diff
@@ -260,15 +268,15 @@ private void sendPublishQos1(Topic topic, MqttQoS qos, ByteBuf payload) {
...
+            EnqueuedMessage old = inflightWindow.put(packetId, new PublishedMessage(topic, qos, payload, retained));
             // If there already was something, release it.
             if (old != null) {
                 old.release();
                 inflightSlots.incrementAndGet();
             }
             inflightTimeouts.add(new InFlightPacket(packetId, FLIGHT_BEFORE_RESEND_MS));
 
-            MqttPublishMessage publishMsg = MQTTConnection.notRetainedPublishWithMessageId(topic.toString(), qos,
+            MqttPublishMessage publishMsg = MQTTConnection.createNonRetainedPublishMessage(topic.toString(), qos,
```
andsel:
```suggestion
            MqttPublishMessage publishMsg = MQTTConnection.createNotRetainedPublishMessage(topic.toString(), qos,
```

broker/src/main/java/io/moquette/broker/Session.java
```diff
@@ -278,37 +286,37 @@ private void sendPublishQos1(Topic topic, MqttQoS qos, ByteBuf payload) {
...
+            EnqueuedMessage old = inflightWindow.put(packetId, new SessionRegistry.PublishedMessage(topic, qos, payload, retained));
             // If there already was something, release it.
             if (old != null) {
                 old.release();
                 inflightSlots.incrementAndGet();
             }
             inflightTimeouts.add(new InFlightPacket(packetId, FLIGHT_BEFORE_RESEND_MS));
 
-            MqttPublishMessage publishMsg = MQTTConnection.notRetainedPublishWithMessageId(topic.toString(), qos,
+            MqttPublishMessage publishMsg = MQTTConnection.createNonRetainedPublishMessage(topic.toString(), qos,
```
andsel:
```suggestion
            MqttPublishMessage publishMsg = MQTTConnection.createNotRetainedPublishMessage(topic.toString(), qos,
```

broker/src/main/java/io/moquette/broker/Session.java
```diff
@@ -413,7 +421,7 @@ private void drainQueueToConnection() {
             }
             inflightTimeouts.add(new InFlightPacket(sendPacketId, FLIGHT_BEFORE_RESEND_MS));
             final SessionRegistry.PublishedMessage msgPub = (SessionRegistry.PublishedMessage) msg;
-            MqttPublishMessage publishMsg = MQTTConnection.notRetainedPublishWithMessageId(
+            MqttPublishMessage publishMsg = MQTTConnection.createNonRetainedPublishMessage(
```
andsel:
```suggestion
            MqttPublishMessage publishMsg = MQTTConnection.createNotRetainedPublishMessage(
```

broker/src/main/java/io/moquette/broker/Session.java
```diff
@@ -229,25 +229,33 @@ public void processPubComp(int messageID) {
 //                m_interceptor.notifyMessageAcknowledged(interceptAckMsg);
     }
 
-    public void sendPublishOnSessionAtQos(Topic topic, MqttQoS qos, ByteBuf payload) {
+    public void sendRetainedPublishOnSessionAtQos(Topic topic, MqttQoS qos, ByteBuf payload) {
+        sendPublishOnSessionAtQos(topic, qos, payload, true);
+    }
+
+    public void sendNonRetainedPublishOnSessionAtQos(Topic topic, MqttQoS qos, ByteBuf payload) {
```
andsel:
```suggestion
    public void sendNotRetainedPublishOnSessionAtQos(Topic topic, MqttQoS qos, ByteBuf payload) {
```

broker/src/test/java/io/moquette/broker/SessionTest.java
```diff
@@ -60,7 +60,7 @@ public void testPubAckDrainMessagesRemainingInQueue() {
 
     private void sendQoS1To(Session client, Topic destinationTopic, String message) {
         final ByteBuf payload = ByteBufUtil.writeUtf8(UnpooledByteBufAllocator.DEFAULT, message);
-        client.sendPublishOnSessionAtQos(destinationTopic, MqttQoS.AT_LEAST_ONCE, payload);
+        client.sendNonRetainedPublishOnSessionAtQos(destinationTopic, MqttQoS.AT_LEAST_ONCE, payload);
```
andsel:
```suggestion
        client.sendNotRetainedPublishOnSessionAtQos(destinationTopic, MqttQoS.AT_LEAST_ONCE, payload);
```

broker/src/main/java/io/moquette/broker/Session.java
```diff
@@ -229,25 +229,33 @@ public void processPubComp(int messageID) {
...
+        sendPublishOnSessionAtQos(topic, qos, payload, true);
```
andsel: I'm little bit confused about this. The method name says "not retained" but the `retain` flag is passed to `true`.
If this didn't trigger any test error, would it mean that tests doesn't cover enough, for example checking the retain flag in published message the subscriber?
Please update the test to catch this and then switch the flag that prove the fix solved.

andsel (changes requested):
Hi @hylkevds, left just a couple of notes about the naming (`not` vs `non`) plus I think there is an issue with tests that doesn't cover the retained flag to `true` while is should be to `false` .  However we are close to get it merged.

hylkevds:
> Hi @hylkevds, left just a couple of notes about the naming (`not` vs `non`) 

Is that a convention you use? Since nonretained is the proper English variant. the sentence

> send (a) not retained publish

is not correct English. While the sentence

> send (a) nonretained publish

is correct... Hence my automatic use of the "non" variant.
Even the MQTT standard uses non-retained...

Since there are other instances of not- i've committed your suggestions. It's better to do language changes like that, if we decide to make them, in a separate PR.

broker/src/main/java/io/moquette/broker/Session.java
hylkevds: Ohh, good catch!
I'll extend the test to check that normal messages are not sent with a retained flag.

hylkevds: I also added a test to check if QoS0 retained messages cancel an existing retain.

andsel:
@hylkevds about https://github.com/moquette-io/moquette/pull/658#issuecomment-1050650027 my bad, keep the `non`

broker/src/test/java/io/moquette/integration/ServerIntegrationRetainTest.java
```diff
@@ -0,0 +1,232 @@
...
+        }
+    }
+
+    private void validateRetainedFlagNotSet(MqttMessage message) {
+        Assertions.assertFalse(message.isRetained(), "Directly published version of messages should not have the retained flag set");
+    }
+
+    private void validateMustReceive(int qosPub, int qosSub) {
+        final boolean messageReceived = callbackSubscriber.isMessageReceived();
+        Assertions.assertTrue(messageReceived, "Expected a message retained at QoS " + qosPub + ".");
```
andsel: Just a note on `Assertions.assertTrue` Assertions.assertFalse`, in all the other part of the code I used the static import of the assert method, for example `assertEquals`, we should keep the same convention.
Last step then we can marge.

andsel (changes requested):
Left a minor nitpick

broker/src/test/java/io/moquette/integration/ServerIntegrationRetainTest.java
hylkevds: > @hylkevds about [#658 (comment)](https://github.com/moquette-io/moquette/pull/658#issuecomment-1050650027) my bad, keep the `non`

For consistency, shall I also change all other occurrences of notRetained to nonRetained in code that did not change otherwise? Like the `PostOfficeInternalPublishTest` ?

andsel: Yes please, but do in another follow up PR

hylkevds: Ok, I'll use "not" in this PR, and create a new "spelling" PR to change all uses of notRetained to nonRetained

PR #660, opened 2022-02-25
Fix #659: use buffer.toString(Charset) to directly get a String

### LINKED ISSUE #659: wrong way to get a string out of ByteBuf in src/main/java/io/moquette/testembedded/EmbeddedLauncher.java
### Expected behavior
in the file 
https://github.com/moquette-io/moquette/blob/main/embedding_moquette/src/main/java/io/moquette/testembedded/EmbeddedLauncher.java

should be 

	/**
		 * has some issue with the dependency on netty
		 * keep throwing java.lang.UnsupportedOperationException: direct buffer
		 */
		@Override
		public synchronized void onPublish(InterceptPublishMessage msg) {
			// this is the correct way to do it from here
			// https://stackoverflow.com/questions/52658774/netty-io-netty-buffer-bytebuf-array-throws-exception-direct-buffer/52659401
			ByteBuf buf = msg.getPayload();
			final String decodedPayload = buf.toString(StandardCharsets.UTF_8);

			System.out.println("Received on topic: " + msg.getTopicName() + " content: " + decodedPayload);
		}
	}

### Actual behavior

 @Override
        public void onPublish(InterceptPublishMessage msg) {
            final String decodedPayload = new String(msg.getPayload().array(), UTF_8);
            System.out.println("Received on topic: " + msg.getTopicName() + " content: " + decodedPayload);
        }
### Steps to reproduce

when running with a message generator that fires a message every second you get 
java.lang.UnsupportedOperationException: direct buffer
 from time to time and than after 10-15 sec constantly

### Moquette MQTT version
0.15 

11

windows 10

hylkevds:
When the goal is to turn the contents of a buffer into a String it is
more efficient and simpler to directly use the toString(Charset) method
of ByteBuf instead of fetching the byte data and manually creating a
string.

This PR replaces the two spots where buffers are turned into Strings using the "old" way.

### DIFF
--- a/broker/src/main/java/io/moquette/broker/DebugUtils.java
+++ b/broker/src/main/java/io/moquette/broker/DebugUtils.java
@@ -23,17 +23,7 @@ import java.nio.charset.StandardCharsets;
 public final class DebugUtils {
 
     public static String payload2Str(ByteBuf content) {
-        final ByteBuf copy = content.copy();
-        final byte[] bytesContent;
-        if (copy.isDirect()) {
-            final int size = copy.readableBytes();
-            bytesContent = new byte[size];
-            copy.readBytes(bytesContent);
-            copy.release();
-        } else {
-            bytesContent = copy.array();
-        }
-        return new String(bytesContent, StandardCharsets.UTF_8);
+        return content.copy().toString(StandardCharsets.UTF_8);
     }
 
     private DebugUtils() {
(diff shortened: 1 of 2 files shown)


andsel (approved):
LGTM

andsel:
Thanks @hylkevds to have spotted and fixed it!

PR #669, opened 2022-04-01
Made RoutingResults public

hylkevds:
Server.internalPublish returns the RoutingResults, but PostOffice.RoutingResults was not public.
Since PostOffice itself is not public either, RoutingResults had to be moved out of this class to its own file.


andsel (approved):
LGTM

PR #670, opened 2022-04-01
Fixed #667: NPE in InterceptConnectMessage.getPassword

### LINKED ISSUE #667: InterceptConnectMessage.getPassword NPE
### Expected behavior
`InterceptConnectMessage.getPassword` returns null or Optional when no password is sent to the moquette server.

### Actual behavior
`InterceptConnectMessage.getPassword` throws NullPointerException.

### Steps to reproduce

I tried serializing `InterceptConnectMessage` to json, this reflectively calls `getPassword` which I am not passing in any clients yet. The lack of a password in `MqttConnectMessage.payload` causes an NPE when you call `.getBytes(StandardCharsets.UTF_8)`:

```java
    public byte[] getPassword() {
        return this.msg.payload().password().getBytes(StandardCharsets.UTF_8);
    }
```

### Alternatives
This was discovered using rather insensitive logging (I just log all messages as JSON right now for debugging purposes). User principal should not get logged, so I really shouldn't be doing this. A `toString` implementation that conceals sensitive data would be a reasonable secure alternative (new feature) for my use case, but the null condition happens due to client input and arguably should be checked for regardless.

### Moquette MQTT version
0.15

```
openjdk version "11.0.15" 2022-04-19
OpenJDK Runtime Environment (build 11.0.15+3)
OpenJDK 64-Bit Server VM (build 11.0.15+3, mixed mode)
```
```
Linux myhostname 5.16.15-arch1-1 #1 SMP PREEMPT Thu, 17 Mar 2022 00:30:09 +0000 x86_64 GNU/Linux
```

hylkevds:
InterceptConnectMessage.getPassword used msg.payload().password().getBytes but msg.payload() could return null. Additionally, msg.payload() is depricated.
Instead, InterceptConnectMessage.getPassword now directly returns msg.payload().passwordInBytes() and thus never throws an NPE and avoids two re-coding steps.

### DIFF
--- a/broker/src/main/java/io/moquette/interception/messages/InterceptConnectMessage.java
+++ b/broker/src/main/java/io/moquette/interception/messages/InterceptConnectMessage.java
@@ -18,8 +18,6 @@ package io.moquette.interception.messages;
 
 import io.netty.handler.codec.mqtt.MqttConnectMessage;
 
-import java.nio.charset.StandardCharsets;
-
 public class InterceptConnectMessage extends InterceptAbstractMessage {
 
     private final MqttConnectMessage msg;
@@ -82,6 +80,6 @@ public class InterceptConnectMessage extends InterceptAbstractMessage {
     }
 
     public byte[] getWillMessage() {
-        return msg.payload().willMessage().getBytes(StandardCharsets.UTF_8);
+        return msg.payload().willMessageInBytes();
     }
 }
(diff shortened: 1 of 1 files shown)


hylkevds:
I've also fixed the same issue for getWillMessage() that used willMessage() instead of willMessageInBytes()

andsel (approved):
LGTM

andsel:
Thanks @hylkevds :+1:

PR #675, opened 2022-05-21
feat: peer cert as username

jbutler:
This PR adds enables authenticating clients based on their peer certificates without changing the existing authentication interface. This functionality is enabled by the addition of a new configuration option which passes the client certificate PEM in place of `username`. The EMQx broker also has the same configuration option.

There is also a very small change to behavior with this PR. Previously, clients that did not provide a password were immediately rejected if anonymous sessions were disabled. As far as I can tell, the spec allows username without a password. The code will now allow this and pass the username and empty password to the authenticator.

_Why is this needed?_
The existing auth interfaces are insufficient when clients need to be identified based on their client certificate. The problem arises if an MQTT client connects with a duplicate client ID, and then an authorization request for that client ID is made. At that point, there is a period of time when the authorizer cannot know if the request is for the old session or the new session. This is problematic since there is no graceful way to handle this.

If the authorizer allows the operation, then a malicious client could exploit this to publish (or subscribe) to topics it should not normally be authorized on.

If the authorizer assumes that the operation is coming from the old client and denies the operation, then it is possible for operations which should be allowed to be be silently rejected (as Moquette does not disconnect clients after authZ failures).

By passing in the client certificate as username, this allows the authorizer to uniquely identify clients.

_Alternative solutions_
Alternatively, we could pass additional context along with authorization requests to enable a custom authorizer to uniquely identify clients. This extra context could be an opaque object provided by the Authenticator/Authorizer itself. Or it could be as simple as monotonically increasing "connection ID" which is tied to the `MQTTConnection` itself. However, this approach is trickier since it would likely break existing custom authenticators / authorizers (or involves introducing a new interface)

### DIFF
--- a/broker/src/main/java/io/moquette/BrokerConstants.java
+++ b/broker/src/main/java/io/moquette/BrokerConstants.java
@@ -49,6 +49,7 @@ public final class BrokerConstants {
     public static final String KEY_STORE_PASSWORD_PROPERTY_NAME = "key_store_password";
     public static final String KEY_MANAGER_PASSWORD_PROPERTY_NAME = "key_manager_password";
     public static final String ALLOW_ANONYMOUS_PROPERTY_NAME = "allow_anonymous";
+    public static final String PEER_CERTIFICATE_AS_USERNAME = "peer_certificate_as_username";
     public static final String REAUTHORIZE_SUBSCRIPTIONS_ON_CONNECT = "reauthorize_subscriptions_on_connect";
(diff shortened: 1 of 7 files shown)


MikeDombo:
@andsel would you consider this change? I can pick it up and rebase as may be needed.

andsel:
Hi @MikeDombo yes please, you could rebase it, then this change of behavior in Moquette I think it's better to be explicitly enabled by a feature flag.
Once done ask me as reviewer and I'll  take a look.

PR #680, opened 2022-06-09
Add test for Topic.isValid

lacinoire:
Hey 😊
I want to contribute the following test:

Test that `retainedTopic.isValid()` is true.
This tests the method [`Topic.isValid`](https://github.com/moquette-io/moquette/blob/3e6043be04b915879804f2740bd7d500331e2f24/broker/src/main/java/io/moquette/broker/subscriptions/Topic.java#L154).
This test is based on the test [`testRetainedOnTopicReturnsExactTopicMatch`](https://github.com/lacinoire/moquette/blob/3e6043be04b915879804f2740bd7d500331e2f24/broker/src/test/java/io/moquette/broker/MemoryRetainedRepositoryTest.java#L92).

Curious to hear what you think!

(I wrote this test as part of a research study at TU Delft. [Find out more](https://github.com/lacinoire/lacinoire/blob/main/README.md))

### DIFF
--- a/broker/src/test/java/io/moquette/broker/MemoryRetainedRepositoryTest.java
+++ b/broker/src/test/java/io/moquette/broker/MemoryRetainedRepositoryTest.java
@@ -20,6 +20,7 @@ import io.moquette.broker.subscriptions.Topic;
 import io.netty.buffer.Unpooled;
 import io.netty.handler.codec.mqtt.MqttMessageBuilders;
 import io.netty.handler.codec.mqtt.MqttQoS;
+import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
 
 import java.util.List;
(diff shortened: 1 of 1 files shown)


hylkevds:
I'm guessing you're trying to automatically generate tests as part of your research?

If so, then I think this test is a good example of a failure :)
- The test used as a base has nothing to do with the method being tested, and as a result most of the code in the test is irrelevant. The test, as written, could be reduced to two lines.
- It tests a method returning a boolean value, but only tests the situation with a `true` return value.

A test for `isValid()` should have analysed the code used to generate the value being returned by `isValid()`, which is `Token.valid`, which in turn is only set in `Topic.getTokens()`, which bases the value on `Topic.parseTopic(String topic)` throwing an exception or not. So your test generator should have generated a set of Tests, one for each way that `Topic.parseTopic(String topic)` can throw an exception.

lacinoire:
Indeed you spotted us correctly here :)
Thank you for the in-depth feedback on the issues with this test and what a better test for `isValid()` would test! 🙏🏼 

Unfortunately our current tool cannot produce a test like this yet, so I will close this PR 🙂
Thanks for helping us with your input 😊

hylkevds:
You're welcome :)
I think generating test is a fascinating topic that could really improve the reliability of development, so feel free to give it another try in the future.

PR #691, opened 2022-11-23
Persisted unsafe queues implementation

andsel:
## Release notes
Implements a disk persistent queues system that is not thread safe.

## What it does?
Create the implementation of segmented queues. The data is stored in mmapped memory pages, where each page is divided in segments. Each queue own a list of segments. The access to the each segment is intended to be done by the same thread, so no serialization is needed. The only critical section, guarded by a lock, is the interaction to require or release segment.

- close #638

### DIFF
--- a/broker/src/main/java/io/moquette/broker/unsafequeues/Queue.java
+++ b/broker/src/main/java/io/moquette/broker/unsafequeues/Queue.java
@@ -1,0 +1,420 @@
+package io.moquette.broker.unsafequeues;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.nio.Buffer;
+import java.nio.ByteBuffer;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.concurrent.locks.ReentrantLock;
+
+/**
+ * Not thread safe disk persisted queue.
+ * */
+public class Queue {
+    private static final Logger LOG = LoggerFactory.getLogger(Queue.class);
+
+    public static final int LENGTH_HEADER_SIZE = 4;
+    private final String name;
+    /* Last wrote byte, point to head byte */
+    private final AtomicReference<PointerAndSegment> head;
+
+    /* First readable byte, point to the last occupied byte */
+    private final AtomicReference<VirtualPointer> currentTailPtr;
+    private final AtomicReference<Segment> tailSegment;
+
+    private final QueuePool queuePool;
+    private final PagedFilesAllocator.AllocationListener allocationListener;
+    private final ReentrantLock lock = new ReentrantLock();
+
+    Queue(String name, Segment headSegment, VirtualPointer currentHeadPtr,
+          Segment tailSegment, VirtualPointer currentTailPtr,
+          SegmentAllocator allocator, PagedFilesAllocator.AllocationListener allocationListener, QueuePool queuePool) {
+        this.name = name;
+        this.head = new AtomicReference<>(new PointerAndSegment(currentHeadPtr, headSegment));
+        this.currentTailPtr = new AtomicReference<>(currentTailPtr);
+        this.tailSegment = new AtomicReference<>(tailSegment);
+        this.allocationListener = allocationListener;
+        this.queuePool = queuePool;
+    }
+
+    /**
+     * @throws QueueException if an error happens during access to file.
+     * */
+    public void enqueue(ByteBuffer payload) throws QueueException {
+        final PointerAndSegment res = spinningMove(LENGTH_HEADER_SIZE + payload.remaining());
+        if (res != null) {
+            // in this case all the payload is contained in the current head segment
+            LOG.trace("CAS insertion at: {}", res.pointer);
+            writeData(res.segment, res.pointer, payload);
+            LOG.trace("CAS written at {}", res.pointer);
+            return;
+        } else {
+            LOG.trace("No lock free enqueue");
+            // the payload can't be fully contained into the current head segment and needs to be splitted
+            // with another segment. To request the next segment, it's needed to be done in global lock.
+            lock.lock();
+
+            final int dataSize = payload.remaining();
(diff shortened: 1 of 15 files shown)


hylkevds:
This work looks really interesting. How far do you think you are till the first integration with Moquette?

andsel:
I think we are close, just want to change the API of `dequeue` to return `Optional.empty()` instead of null to signal the empty queue condition.

andsel:
Then integrate into Moquette, with feature flag to enable it and create a new release.

hylkevds:
I'm looking forward to performance-testing it in combination with the work from #608 (and #648). That should greatly increase the stability of long-running servers.

broker/src/main/java/io/moquette/broker/unsafequeues/Queue.java
```diff
@@ -0,0 +1,321 @@
...
+            return currentHeadPtr.moveForward(1).compareTo(currentTailPtr) == 0;
+        }
+    }
+
+    /**
+     * Read next message or return null if the queue has no data.
+     * */
+    public Optional<ByteBuffer> dequeue() throws QueueException {
+        if (!currentHeadPtr.isGreaterThan(currentTailPtr)) {
+            if (currentTailPtr.compareTo(currentHeadPtr) > 0) {
```
hylkevds: For symmetry, could use `currentTailPtr.isGreaterThan(currentHeadPtr)` instead of a direct compare.

broker/src/main/java/io/moquette/broker/unsafequeues/Queue.java
```diff
@@ -0,0 +1,321 @@
...
+            this.pointer = pointer;
+            this.payloadLength = payloadLength;
+        }
+    }
+
+    // TO BE called owning the lock
+    private CrossSegmentHeaderResult decodeCrossHeader(Segment segment, VirtualPointer pointer) throws QueueException {
+        // read first part
+        ByteBuffer lengthBuffer = ByteBuffer.allocate(LENGTH_HEADER_SIZE);
+        final ByteBuffer partialHeader = segment.readAllBytesAfter(pointer);
```
hylkevds, line 230: Would it be significantly more efficient to make the segment read methods accept an existing Buffer to put the data into, instead of having those read methods create new buffers that are then copied into the already existing buffer?

broker/src/main/java/io/moquette/broker/unsafequeues/Queue.java
```diff
@@ -0,0 +1,321 @@
...
+    private VirtualPointer currentHeadPtr;
+    private Segment headSegment;
+
+    /* First readable byte, point to the last occupied byte */
+    private VirtualPointer currentTailPtr;
+    private Segment tailSegment;
+
+    private final QueuePool queuePool;
+    private final PagedFilesAllocator.AllocationListener allocationListener;
+    private final ReentrantLock lock = new ReentrantLock();
```
hylkevds: If the queue is explicitly not thread-safe, then what is the function of this lock? It's only used in the (de)queue methods, when reading cross-segment data. The comments call it a global lock, but this seems to be a queue-local lock... I would expect a global lock to be instantiated in the QueuePool. But there it should be enough to shortly lock the QueuePool in the `nextFreeSegment()` method.

hylkevds (commented):
I've not gone though the complete code yet, but added some intermediate questions...

broker/src/main/java/io/moquette/broker/unsafequeues/Queue.java
andsel: Yes you are right!

andsel: I think that to correctly manage the handling of the queue segments the data structures used to track those has to be accessed serially.
The data structures that manage the segment allocation, free and reassignment are:
- `queueSegments`
- `recycledSegments`
in `QueuePool` class.

The access should be serialized only operating phase, when various session threads ask for next free segment.
However we can't leave uncovered by lock in other cases, like queue definition loads, during `QueuePool` instantiation, during close or during the retrieval of next tail segment of a queue.

andsel: Maybe I'm little wrong. The lock should guard only the creation or reuse of a Segment, so only `recycledSegments` are part of the critical section, not `queueSegments`.

hylkevds: Do I understand the following correctly:
- Each single Queue is only accessed by a single Thread
- The QueuePool assigns segments to Queues
- Once a Segment is assigned to a Queue, only this queue touches the segment
- A queue can return a segment to the Pool. It is then added to recycledSegments.

I suspect the lock is only needed when:
- a Queue requests a segment from the pool, since it would be bad if two Queues get the same segment.
- closing/deleting an entire page (if that ever happens) since it would be bad if a page is deleted with a segment that a Queue just received.

andsel: Yes you understand right

broker/src/main/java/io/moquette/broker/unsafequeues/QueuePool.java
```diff
@@ -0,0 +1,486 @@
...
+import java.util.concurrent.locks.ReentrantLock;
+import java.util.stream.Collectors;
+
+import static io.moquette.broker.unsafequeues.PagedFilesAllocator.PAGE_SIZE;
+
+public class QueuePool {
+
+    private static final Logger LOG = LoggerFactory.getLogger(QueuePool.class);
+
+    static final boolean queueDebug = Boolean.parseBoolean(System.getProperty("moquette.queue.debug", "false"));
```
hylkevds, line 34: Unused field?

hylkevds (approved):
Happy new Year!
Looks good. I only noticed one unused field :)

I guess the next step is to implement the different repositories based on this?

broker/src/main/java/io/moquette/broker/unsafequeues/QueuePool.java
hylkevds: Ah, no, it's used in two other classes.

hylkevds (dismissed):

PR #692, opened 2022-11-26
Reshape QueueRepository API

andsel:
## What it does?
Changes signature and name IQueueRepository service to be more fine grained, avoid the full retrieval of queue to process externally.

### DIFF
--- a/broker/src/main/java/io/moquette/broker/IQueueRepository.java
+++ b/broker/src/main/java/io/moquette/broker/IQueueRepository.java
@@ -1,13 +1,13 @@
 package io.moquette.broker;
 
-import java.util.Map;
 import java.util.Queue;
+import java.util.Set;
 
 public interface IQueueRepository {
 
-    Queue<SessionRegistry.EnqueuedMessage> createQueue(String cli, boolean clean);
+    Set<String> listQueueNames();
 
-    void removeQueue(String cli);
+    boolean containsQueue(String clientId);
 
-    Map<String, Queue<SessionRegistry.EnqueuedMessage>> listAllQueues();
+    Queue<SessionRegistry.EnqueuedMessage> getOrCreateQueue(String clientId);
 }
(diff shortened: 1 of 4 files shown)


hylkevds:
Did you have a specific reason for removing the `removeQueue` method? It's rather important for not leaking resources (#608)

andsel:
Hi @hylkevds `removeQueue` was actually an used method. So I would be useful if we identify the correct point where it has to be invoked, so in the mean time it's useless to keep the code till I have a right place where it has to be called.
It could happen that once I identify the point where the remove it's needed that method has completely another shape, for example "remove queues older than x days" instead of just plain queue removal.

hylkevds:
Yeah, remove not being used is the reason Moquette leaks memory and #608 exists.
My prototype in #648 uses the remove method.

andsel:
Ahh ok.
But at this point I don't understand why with #648 the `removeQueue` is defined in 2 places:
- [H2 queues](https://github.com/moquette-io/moquette/pull/648/files#diff-609de4ff388c0b9030b6100de2a0c1e51fb82457ad3b7b5b30c0213c8ddc8c78R44) 
- [In memory queues](https://github.com/moquette-io/moquette/pull/648/files#diff-44c900085e4a6f591ab700661b2ddf82634f7c24dfb0169b375d6a7e2eee733fR21)

but never invoked. So it seems like the method is defined but never used.

hylkevds:
Hmm, indeed. That's odd. I think a call got lost somewhere along the way...

andsel:
If you can, please create a PR to fix it :pray:

hylkevds:
Yeah, the Session is removed from the pool, the session is cleaned (in Session.cleanup()), and all messages from the session queues are deleted, but the queue itself is not removed. And since the queue is empty by this point, I never noticed, since it takes almost no resources... But that's a bug :)

Issue #695, opened 2022-11-30
[Meta] MQTT 5 support development


andsel:
This issue is a meta issue to track all the changes that needs to be done in order to support MQTT5 version of the protocol.
 
- [x] Update the Netty to the latest to have full MQTT 5 version protocol support #696
- [ ] Connection management
  - [x] accepts a v5 connection message
    - [x]  avoid to resend `PUBLISH` and `PUBREL` with a [timed handler](https://github.com/moquette-io/moquette/blob/3b2b85f924f689ca30a26093604ed6083faf37eb/broker/src/main/java/io/moquette/broker/InflightResender.java#L32), spec [[MQTT-4.4.0-1](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901238)]
"When a Client reconnects with Clean Start set to 0 and a session is present, both the Client and Server MUST resend any unacknowledged `PUBLISH` packets (where QoS > 0) and `PUBREL` packets using their original Packet Identifiers. This is the only circumstance where a Client or Server is REQUIRED to resend messages. Clients and Servers MUST NOT resend messages at any other time." #697
    - [x]  Reply to a connect with the capabilites offered (Retain Available, Maximum QoS, Server Keep Alive) in [ConnACK properties](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901080). #709
    - [x] Reply with proper `CONNACK` reason strings and reason codes. #713
    - [x]  Switch from `cleanSession` to `cleanStart` flag and `SessionExpiryInterval` property #753 and #764
  - [x] #767
 The Server delays publishing the Client’s Will Message until the Will Delay Interval has passed 
or the Session ends, whichever happens first. If a new Network Connection to this Session
 is made before the Will Delay Interval has passed, the Server MUST NOT send the Will Message [MQTT-3.1.3-9].
  - [x] #789
  - [ ] **Assigned ClientID:** In cases where the ClientID is assigned by the Server, return the assigned ClientID. This also lifts the restriction that Server assigned ClientIDs can only be used with Clean Session=1 connections.
  - [ ] **Server reference:** Allow the Server to specify an alternate Server to use on CONNACK or DISCONNECT. This can be used as a redirect or to do provisioning.
  - [ ] **Maximum Packet Size:** Allow the Client and Server to independently specify the maximum packet size they support. It is an error for the session partner to send a larger packet.

[...]

PR #697, opened 2022-12-02
Avoid to pub retries on timeout

andsel:
Originally Moquette sent retries of PUB and PUBREL packets basing on a ACK timeout. This made to re-send PUB also on open TCP connections. With [MQTT5 - 4.4.0-1](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901238) it's specified that retries **must** happen only on reconnected not clean start sessions.

## What it does
This PR stores the MQTT version into the Session instance and use that to keep the existing behavior for inflight resends (happening on a timeout basis on ACK received) in case the version is MQTT 3.1 or MQTT 3.1.1.
When the version of the Session is MQTT 5 it removes the resend on PUB ACK timeouts and switch to send only in case the same client reconnects with `cleanStart` = 0 and there is any peding publishes in the flight zone to get acknowledged.

To test this use the raw `Client` has been extended, so now can also subscribe and collect publish messages.

### DIFF
--- a/.github/workflows/maven_build.yml
+++ b/.github/workflows/maven_build.yml
@@ -5,9 +5,9 @@ name: Java CI with Maven
 
 on:
     push:
-        branches: [ main ]
+        branches: [ main, mqtt5_development ]
     pull_request:
-        branches: [ main ]
+        branches: [ main, mqtt5_development ]
 
 jobs:
     test:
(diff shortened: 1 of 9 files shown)


andsel:
Hi @hylkevds, this is one of the MQTT5 PRs to solve the items in meta #695.
It mainly changes the `MQTTConnection` and `Session` to send the retries of inflight PUB and PUBREL only when a client with same clientd and not cleanStart connects, and not on timeout basis like implemented for MQTT3.
I ask if you would like be the reviewer of this changes :-)

hylkevds:
Of course I'll be happy to review :)
May take a bit though...

andsel:
Thank you a lot, no hurry :-)

PR #704, opened 2023-01-15
Created adapter layer to use Segmented queues

andsel:
## What the PR does
Wraps the segmented queues into  session queues. This storage engine could b enabled with configuration `persistent_queue_type` setting, which accepts `h2` or `segmented`, defaulting to `segmented` if option is not specified.
Added `close` method to session registry to flush all the  meta information that compose checkpoint file of segmented queues.


andsel:
@hylkevds with this PR the segmented queues are usable from the broker.

PR #721, opened 2023-02-13
Implement and use a Session data storage

andsel:
Introduce the concept of SessionsRepository to store session's data that are not subscriptions or queues; those are already persisted with their repository instances.
This is a step to move to MQTT5 which benefit also the MQTT3 implementation.

## What does it do?
- introduce `ISessionRepository` interface and its H2 implementation
- the session data are stored in a new data class named `SessionData`
- implements serializitaion/deserialization of `SessionData` for H2

## Task list
- [x] Use `SessionData` reference inside `Session` for the data part, to avoid moving data between object instances

### DIFF
--- a/broker/src/main/java/io/moquette/broker/ISessionsRepository.java
+++ b/broker/src/main/java/io/moquette/broker/ISessionsRepository.java
@@ -1,0 +1,77 @@
+package io.moquette.broker;
+
+import io.netty.handler.codec.mqtt.MqttVersion;
+
+import java.time.Instant;
+import java.util.Collection;
+import java.util.Objects;
+
+/**
+ * Used to store data about persisted sessions like MQTT version, session's properties.
+ * */
+public interface ISessionsRepository {
+
+    // Data class
+    final class SessionData {
+        private final String clientId;
+        private final Instant created;
+        final MqttVersion version;
+        final long expiryInterval;
+
+        public SessionData(String clientId, Instant created, MqttVersion version, long expiryInterval) {
+            this.clientId = clientId;
+            this.created = created;
+            this.version = version;
+            this.expiryInterval = expiryInterval;
+        }
+
+        public String clientId() {
+            return clientId;
+        }
+
+        public MqttVersion protocolVersion() {
+            return version;
+        }
+
+        public Instant created() {
+            return created;
+        }
+
+        public long expiryInterval() {
+            return expiryInterval;
+        }
+
+        @Override
+        public boolean equals(Object o) {
+            if (this == o) return true;
+            if (o == null || getClass() != o.getClass()) return false;
+            SessionData that = (SessionData) o;
+            return clientId.equals(that.clientId);
+        }
+
+        @Override
+        public int hashCode() {
+            return Objects.hash(clientId);
+        }
+
+        @Override
+        public String toString() {
+            return "SessionData{" +
+                "clientId='" + clientId + '\'' +
+                ", created=" + created +
+                ", version=" + version +
+                ", expiryInterval=" + expiryInterval +
+                '}';
+        }
+    }
+
+    /**
+     * @return the full list of persisted sessions data.
+     * */
+    Collection<SessionData> list();
+
+    /**
+     * Save data composing a session, es MQTT version, creation date and properties but not queues or subscriptions.
+     * */
+    void saveSession(SessionData session);
+}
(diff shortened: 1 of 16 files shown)


andsel:
Hi @hylkevds I would like to ask if you could be the reviewer for this PR?

broker/src/main/java/io/moquette/broker/Session.java
```diff
@@ -46,6 +46,7 @@
 class Session {
 
     private static final Logger LOG = LoggerFactory.getLogger(Session.class);
+    static final int INFINITE_EXPIRY = 0xFFFFFFFF;
```
hylkevds: It's probably better to use `Integer.MAX_VALUE` instead of a magic number.

broker/src/main/java/io/moquette/broker/SessionRegistry.java
```diff
@@ -236,6 +243,7 @@ private Session createNewSession(MqttConnectMessage msg, String clientId) {
         }
 
         newSession.markConnecting();
+        sessionsRepository.saveSession(new ISessionsRepository.SessionData(clientId, Instant.now(), MqttVersion.MQTT_3_1_1, INFINITE_EXPIRY));
```
hylkevds: Never expiring sessions are a big problem for public-facing brokers. Users connecting with random UUIDs and cleanSession=false _will_ fill up all space until the broker crashes. In FROST I'm using a modified Moquette for this reason.
A configurable default that is not infinite would be really appreciated.

broker/src/main/java/io/moquette/broker/ISessionsRepository.java
```diff
@@ -0,0 +1,77 @@
...
+ * Used to store data about persisted sessions like MQTT version, session's properties.
+ * */
+public interface ISessionsRepository {
+
+    // Data class
+    final class SessionData {
+        private final String clientId;
+        private final Instant created;
+        final MqttVersion version;
+        final long expiryInterval;
```
hylkevds: The expiry code is not there yet, but which time instant will the `expiryInterval` be evaluated against?
Calculating it against the "created" time of the session makes little sense. I would expect it to be evaluated against the disconnect time, to get the expire time. But that would require the disconnect time in the SessionData, not the created time.

Alternatively, instead of an expiryInterval, we could store the expireInstant at which the session will expire, and set that when the connection is lost.

hylkevds (commented):
I've added some initial thoughts.

broker/src/main/java/io/moquette/broker/Session.java
andsel: It's not a magic number, it's simply an unsigned 32 bit full of 1, which `4,294,967,295`, however 32 bit int ranges between `2,147,483,647` and `-2,147,483,648`. We should save it in a long to do not loose the sign.

If we use `Integer.MAX_VALUE` we would loose all the values between `2,147,483,647` and `4,294,967,295 so if we go for it, we loose last bit. Given that `unit` is ~136 years and `Integer.MAX_VALUE` is ~68 years, I think we could use the signed integer max value.

broker/src/main/java/io/moquette/broker/SessionRegistry.java
andsel: That's part of MQTT5 migration, this PR is just to be feature parity with MQTT 3.1 but introducing the session's store.
We could address this in a follow up PR, so that setting could be read from the `moquette.conf`, but in general is a feature that's part of MQTT5 specification.

broker/src/main/java/io/moquette/broker/Session.java
hylkevds: That's pretty much exactly the definition of a "magic number" and why they are a bad idea :)
If you change it to a long, using `Long.MAX_VALUE` would be better too.

But regardless, there are clearer ways to specify the default max session duration. Something like:
```
Duration.ofDays(365*100).toMillis()
```
That both makes clear that the result is in milliseconds, and 100 years. Unfortunately the Java time classes are not very good, so saying `Duration.of(100, YEARS)` doesn't work.

broker/src/main/java/io/moquette/broker/SessionRegistry.java
andsel: requested in #732

broker/src/main/java/io/moquette/broker/Session.java
andsel: Fixed with 8fa896f83802ffae149bfd6b3396945feb912c77

broker/src/main/java/io/moquette/broker/ISessionsRepository.java
andsel: Solved with e11fd3446ccd2b8a8259cf7dda08b675f54114fb

hylkevds (approved):
Looks good so far.

PR #722, opened 2023-02-13
Fix/unsafe queues

hylkevds:
This fixes various bugs in the unsafeQueue implementation, originally from #716 :
- Head and Tail segments misbehaved when a read or write finished at exactly a Segment boundary
- Removed calls to Segment.mappedBuffer.position(), since this buffer is shared across all Segments in a Page
- Wiping of Segments overwrote the wrong parts of the page
- QueuePool was static, preventing multiple queue instances from being used at the same time

This PR bases on #724 and #728
This PR is followed by #723

### DIFF
--- a/broker/src/main/java/io/moquette/broker/unsafequeues/Queue.java
+++ b/broker/src/main/java/io/moquette/broker/unsafequeues/Queue.java
@@ -168,6 +168,9 @@ public class Queue {
             // head and tail pointer are the same, the queue is empty
             return Optional.empty();
         }
+        if (tailSegment == null) {
+            tailSegment = queuePool.openNextTailSegment(name).get();
+        }
 
         LOG.debug("currentTail is {}", currentTailPtr);
         if (containsHeader(tailSegment, currentTailPtr)) {
@@ -240,7 +243,7 @@ public class Queue {
 
         // read second part
         final int remainingHeaderSize =  LENGTH_HEADER_SIZE - consumedHeaderSize;
-        Segment nextTailSegment = queuePool.openNextTailSegment(name);
+        Segment nextTailSegment = queuePool.openNextTailSegment(name).get();
         lengthBuffer.put(nextTailSegment.read(nextTailSegment.begin, remainingHeaderSize));
         final VirtualPointer dataStart = pointer.moveForward(LENGTH_HEADER_SIZE);
         int payloadLength = ((ByteBuffer) lengthBuffer.flip()).getInt();
@@ -260,15 +263,14 @@ public class Queue {
             createdBuffers.add(buffer);
             final boolean segmentCompletelyConsumed = (segment.bytesAfter(scan) + 1) == availableDataLength;
             scan = scan.moveForward(availableDataLength);
-            final boolean consumedQueue = scan.isGreaterThan(currentHead());
             remaining -= buffer.remaining();
 
-            if (remaining > 0 || (segmentCompletelyConsumed && !consumedQueue)) {
+            if (remaining > 0 || segmentCompletelyConsumed) {
                 queuePool.consumedTailSegment(name);
                 if (QueuePool.queueDebug) {
                     segment.fillWith((byte) 'D');
                 }
-                segment = queuePool.openNextTailSegment(name);
+                segment = queuePool.openNextTailSegment(name).orElse(null);
             }
         } while (remaining > 0);
 
--- a/broker/src/main/java/io/moquette/broker/unsafequeues/Queue.java
+++ b/broker/src/main/java/io/moquette/broker/unsafequeues/Queue.java
@@ -186,7 +186,8 @@ public class Queue {
             // tail must be moved to the next byte to read, so has to move to
             // header size + payload size + 1
             final int fullMessageSize = payloadLength + LENGTH_HEADER_SIZE;
-            if (tailSegment.hasSpace(existingTail, fullMessageSize + 1)) {
+            long remainingInSegment = tailSegment.bytesAfter(existingTail) + 1;
+            if (remainingInSegment >= fullMessageSize + 1) {
                 // currentSegments contains fully the payload
                 currentTailPtr = existingTail.moveForward(fullMessageSize);
                 // read data from currentTail + 4 bytes(the length)
@@ -196,6 +197,14 @@ public class Queue {
             } else {
                 // payload is split across currentSegment and next ones
                 VirtualPointer dataStart = existingTail.moveForward(LENGTH_HEADER_SIZE);
+
+                if (remainingInSegment - LENGTH_HEADER_SIZE == 0) {
+                    queuePool.consumedTailSegment(name);
+                    if (QueuePool.queueDebug) {
+                        tailSegment.fillWith((byte) 'D');
+                    }
+                    tailSegment = queuePool.openNextTailSegment(name).get();
+                }
 
                 LOG.debug("Loading payload size {}", payloadLength);
                 return Optional.of(loadPayloadFromSegments(payloadLength, tailSegment, dataStart));
@@ -242,7 +251,7 @@ public class Queue {
         }
 
         // read second part
-        final int remainingHeaderSize =  LENGTH_HEADER_SIZE - consumedHeaderSize;
+        final int remainingHeaderSize = LENGTH_HEADER_SIZE - consumedHeaderSize;
         Segment nextTailSegment = queuePool.openNextTailSegment(name).get();
         lengthBuffer.put(nextTailSegment.read(nextTailSegment.begin, remainingHeaderSize));
         final VirtualPointer dataStart = pointer.moveForward(LENGTH_HEADER_SIZE);
@@ -258,7 +267,7 @@ public class Queue {
 
         do {
             LOG.debug("Looping remaining {}", remaining);
-            int availableDataLength = Math.min(remaining, (int) segment.bytesAfter(scan) + 1);
+            final int availableDataLength = Math.min(remaining, (int) segment.bytesAfter(scan) + 1);
             final ByteBuffer buffer = segment.read(scan, availableDataLength);
             createdBuffers.add(buffer);
             final boolean segmentCompletelyConsumed = (segment.bytesAfter(scan) + 1) == availableDataLength;
--- a/broker/src/main/java/io/moquette/broker/unsafequeues/Queue.java
+++ b/broker/src/main/java/io/moquette/broker/unsafequeues/Queue.java
@@ -187,8 +187,8 @@ public class Queue {
             // header size + payload size + 1
             final int fullMessageSize = payloadLength + LENGTH_HEADER_SIZE;
             long remainingInSegment = tailSegment.bytesAfter(existingTail) + 1;
-            if (remainingInSegment >= fullMessageSize + 1) {
-                // currentSegments contains fully the payload
+            if (remainingInSegment > fullMessageSize) {
+                // tail segment fully contains the payload with space left over
                 currentTailPtr = existingTail.moveForward(fullMessageSize);
                 // read data from currentTail + 4 bytes(the length)
                 final VirtualPointer dataStart = existingTail.moveForward(LENGTH_HEADER_SIZE);
--- a/broker/src/main/java/io/moquette/broker/unsafequeues/QueuePool.java
+++ b/broker/src/main/java/io/moquette/broker/unsafequeues/QueuePool.java
@@ -17,6 +17,7 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.Properties;
 import java.util.Set;
 import java.util.TreeSet;
@@ -426,15 +427,14 @@ public class QueuePool {
         }
     }
 
-    Segment openNextTailSegment(String name) throws QueueException {
+    Optional<Segment> openNextTailSegment(String name) throws QueueException {
         // definition from QueuePool.queueSegments
         final QueueName queueName = new QueueName(name);
         final LinkedList<SegmentRef> segmentRefs = queueSegments.get(queueName);
 
         final SegmentRef pollSegment = segmentRefs.peekLast();
         if (pollSegment == null) {
-            LOG.error("Queue segments is empty for queue {}, segment references: {}", queueName, segmentRefs);
-            throw new IllegalStateException("Opening tail segment can't never go in empty queue, because it's checked upfront in Queue");
+            return Optional.empty();
         }
 
         final Path pageFile = dataPath.resolve(String.format("%d.page", pollSegment.pageId));
@@ -451,7 +451,7 @@ public class QueuePool {
 
         final SegmentPointer begin = new SegmentPointer(pollSegment.pageId, pollSegment.offset);
         final SegmentPointer end = new SegmentPointer(pollSegment.pageId, pollSegment.offset + segmentSize - 1);
-        return new Segment(tailPage, begin, end);
+        return Optional.of(new Segment(tailPage, begin, end));
     }
 
     /**
--- a/broker/src/main/java/io/moquette/broker/unsafequeues/Segment.java
+++ b/broker/src/main/java/io/moquette/broker/unsafequeues/Segment.java
@@ -44,17 +44,20 @@ final class Segment {
     }
 
     void write(SegmentPointer offset, ByteBuffer content) {
-        final ByteBuffer buffer = (ByteBuffer) mappedBuffer.position(offset.offset());
         checkContentStartWith(content);
-        buffer.put(content);
+        final int startPos = offset.offset();
+        final int endPos = startPos + content.remaining();
+        for (int i = startPos; i < endPos; i++) {
+            mappedBuffer.put(i, content.get());
+        }
     }
 
     // fill the segment with value bytes
     void fillWith(byte value) {
         LOG.debug("Wipe segment {}", this);
-        final ByteBuffer buffer = (ByteBuffer) mappedBuffer.position(this.begin.offset());
-        for (int i = 0; i < size(); i++) {
-            buffer.put(i, value);
+        final int target = begin.offset() + (int)size();
+        for (int i = begin.offset(); i < target; i++) {
+            mappedBuffer.put(i, value);
         }
     }
 
@@ -66,9 +69,11 @@ final class Segment {
     }
 
     void write(VirtualPointer offset, ByteBuffer content) {
-        final int pageOffset = rebasedOffset(offset);
-        final ByteBuffer buffer = (ByteBuffer) mappedBuffer.position(pageOffset);
-        buffer.put(content);
+        final int startPos = rebasedOffset(offset);
+        final int endPos = startPos + content.remaining();
+        for (int i = startPos; i < endPos; i++) {
+            mappedBuffer.put(i, content.get());
+        }
     }
 
     /**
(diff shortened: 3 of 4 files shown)


broker/src/main/java/io/moquette/broker/unsafequeues/QueuePool.java
```diff
@@ -435,8 +435,7 @@ Segment openNextTailSegment(String name) throws QueueException {
 
         final SegmentRef pollSegment = segmentRefs.peekLast();
         if (pollSegment == null) {
-            LOG.error("Queue segments is empty for queue {}, segment references: {}", queueName, segmentRefs);
-            throw new IllegalStateException("Opening tail segment can't never go in empty queue, because it's checked upfront in Queue");
+            return null;
```
andsel: I prefer to avoid using `null` to notify the caller that no other segments are available for the queue.
In such case the `Optional` is more suited.

broker/src/main/java/io/moquette/broker/unsafequeues/Segment.java
```diff
@@ -66,9 +69,11 @@ private void checkContentStartWith(ByteBuffer content) {
     }
 
     void write(VirtualPointer offset, ByteBuffer content) {
-        final int pageOffset = rebasedOffset(offset);
-        final ByteBuffer buffer = (ByteBuffer) mappedBuffer.position(pageOffset);
-        buffer.put(content);
+        final int startPos = rebasedOffset(offset);
```
andsel, line 72: Would prefer to keep the original name `pageOffset` instead of `startPos`

broker/src/main/java/io/moquette/broker/unsafequeues/Segment.java
```diff
@@ -44,9 +44,12 @@ public long bytesAfter(VirtualPointer mark) {
...
 
     void write(SegmentPointer offset, ByteBuffer content) {
-        final ByteBuffer buffer = (ByteBuffer) mappedBuffer.position(offset.offset());
         checkContentStartWith(content);
-        buffer.put(content);
+        final int startPos = offset.offset();
+        final int endPos = startPos + content.remaining();
+        for (int i = startPos; i < endPos; i++) {
+            mappedBuffer.put(i, content.get());
+        }
```
andsel, line 52: This is duplicated so could be extracted in a method.

broker/src/main/java/io/moquette/broker/unsafequeues/QueuePool.java
```diff
@@ -34,7 +34,7 @@ public class QueuePool {
 
     static final boolean queueDebug = Boolean.parseBoolean(System.getProperty("moquette.queue.debug", "false"));
 
-    private static SegmentAllocationCallback callback;
```
andsel: `QueuePool` is intended to be static, so one queue pool per process, why is the intention to avoid a singleton, do you expect to use multiple pools?

broker/src/main/java/io/moquette/broker/unsafequeues/Segment.java
```diff
@@ -55,9 +55,9 @@ void write(SegmentPointer offset, ByteBuffer content) {
     // fill the segment with value bytes
     void fillWith(byte value) {
         LOG.debug("Wipe segment {}", this);
-        final ByteBuffer buffer = (ByteBuffer) mappedBuffer.position(this.begin.offset());
-        for (int i = 0; i < Segment.SIZE; i++) {
-            buffer.put(i, value);
+        final int target = begin.offset()+Segment.SIZE;
```
andsel: As before this could be a shared method

andsel (changes requested):
Hi @hylkevds thanks for your contribution.
I left some comment, mainly a doubt to use `null` to say "no more segments for the queue" from the `openNextTailSegment` in `QueuePool`.

I think this should be covered with tests, and I can understand the tests are present in the PR #723 and #724.
The #724 is the one that parameterize the size of segments and queue to ease the tests. So I think would be better to first have such PR to be merged, then a test that exposes the problem, and finally the code the fix. The last 2 steps  could be merged in one PR.
 
Sorry to ask you other effort, but it would greatly help me make evident the problem and the fix.

broker/src/main/java/io/moquette/broker/unsafequeues/QueuePool.java
hylkevds: I can quite easily imagine the need for multiple pools. One may want to run multiple instances of Moquette side-by-side, or use the queue pool for other reasons in the project Moquette is embedded in. I can even see this queue pool becoming useful as a separate library.

But mainly, there is also no reason to limit things to only one pool. The change is only minimal.

andsel: Ok, I thought in its first instance as a Moquette internal, but understand your proposal. Maybe that aspect could be separated in a separate PR

hylkevds:
> Hi @hylkevds thanks for your contribution. I left some comment, mainly a doubt to use `null` to say "no more segments for the queue" from the `openNextTailSegment` in `QueuePool`.
> 
> I think this should be covered with tests, and I can understand the tests are present in the PR #723 and #724. The #724 is the one that parameterize the size of segments and queue to ease the tests. So I think would be better to first have such PR to be merged, then a test that exposes the problem, and finally the code the fix. The last 2 steps could be merged in one PR.

I can refactor the PRs to put the test and flexible-size pages commit at the front. 
The test will fail until the last commit. The error message is not as helpful as you might hope, since the test can only detect that there is corruption of the queue, but not point at the cause of the corruption. Usually because the corruption happened several cycles before it becomes evident.

andsel:
> I can refactor the PRs to put the test and flexible-size pages commit at the front.

Yes please, it's important to separate the PR that parameterize the segment/page size, from the one that introduces the test that fails

broker/src/main/java/io/moquette/broker/unsafequeues/Segment.java
hylkevds: I would ignore these duplicate loops for now, since if we change the Segment to have its own slice, the loops will go away completely, and be replaced by a single call to slice.put(content);

hylkevds:
> Hi @hylkevds thanks for your contribution. I left some comment, mainly a doubt to use `null` to say "no more segments for the queue" from the `openNextTailSegment` in `QueuePool`.

`QueuePool.openNextTailSegment()` now returns an Optional.

broker/src/main/java/io/moquette/BrokerConstants.java
```diff
@@ -32,6 +32,10 @@ public final class BrokerConstants {
     public static final String DATA_PATH_PROPERTY_NAME = "data_path";
     public static final String PERSISTENT_QUEUE_TYPE_PROPERTY_NAME = "persistent_queue_type"; // h2 or segmented, default h2
     public static final String PERSISTENCE_ENABLED_PROPERTY_NAME = "persistence_enabled"; // true or false, default true
+    public static final String SEGMENTED_QUEUE_PAGE_SIZE = "queue_page_size";
+    public static final int DEFAULT_SEGMENTED_QUEUE_PAGE_SIZE = 64 * 1024 * 1024;
+    public static final String SEGMENTED_QUEUE_SEGMENT_SIZE = "queue_segment_size";
+    public static final int DEFAULT_SEGMENTED_QUEUE_SEGMENT_SIZE = 4 * 1024 * 1024;
```
andsel:
```suggestion
    public static final String SEGMENTED_QUEUE_PAGE_SIZE = "queue_page_size";
    public static final int MB = 1024 * 1024;    
    public static final int DEFAULT_SEGMENTED_QUEUE_PAGE_SIZE = 64 * MB;
    public static final String SEGMENTED_QUEUE_SEGMENT_SIZE = "queue_segment_size";
    public static final int DEFAULT_SEGMENTED_QUEUE_SEGMENT_SIZE = 4 * MB;
```

hylkevds:
> > I can refactor the PRs to put the test and flexible-size pages commit at the front.
> 
> Yes please, it's important to separate the PR that parameterize the segment/page size, from the one that introduces the test that fails

I've refactored the set of PRs, and split out some more parts.
The order is now:
- [x] #724: Configurable Queue Sizes
- [x] #728: Test for SegmentedQueues (new)
- [ ] #722: Fixes for the core Queue
- [ ] #723: Fix for the SegmentPersisted Queue

Separate are now:
- [ ] #725: Moved some Session calls to the command queue
- [ ] #726: Avoid attempt at purge of already purged session
- [ ] #727: Make QueuePool.callback non-static, but final (new)

andsel:
Hi @hylkevds, please split this PR at the commit  8341efc

The other 3 commits: c9328c8 e888fa4  and 73d6488 should go in another PR. 
In this way the reviewing process becomes easier for me :pray:

andsel:
So this PR could be considered as closed?

hylkevds:
> So this PR could be considered as closed?

Depends if you want to apply all these commits in one go, or as separate merges. The PRs depend on each other, so the next one always also contains the commits of the one before.
If you merge #723 you also merge the commits of #724, #728 and #722
But you can first review #724, then #728, then #722 and finally #723 to keep them separate

andsel:
@hylkevds #724 and #722 has been merged, please rebase on `main`

broker/src/main/java/io/moquette/broker/unsafequeues/Queue.java
```diff
@@ -168,6 +168,9 @@ public Optional<ByteBuffer> dequeue() throws QueueException {
             // head and tail pointer are the same, the queue is empty
             return Optional.empty();
         }
+        if (tailSegment == null) {
```
andsel, line 171: I don't like too much the null check, having that `openNextTailSegment` could now return an  `Optional`

andsel (commented):
I've a doubt in continue using the null check on `tailSegment`

broker/src/main/java/io/moquette/BrokerConstants.java
hylkevds: That was already changed, by the rebase.

broker/src/main/java/io/moquette/broker/unsafequeues/Queue.java
hylkevds: `tailSegment` is accessed very often. Changing `Queue.tailSegment` into an optional wouldn't be good for performance, and spread lots of `tailSegment.get()` calls through the class. All of these `get()` calls would just cause an exception when the optional is empty, since that would mean the queue is corrupt.

openNextTailSegment does now return an optional, so the `null` is a Queue internal value that is never communicated to the outside, except when the queue is corrupt and a NullPointerException happens.

andsel: I  see, maybe we can switch to NullObject pattern in follow up PR.

broker/src/main/java/io/moquette/broker/unsafequeues/Queue.java
```diff
@@ -183,7 +186,8 @@ public Optional<ByteBuffer> dequeue() throws QueueException {
             // tail must be moved to the next byte to read, so has to move to
             // header size + payload size + 1
             final int fullMessageSize = payloadLength + LENGTH_HEADER_SIZE;
-            if (tailSegment.hasSpace(existingTail, fullMessageSize + 1)) {
+            long remainingInSegment = tailSegment.bytesAfter(existingTail) + 1;
+            if (remainingInSegment >= fullMessageSize + 1) {
```
andsel: It's not clear to me why the `hasSpace` isn't sufficient for the condition, please could you expand more on this?
It's not just an inline of `hasSpace` it contains a `+1` that wasn't present before.

broker/src/main/java/io/moquette/broker/unsafequeues/Queue.java
```diff
@@ -255,20 +267,20 @@ private ByteBuffer loadPayloadFromSegments(int remaining, Segment segment, Virtu
 
         do {
             LOG.debug("Looping remaining {}", remaining);
-            int availableDataLength = Math.min(remaining, (int) segment.bytesAfter(scan) + 1);
+            final long bytesAfterScan = segment.bytesAfter(scan);
```
andsel: Is there a motivation to extract in the variable `bytesAfterScan` the result of  `segment.bytesAfter(scan)` or is just a form of local caching?

broker/src/main/java/io/moquette/broker/unsafequeues/Queue.java
```diff
@@ -255,20 +267,20 @@ private ByteBuffer loadPayloadFromSegments(int remaining, Segment segment, Virtu
...
+            final int availableDataLength = Math.min(remaining, (int) bytesAfterScan + 1);
             final ByteBuffer buffer = segment.read(scan, availableDataLength);
             createdBuffers.add(buffer);
-            final boolean segmentCompletelyConsumed = (segment.bytesAfter(scan) + 1) == availableDataLength;
+            final boolean segmentCompletelyConsumed = (bytesAfterScan + 1) == availableDataLength;
             scan = scan.moveForward(availableDataLength);
-            final boolean consumedQueue = scan.isGreaterThan(currentHead());
```
andsel, line 263: `consumedQueue` is not such useful for the sake of the loading, but would be helpful to check for bad condition that should never happen.
We shouldn't never been in such a case where `remaining > 0 && consumedQueue`.

I think we could insert a security check before the next `if` statement such as:
```java
if (remaining > 0 && consumedQueue) {
    throw new QueueException("Queue " + name + " has been completely consumed while still has some " + remaining + " bytes to be read");
}

**Review by andsel — COMMENTED** [review]
I've left some doubts I have, if you could clarify me, would be great :-)

File: broker/src/main/java/io/moquette/broker/unsafequeues/Queue.java (hunk 8 above)
Comment (hylkevds) [reply broker/src/main/java/io/moquette/broker/unsafequeues/Queue.java]: The +1 has something to do with the exact position of the read pointer versus the write pointer. I didn't change anything in those semantics. The code in the `if` does exactly the same as it did before, except now with a direct call to `bytesAfter` instead of wrapping it in `hasSpace`.

I couldn't use `hasSpace`, since that returns a boolean, and I need the exact number. If the read will end exactly at the Segment boundary, the segment has to be returned to the pool. Otherwise the next read will start re-reading the segment we just finished instead of opening the next segment. The same thing happened when the write ended at exactly the Segment boundary.

In the end, these two bugs come down to the method `VirtualPointer.segmentOffset()`. That returns 0 when the pointer is at the end of the Segment, and thus the calling code thinks the pointer is at the beginning of the Segment and doesn't go to the next segment. Therefore we must specifically check if a read or write ends at the segment boundary, and ensure we move to the next Segment (or at least release this Segment)

If we change to Segments with their own buffer slice, we can probably simplify this a lot, since that buffer slice will have an exact read and write position in the buffer and we no longer need to calculate it using `VirtualPointer.segmentOffset()`.

File: broker/src/main/java/io/moquette/broker/unsafequeues/Queue.java (hunk 9 above)
Comment (hylkevds) [reply broker/src/main/java/io/moquette/broker/unsafequeues/Queue.java]: Just local caching. `bytesAfter` is not really cheap, since it creates and discards two new Objects. So caching the value instead of calling it several times is better.

File: broker/src/main/java/io/moquette/broker/unsafequeues/Queue.java (hunk 10 above)
Comment (hylkevds) [reply broker/src/main/java/io/moquette/broker/unsafequeues/Queue.java:263]: Yes, we could add an explicit check there. On the other hand, currently that situation will throw a `BufferUnderflowException`. So there is no direct danger that that form of corruption will go unnoticed.

File: broker/src/main/java/io/moquette/broker/unsafequeues/Queue.java (hunk 9 above)
Comment (andsel) [reply broker/src/main/java/io/moquette/broker/unsafequeues/Queue.java]: I would avoid such kind of optimizations and go with a more explicit and declarative way.
Usually, the JVM's JIT is able to inline calls such these.

File: broker/src/main/java/io/moquette/broker/unsafequeues/Queue.java (hunk 10 above)
Comment (andsel) [reply broker/src/main/java/io/moquette/broker/unsafequeues/Queue.java:263]: Good point! So no need for explicit check :+1:

File: broker/src/main/java/io/moquette/broker/unsafequeues/Queue.java (hunk 8 above)
Comment (andsel) [reply broker/src/main/java/io/moquette/broker/unsafequeues/Queue.java]: Right,  I could with the motivation, but maybe the in-lining  of `hasSpace` is not correct.
`hadSpace` is defined as:
```java
bytesAfter(mark) >= length
```

the in-lined if becomes:
```java
tailSegment.bytesAfter(existingTail) >= fullMessageSize + 1
```
so extracting the `tailSegment.bytesAfter(existingTail)` into the variable `remainingInSegment` would become:
```java
final long remainingSegment = tailSegment.bytesAfter(existingTail);
if (remainingSegment >= fullMessageSize + 1) {
  ...
}
```

Maybe I'm wrong, but the proposed change seems to be off-by-one.

andsel (commented):
@hylkevds thank's a ton for all your explanation. I've only left a doubt on a possible off by one error.

hylkevds:
Hmm, let me go over that possible off-by-one again.
Old code:
```
if (tailSegment.hasSpace(existingTail, fullMessageSize + 1))
```
Expands through `bytesAfter(mark) >= length` using `mark => existingTail` and `length => fullMessageSize + 1` to:
```
tailSegment.bytesAfter(existingTail) >= fullMessageSize + 1
```
But we now have 
```
tailSegment.bytesAfter(existingTail) + 1
```

Hmm,  indeed, there seems to be a +1 too many... But why does it not break the Tests. They actually validate the message content. Removing either of the two +1's breaks the test (yay for tests!) so apparently it's correct... but I can't remember why I changed it! I know I've been stepping through this code in debug mode, with specific message sizes, to validate what happens. Maybe i changed it and forgot to list it as a bug that was fixed.

I'll have to have a closer look at it again...

hylkevds:
Debugging the potential off-by-one problem. Using a segment size of 1000, and an in-queue message size of 500.
Reading the first message results in a `remainingInSegment` of 1000 with the `+1`. `tailSegment.bytesAfter(existingTail)` returns 999. So the first `+1` is correct.

The second one then also becomes clear. The block in that `if` should only run if it is a simple read without any side effects. In practice that means if there is at least one byte left after we read the full message. If the segment is exactly emptied by the read, we already have a special case, and have to use the second block.

I propose we change the `if` to just a `>` without the `+1` instead of a `>=` with a `+1`. And change the comment for that block to `currentSegments contains fully the payload _with space left over_`
That should be clearer.

broker/src/main/java/io/moquette/broker/unsafequeues/Queue.java
hylkevds: I've removed that change.

broker/src/main/java/io/moquette/broker/unsafequeues/Queue.java
```diff
@@ -183,8 +186,9 @@ public Optional<ByteBuffer> dequeue() throws QueueException {
             // tail must be moved to the next byte to read, so has to move to
             // header size + payload size + 1
             final int fullMessageSize = payloadLength + LENGTH_HEADER_SIZE;
-            if (tailSegment.hasSpace(existingTail, fullMessageSize + 1)) {
-                // currentSegments contains fully the payload
+            long remainingInSegment = tailSegment.bytesAfter(existingTail) + 1;
+            if (remainingInSegment > fullMessageSize) {
+                // currentSegments contains fully the payload with space left over
```
andsel:
```suggestion
                // tail segment contains fully the payload with space left over
```

andsel (dismissed):
LGTM just one nit pick in the comment.
Great work and thanks for all your patience :-)

broker/src/main/java/io/moquette/broker/unsafequeues/Queue.java
hylkevds: Fixed now.
I've updated it in the previous commit, to avoid having several commits that change things back and forth.
I've also merged the "revert" commit into the commit that made the original change for clarity.

PR #723, opened 2023-02-13
Fix/segment persistent queues

hylkevds:
This PR builds on #724, #728 and #722 and contains further fixes from #716:

- Make a copy of the message buffer (one for each command thread) so access patterns don't interfere across threads


andsel:
@hylkevds please can you fix the conflicts realigning to the current `main` so that we can merge it? :pray:

andsel (approved):
LGTM

PR #724, opened 2023-02-13
Fix/configurable queue sizes

hylkevds:
This PR makes the PageSize and SegmentSize configurable.

This makes testing easier, since certain bugs appear faster on smaller queues. Also, for low-volume high-user-count servers, smaller segments are more efficient, since at least one Segment is used for each user.

This is the first UnsafeQueue related PR from #716.
PRs #728, #722 and #723 will build on this one

### DIFF
--- a/broker/src/main/java/io/moquette/BrokerConstants.java
+++ b/broker/src/main/java/io/moquette/BrokerConstants.java
@@ -32,6 +32,10 @@ public final class BrokerConstants {
     public static final String DATA_PATH_PROPERTY_NAME = "data_path";
     public static final String PERSISTENT_QUEUE_TYPE_PROPERTY_NAME = "persistent_queue_type"; // h2 or segmented, default h2
     public static final String PERSISTENCE_ENABLED_PROPERTY_NAME = "persistence_enabled"; // true or false, default true
+    public static final String SEGMENTED_QUEUE_PAGE_SIZE = "queue_page_size";
+    public static final int DEFAULT_SEGMENTED_QUEUE_PAGE_SIZE = 64 * 1024 * 1024;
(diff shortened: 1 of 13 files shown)


broker/src/main/java/io/moquette/BrokerConstants.java
```diff
@@ -32,6 +32,10 @@ public final class BrokerConstants {
     public static final String DATA_PATH_PROPERTY_NAME = "data_path";
     public static final String PERSISTENT_QUEUE_TYPE_PROPERTY_NAME = "persistent_queue_type"; // h2 or segmented, default h2
     public static final String PERSISTENCE_ENABLED_PROPERTY_NAME = "persistence_enabled"; // true or false, default true
+    public static final String SEGMENTED_QUEUE_PAGE_SIZE = "queue_page_size";
+    public static final int DEFAULT_SEGMENTED_QUEUE_PAGE_SIZE = 64 * 1024 * 1024;
+    public static final String SEGMENTED_QUEUE_SEGMENT_SIZE = "queue_segment_size";
+    public static final int DEFAULT_SEGMENTED_QUEUE_SEGMENT_SIZE = 4 * 1024 * 1024;
```
andsel:
```suggestion
    public static final int MB = 1024 * 1024;
    public static final int DEFAULT_SEGMENTED_QUEUE_PAGE_SIZE = 64 * MB;
    public static final String SEGMENTED_QUEUE_SEGMENT_SIZE = "queue_segment_size";
    public static final int DEFAULT_SEGMENTED_QUEUE_SEGMENT_SIZE = 4 * MB;
```

andsel (changes requested):
LGTM, jus one nitpick to extract the MB constant

andsel (approved):
LGTM

PR #725, opened 2023-02-13
Moved session calls to the session-queue

hylkevds:
There were some non-queue related threading issues that popped up during testing of the Unsafe Queues. Some Session actions were not put on the command queue.

### DIFF
--- a/broker/src/main/java/io/moquette/broker/MQTTConnection.java
+++ b/broker/src/main/java/io/moquette/broker/MQTTConnection.java
@@ -236,7 +236,11 @@ final class MQTTConnection {
                         // OK continue with sending queued messages and normal flow
 
                         if (result.mode == SessionRegistry.CreationModeEnum.REOPEN_EXISTING) {
-                            result.session.sendQueuedMessagesWhileOffline();
+                            final Session session = result.session;
+                            postOffice.routeCommand(session.getClientID(), "sendOfflineMessages", () -> {
+                                session.sendQueuedMessagesWhileOffline();
+                                return null;
+                            });
                         }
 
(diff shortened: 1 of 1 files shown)


broker/src/main/java/io/moquette/broker/MQTTConnection.java
```diff
@@ -543,7 +547,7 @@ void sendIfWritableElseDrop(MqttMessage msg) {
     public void writabilityChanged() {
         if (channel.isWritable()) {
             LOG.debug("Channel is again writable");
-            bindedSession.writabilityChanged();
+            queueDrainQueueCommand();
```
andsel: Why invoking asynchronously the `flushAllQueuedMessages` method instead of keeping the original `writabilityChanged` ?

andsel (changes requested):
Left a comment on a possible mismatch.

broker/src/main/java/io/moquette/broker/MQTTConnection.java
hylkevds: Indeed, that like should not have been in this commit.

hylkevds: Ah, actually, it does the same.
It's a call to the session, so it must happen in the command loop. And since `bindedSession.writabilityChanged();` in turn only calls `drainQueueToConnection()` we can just use the same function to directly call that method through the command queue.

hylkevds: I've changed it back to a direct call to `bindedSession.writabilityChanged()`, wrapped in a queue call.

andsel (approved):
LGTM

PR #726, opened 2023-02-13
Avoid attempt at purge of already purged session

hylkevds:
The purge method can be called from both the disconnect message handler and from the connectionLost handler. If the disconnect happened normally then the connectionLost handler doesn't need to purge any more. Thus there is no need to throw an Exception.


andsel:
Hi @hylkevds I think this is a duplicated of #735, sorry

andsel:
Do you think we can close it?

hylkevds:
Yeah, looks like #735 fixes the problem too.
:+1:

PR #727, opened 2023-02-17
Make QueuePool.callback non-static, but final

hylkevds:
There may be multiple queue pools, a static callback breaks this.


andsel:
@hylkevds would be great if you can fix the conflicts so that later we can merge it :+1:

PR #728, opened 2023-02-17
Fix/add test for segment persistent queues

hylkevds:
Added a test class for the SegmentPersistentQueues.
This builds on #724.
This is followed by #722 and #723.

### DIFF
--- a/broker/src/test/java/io/moquette/persistence/SegmentPersistentQueueTest.java
+++ b/broker/src/test/java/io/moquette/persistence/SegmentPersistentQueueTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2012-2018 The original author or authors
+ * Copyright (c) 2012-2023 The original author or authors
  * ------------------------------------------------------
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
--- a/broker/src/test/java/io/moquette/persistence/SegmentPersistentQueueTest.java
+++ b/broker/src/test/java/io/moquette/persistence/SegmentPersistentQueueTest.java
@@ -1,0 +1,266 @@
+/*
+ * Copyright (c) 2012-2018 The original author or authors
+ * ------------------------------------------------------
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * and Apache License v2.0 which accompanies this distribution.
+ *
+ * The Eclipse Public License is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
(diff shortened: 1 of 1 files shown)


andsel:
Hi @hylkevds, #724 has been merge,please could you rebase this on `main` ?

broker/src/test/java/io/moquette/persistence/SegmentPersistentQueueTest.java
```diff
@@ -0,0 +1,266 @@
+/*
+ * Copyright (c) 2012-2018 The original author or authors
```
andsel:
```suggestion
 * Copyright (c) 2012-2023 The original author or authors
```

broker/src/test/java/io/moquette/persistence/SegmentPersistentQueueTest.java
```diff
@@ -0,0 +1,266 @@
...
+
+    private int queueIndex = 0;
+    private static final String QUEUE_NAME = "test";
+
+    @TempDir
+    static Path tempQueueFolder;
+
+    @BeforeAll
+    public static void beforeAll() throws IOException, QueueException {
+        System.setProperty("moquette.queue.debug", "false");
```
andsel, line 60: It should be already false, because false is the default value when the property is not set.

broker/src/test/java/io/moquette/persistence/SegmentPersistentQueueTest.java
```diff
@@ -0,0 +1,266 @@
...
+        msg1.release();
+    }
+
+    private void createAndAddToQueues(String topic, int totalSize) {
+        final PublishedMessage msg1 = createMessage(topic, totalSize);
+        for (SessionMessageQueue<EnqueuedMessage> queue : queues) {
+            msg1.retain();
+            queue.enqueue(msg1);
+        }
+        msg1.release();
```
andsel:
```suggestion
        final PublishedMessage msg = createMessage(topic, totalSize);
        for (SessionMessageQueue<EnqueuedMessage> queue : queues) {
            msg.retain();
            queue.enqueue(msg);
        }
        msg.release();
```

andsel (changes requested):
Left a couple of nitpicks, after that it's good to merge

broker/src/test/java/io/moquette/persistence/SegmentPersistentQueueTest.java
hylkevds: True, but when debugging tests it is easier to turn it on when the option is already there. Less searching for what it is called and how to set it.

andsel (approved):
LGTM

PR #738, opened 2023-03-04
Switched boolean flag to immediate flushing to integer

### LINKED ISSUE #718: immediate_buffer_flush should be in configuration file, defaulting to true.
### Expected behavior
Correct performance behavior, similar to Mosquitto or Aedes (less than 20ms for message to propagate from emitter to subscriber on localhost).
The parameter `immediate_buffer_flush` is not listed in the default conf file, and not mentioned in the documentation.
The default value does not seem appropriate for normal use cases.

### Actual behavior
Requests are buffered, resulting in delays of 1 second.

### Steps to reproduce
Run with default configuration.

### Moquette MQTT version
0.16, 0.17-SNAPSHOT

n/a

n/a

andsel:
## Release notes
Introduces the ability to specify the flush interval for IO write operations, changing the default, to an immediate flush.

## What does it do?
- Changed the value of configuration setting `immediate_buffer_flush` from boolean to a number (expressing milliseconds)
- Added 'buffer_flush_millis' setting which overrides the deprecated 'immediate_buffer_flush'
- Changed default behavior from "no flush on every write" to flush on each write.
- Used the value of the new `buffer_flush_millis` settings to initialize the auto flusher.

Fixes #718

### DIFF
--- a/broker/src/main/java/io/moquette/broker/BrokerConfiguration.java
+++ b/broker/src/main/java/io/moquette/broker/BrokerConfiguration.java
@@ -37,7 +37,7 @@ class BrokerConfiguration {
         if (bufferFlushMillisProp != null && !bufferFlushMillisProp.isEmpty()) {
             switch (bufferFlushMillisProp.toLowerCase(Locale.ROOT)) {
                 case "immediate":
-                    bufferFlushMillis = BrokerConstants.IMMEDIATE_BUFFER_FLUSH;
+                    bufferFlushMillis = BrokerConstants. IMMEDIATE_BUFFER_FLUSH;
                     break;
                 case "full":
                     bufferFlushMillis = BrokerConstants.NO_BUFFER_FLUSH;
@@ -55,7 +55,7 @@ class BrokerConfiguration {
                     }
             }
         } else {
-            if (props.boolProp(BrokerConstants.IMMEDIATE_BUFFER_FLUSH_PROPERTY_NAME, false)) {
+            if (props.boolProp(BrokerConstants.IMMEDIATE_BUFFER_FLUSH_PROPERTY_NAME, true)) {
                 bufferFlushMillis = BrokerConstants.IMMEDIATE_BUFFER_FLUSH;
             } else {
                 bufferFlushMillis = BrokerConstants.NO_BUFFER_FLUSH;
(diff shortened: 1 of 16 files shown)


hylkevds (commented):
Looks good.
You've not updated the autoflush handler in NewNettyAcceptor yet, right?

andsel:
@hylkevds have to do, I'm going with incremental commits, to make it easy reviewable. Still work in progress, once ready I'll ask you as reviewer.

broker/src/main/java/io/moquette/broker/BrokerConfiguration.java
```diff
@@ -18,26 +18,57 @@
...
+                        }
+                    } catch (NumberFormatException ex) {
+                        throw new IllegalArgumentException(errorMsg);
+                    }
+            }
+        } else {
+            if (props.boolProp(BrokerConstants.IMMEDIATE_BUFFER_FLUSH_PROPERTY_NAME, true)) {
+                bufferFlushMillis = BrokerConstants.IMMEDIATE_BUFFER_FLUSH;
+            } else {
+                bufferFlushMillis = BrokerConstants.NO_BUFFER_FLUSH;
```
hylkevds, line 61: Maybe it's better to default to the 1s value that used to be used when `immediate_buffer_flush` was `false`?

hylkevds (dismissed):
Looks good!

broker/src/main/java/io/moquette/broker/BrokerConfiguration.java
andsel: Good point, done :+1:

jflamy: If you default to 1000ms people coming from Mosquitto will keep reporting bad performance

andsel: The old  `immediate_buffer_flush` default value is now switched from `false` to `true`.
So by default it now flushes immediatly, then if the user customize the `immediate_buffer_flush` to true, the flush interval is 1000ms

jflamy: You mean false I presume.immediate False = 1000; immediate true = 0.  Explicit numerical value ignored if immediate is true.

PR #739, opened 2023-03-20
Implement session expiry cleanup for persisted sessions

andsel:
## Release notes
Implement the logic to expire the not clean sessions.

## What does it do?
Track all expiration time after a session is closed or disconnects. Uses a delay queue to track the not clean session that expires and using a scheduled task, every second, check for the expired sessions and removes from the both SessionRegistry and also from the persistent store (SessionRepositoy).

## Why it's important for the user?
This feature let the user to auto-purge the state of  the not clean session after a certain amount of time configured at broker level in `moquette.conf` file through the `persistent_client_expiration` property which could have the format:
 `<number><unit>` such as 1d for 1 day. Admitted values to unit are `s` `m` `h` `d` `w` `M` `y` which respectively means seconds, minutes, hours, days, weeks, months and years. If not specified this global expiry defaults to INFINITE_EXPIRY which is ~80 years.

- close #662

### DIFF
--- a/broker/src/main/java/io/moquette/broker/SessionRegistry.java
+++ b/broker/src/main/java/io/moquette/broker/SessionRegistry.java
@@ -34,6 +34,7 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.ScheduledExecutorService;
 import java.util.stream.Collectors;
 
 import static io.moquette.broker.Session.INFINITE_EXPIRY;
@@ -122,15 +123,18 @@ public class SessionRegistry {
     private final ISessionsRepository sessionsRepository;
     private final IQueueRepository queueRepository;
     private final Authorizator authorizator;
+    private final ScheduledExecutorService scheduler;
 
     SessionRegistry(ISubscriptionsDirectory subscriptionsDirectory,
                     ISessionsRepository sessionsRepository,
                     IQueueRepository queueRepository,
-                    Authorizator authorizator) {
+                    Authorizator authorizator,
+                    ScheduledExecutorService scheduler) {
         this.subscriptionsDirectory = subscriptionsDirectory;
         this.sessionsRepository = sessionsRepository;
         this.queueRepository = queueRepository;
         this.authorizator = authorizator;
+        this.scheduler = scheduler;
         recreateSessionPool();
     }
 
--- a/broker/src/main/java/io/moquette/broker/SessionRegistry.java
+++ b/broker/src/main/java/io/moquette/broker/SessionRegistry.java
@@ -28,6 +28,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.net.InetSocketAddress;
+import java.time.Clock;
 import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -132,17 +133,29 @@ public class SessionRegistry {
     private final IQueueRepository queueRepository;
     private final Authorizator authorizator;
     private final DelayQueue<ISessionsRepository.SessionData> removableSessions = new DelayQueue<>();
+    private final Clock clock;
 
+    // Used in testing
     SessionRegistry(ISubscriptionsDirectory subscriptionsDirectory,
                     ISessionsRepository sessionsRepository,
                     IQueueRepository queueRepository,
                     Authorizator authorizator,
                     ScheduledExecutorService scheduler) {
+        this(subscriptionsDirectory, sessionsRepository, queueRepository, authorizator, scheduler, Clock.systemDefaultZone());
+    }
+
+    SessionRegistry(ISubscriptionsDirectory subscriptionsDirectory,
+                    ISessionsRepository sessionsRepository,
+                    IQueueRepository queueRepository,
+                    Authorizator authorizator,
+                    ScheduledExecutorService scheduler,
+                    Clock clock) {
         this.subscriptionsDirectory = subscriptionsDirectory;
         this.sessionsRepository = sessionsRepository;
         this.queueRepository = queueRepository;
         this.authorizator = authorizator;
         this.scheduledExpiredSessions = scheduler.scheduleWithFixedDelay(this::checkExpiredSessions, 1, 1, TimeUnit.SECONDS);
+        this.clock = clock;
         recreateSessionPool();
     }
 
@@ -154,6 +167,7 @@ public class SessionRegistry {
             final String expiredAt = expiredSession.expireAt().map(Instant::toString).orElse("UNDEFINED");
             LOG.debug("Removing session {}, expired on {}", expiredSession.clientId(), expiredAt);
             remove(expiredSession.clientId());
+            sessionsRepository.delete(expiredSession);
         }
     }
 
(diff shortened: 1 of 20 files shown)


andsel:
Hi @hylkevds may I ask you if you would like to be my buddy in this effort and provide a review to this functionality, please :pray: ?

broker/src/main/java/io/moquette/broker/SessionRegistry.java
```diff
@@ -120,18 +131,53 @@ public SessionCreationResult(Session session, CreationModeEnum mode, boolean alr
...
     }
 
+    private void checkExpiredSessions() {
+        List<ISessionsRepository.SessionData> expiredSessions = new ArrayList<>();
+        int drainedSessions = removableSessions.drainTo(expiredSessions);
+        LOG.debug("Retrieved {} expired sessions before clock {}", drainedSessions, clock.instant());
+        for (ISessionsRepository.SessionData expiredSession : expiredSessions) {
+            final String expiredAt = expiredSession.expireAt().map(Instant::toString).orElse("UNDEFINED");
+            LOG.debug("Removing session {}, expired on {}", expiredSession.clientId(), expiredAt);
+            remove(expiredSession.clientId());
```
hylkevds, line 179: Since this handles a session, shouldn't the remove/delete be handled by the session thread? 
We may now have a race condition with the session being rehydrated due to a re-connect right as it is being destroyed here.

broker/src/main/java/io/moquette/broker/SessionRegistry.java
```diff
@@ -120,18 +131,53 @@ public SessionCreationResult(Session session, CreationModeEnum mode, boolean alr
...
+            sessionsRepository.delete(expiredSession);
+        }
+    }
+
+    private void trackForRemovalOnExpiration(ISessionsRepository.SessionData session) {
```
hylkevds, line 184: I think we also need the opposite, unTrackForRemovalOnExpiration, for when a session is re-opened.

hylkevds (commented):
Looks good, just one concern about concurrency and one about re-activated sessions not being removed from the expiration queue.

broker/src/main/java/io/moquette/broker/SessionRegistry.java
andsel: Yes, good catch, `remove` method is 
```java
final Session old = pool.remove(clientID);
        if (old != null) {
            old.cleanUp();
        }
```
and `cleanUp`  
https://github.com/moquette-io/moquette/blob/e4e051f1646b379dac3a44f00db5350c18ab5789/broker/src/main/java/io/moquette/broker/Session.java#L493-L497
touches the unsafe queues system.

andsel: maybe this code could batch the removes in just one command for each queue, containing the list of sessions to purge. Maybe it's an optimization, but in case a bunch of sessions expires in the same instant, we avoid to kill the command queues.

hylkevds: That's probably a good idea. It's quite possible for a large number of sessions to expire in the same cleanup run, especially after restarting the server or if for some reason the cleanup had not happened for some time... That could overflow the queue.

andsel: To be able to send task to the session loops without passing the `PostOffice` reference down to the `SessionRegistry` I extracted a specific class (named `SessionEventLoopGroup'). This work is present in PR  #742 and should be merged before this. This PR would be rebased on top of that so that I can later proceed with the fix discussed above.

andsel: Fixed with commit 23173e0b8124d322409ef1e6238f853a1bcf6e05

broker/src/main/java/io/moquette/broker/SessionRegistry.java
andsel: Hi @hylkevds, you are right. It has been fixed with ae786d063dcbc39ce7017dd21f5c4b61c287bed4, so please could you give it another review?

hylkevds: Don't forget to implement it :)

broker/src/main/java/io/moquette/broker/SessionRegistry.java
```diff
@@ -120,18 +132,61 @@ public SessionCreationResult(Session session, CreationModeEnum mode, boolean alr
...
+        if (!session.expireAt().isPresent()) {
+            throw new RuntimeException("Can't track for expiration a session without expiry instant, client_id: " + session.clientId());
+        }
+        LOG.debug("start tracking the session {} for removal", session.clientId());
+        removableSessions.add(session);
+    }
+
+    private void untrackFromRemovalOnExpiration(ISessionsRepository.SessionData session) {
+
```
hylkevds: Needs implementation...

broker/src/main/java/io/moquette/broker/config/IConfig.java
```diff
@@ -79,4 +83,31 @@ public boolean boolProp(String propertyName, boolean defaultValue) {
         }
         return Boolean.parseBoolean(propertyValue);
     }
+
+    public Duration durationProp(String propertyName) {
+        String propertyValue = getProperty(propertyName);
+        final char timeSpecifier = propertyValue.charAt(propertyValue.length() - 1);
+        final TemporalUnit periodType;
+        switch (timeSpecifier) {
```
hylkevds, line 91: Please also add seconds and minutes. In high-volume cases we don't want sessions to stay alive for more than a minute or so...

broker/src/main/java/io/moquette/broker/SessionRegistry.java
andsel: :facepalm:

hylkevds:
While working on the subscription tree this weekend I noticed that subscriptions are not cleaned up when a session is cleaned up.
I noticed because the interceptor didn't get notified of any unsubscribe events when a (clean) session terminated due to a disconnect.

This may need its own issue, WDYT?

andsel:
@hylkevds I think this could be a bug but:

> interceptor didn't get notified of any unsubscribe events

the interceptor is notified only when an UNSUBSCRIBE message is received.  When a clean session is wiped due to a disconnection it doesn't mean that the unsubscribe interceptor is notified, or do you thing it should?

hylkevds:
I think it should, yes.
If an interceptor wants to know about subscribe and unsubscribe events it's most likely to keep track of which topics have subscribers so it know which topics to send data to. For that use-case it is important to get all unsubscribe events, regardless of why one happened.

In our specific use-case there is no simple one-to-one mapping between events and the topics that may be interested in those events. It's an (in some cases extreme) many-to-many mapping. One event may have to be sent to potentially thousands of topics. That's why we need to keep track of exactly which topics actually have subscribers, so we can avoid generating messages for topics that have no subscriptions.

Of course the subscription not being cleaned up and the interceptor not being notified are two distinct, but related, issues.

andsel:
@hylkevds please open an issue describing for that, describing the behavior. It will addressed in a separate PR.

broker/src/main/java/io/moquette/broker/SessionRegistry.java
andsel: Fixed by these 2 commits:
- 250752b98d28ca3701754f090963ed39ccc314f6 fixes test condition to verify effectively as failure when there is no implementation of the method `untrackFromRemovalOnExpiration`
- d2a5d3f235213750e23b410efd67dcf7a5926c1f implements the untracking method

broker/src/main/java/io/moquette/broker/config/IConfig.java
andsel: Fixed with e29ad318ed2e2d72bb0d709c510a100411f975aa

andsel:
Hi @hylkevds I've addressed all your latest concerns. Please could you review it?! :-)

hylkevds (dismissed):
Looks good!

PR #743, opened 2023-04-27
Avoid NPE if broker yet not started

source-c:
make listConnectedClients method safe for all broker states

Formaly, there is nothing to restrict one to call clients listing (`server.listConnectedClients()`) immediatelly after constructing instance using `new Server();`. In this case caller will get NulPointerException which may ruin the runtime.
This PR fixes that.

### DIFF
--- a/broker/src/main/java/io/moquette/broker/Server.java
+++ b/broker/src/main/java/io/moquette/broker/Server.java
@@ -637,8 +637,9 @@ public class Server {
      * Returns null if the broker is not started.
      * */
     public Collection<ClientDescriptor> listConnectedClients() {
-        if (sessions == null) {
-            return null;
+        if (!initialized) {
+            LOG.error("Moquette is not started, MQTT clients listing unavailable");
+            throw new IllegalStateException("Can't get clients list from a integration that is not yet started");
         }
         return sessions.listConnectedClients();
     }
--- a/broker/src/main/java/io/moquette/broker/Server.java
+++ b/broker/src/main/java/io/moquette/broker/Server.java
@@ -639,7 +639,7 @@ public class Server {
     public Collection<ClientDescriptor> listConnectedClients() {
         if (!initialized) {
             LOG.error("Moquette is not started, MQTT clients listing unavailable");
-            throw new IllegalStateException("Can't get clients list from a integration that is not yet started");
+            throw new IllegalStateException("Can't get clients list from a Server that is not yet started");
         }
         return sessions.listConnectedClients();
     }
--- a/broker/src/main/java/io/moquette/broker/Server.java
+++ b/broker/src/main/java/io/moquette/broker/Server.java
@@ -634,8 +634,12 @@ public class Server {
 
     /**
      * Return a list of descriptors of connected clients.
+     * Returns null if the broker is not started.
      * */
     public Collection<ClientDescriptor> listConnectedClients() {
+        if (sessions == null) {
+            return null;
+        }
         return sessions.listConnectedClients();
     }
 }


broker/src/main/java/io/moquette/broker/Server.java
```diff
@@ -634,8 +634,12 @@ public void removeInterceptHandler(InterceptHandler interceptHandler) {
 
     /**
      * Return a list of descriptors of connected clients.
+     * Returns null if the broker is not started.
      * */
     public Collection<ClientDescriptor> listConnectedClients() {
+        if (sessions == null) {
```
andsel: SessionReposistory is assigned during `startServer` method, https://github.com/moquette-io/moquette/blob/f306b6c1b280d8a95b8ce2ce304af9d2dbea60cd/broker/src/main/java/io/moquette/broker/Server.java#L242

When the `startServer` completes, and the server is ready to operate, is sets the `initialized` flag to true 
https://github.com/moquette-io/moquette/blob/f306b6c1b280d8a95b8ce2ce304af9d2dbea60cd/broker/src/main/java/io/moquette/broker/Server.java#L262


The `listConnectedClients`, and maybe others that can be called only after the Server status is initialized, should check for this status, for example:

```java
private void assertServerInitliazed() {
    if (!initialized) {
        throw new IllegalStateException("Can't invoke status method before server is initialized, please call startServer before");
    }
}

public Collection<ClientDescriptor> listConnectedClients() {
   assertServerInitliazed();
   return sessions.listConnectedClients();
}   
```

In this case if it reaches a NPE it's a really bad situation, because after the server transitioned in initialized status, the session's list (which should have a value) isn't valued.

Checking null to return null is not a good way to resolve it.

WDYT?

andsel (changes requested):
Left a suggestion on haw to improve it.
null checking to return a null value is just like moving the problem on the caller, which in worst case would expect an empty collection.

broker/src/main/java/io/moquette/broker/Server.java
source-c: We are working with moquette from Clojure, where `nil` is a possible value of any data type in contrary to pure Java. Beside, the Clojure conditional system is based around `nil` and `false`, where both are representing the values of logical falsity in conditional tests (i.e. anything else is logical truth). 

```
user=> (if nil true false)
false
user=> (if [] true false)
true
```

So, from my perspective having `null` here is more desired exactly for case when the `Server` instance is already created but was not started yet. Moreover, it is a good enough approach, because for service driven architectures, constructing the system is not always the same as bringing it to ready state (i.e. to start serving clients). Sometimes it is highly desired to reuse existing service but to add new interceptors or to modify existing, that in turn highly likely being driven by some logic. Having `null` here clearly indicates the state `not running` when empty list semantically indicates `no clients`. 
Throwing exception is even worse than everything else here. When one may get an exception from the __listing__ method in runtime, strictly speaking - with no any reason, it is strange.

andsel: I understand your point view, where in Clojure, an exception breaks the purity of the method, while a `null` expresses a result.
From an OO point of view, if a method doesn't satisfy the preconditions, it's better to use the exception, with the type clearly expressing what wasn't good.

I would avoid as much as we can null checks, and so also avoid to return null values.

If a method has to return a value which means "no value" it's better to create an Enum or return type class (which complicates the class hierarchy).

source-c: Up to you - You are the author and maintainer :) Let it be exception. We will wrap that in our code.. It is anyway way better than NPE.

andsel: I'll create a PR to solve this in next days

source-c:
I've make changes in code in accordance to your suggestions in the code review.

broker/src/main/java/io/moquette/broker/Server.java
source-c: Please, see the next change provided.

broker/src/main/java/io/moquette/broker/Server.java
```diff
@@ -634,8 +634,13 @@ public void removeInterceptHandler(InterceptHandler interceptHandler) {
 
     /**
      * Return a list of descriptors of connected clients.
+     * Returns null if the broker is not started.
      * */
     public Collection<ClientDescriptor> listConnectedClients() {
+        if (!initialized) {
+            LOG.error("Moquette is not started, MQTT clients listing unavailable");
+            throw new IllegalStateException("Can't get clients list from a integration that is not yet started");
```
```suggestion
            throw new IllegalStateException("Can't get clients list from a Server that is not yet started");
```

andsel (changes requested):
Nitpick for the message of the exception. Moquette users doesn't know what an integration is

source-c:
Entirely agree. Commited.

PR #744, opened 2023-04-29
Feature disconnects

### LINKED ISSUE #469: How to forcefully disconnect the client from the broker?
### Expected behavior

In Version 0.11, we could use the below code snippet to disconnect the client forcefully from the broker.

public boolean disconnectClient(String endpoint) {
		ConnectionDescriptor connectionDescriptor = mqttBroker.getConnectionsManager().getConnection(endpoint);
		mqttBroker.getConnectionsManager().getConnection(endpoint).abort();
		boolean status = mqttBroker.getConnectionsManager().removeConnection(connectionDescriptor);
		return status;
	}

Server.java exposed the API to get the ConnectionsManager through which we were able to abort the connection.

public IConnectionsManager getConnectionsManager() {
        return m_processorBootstrapper.getConnectionDescriptors();
    }

### Actual behavior

As of now in the version 0.12.1, we are unable to find the similar API to abort the client connection. Could you please help us with the information as to how we can disconnect the client forcefully using any other API that's available.

source-c:
- covers and fixes #469
- does clenup
- integrated into Server implementation
- also contains #743

Also fixed an issue of race that caused exception on connect, that is very easy to reproduce:

```
import paho.mqtt.client as mqtt
mqttBroker = "127.0.0.1"

client = mqtt.Client(client_id=b"Temperature_Inside")
client.connect(mqttBroker)
res = client.publish("/TEMPERATURE", randNumber)
```

that causes:

```
20:33:56.213 [nioEventLoopGroup-13-1] DEBUG io.moquette.broker.MQTTConnection -- Received MQTT message, type: CONNECT
20:33:56.214 [nioEventLoopGroup-13-1] DEBUG io.moquette.broker.PostOffice -- Routing cmd [CONN] for session [Temperature_Inside] to event processor 7
20:33:56.215 [nioEventLoopGroup-13-1] DEBUG io.moquette.broker.metrics.MQTTMessageLogger -- C->B PUBLISH <null> to topics </TEMPERATURE>
20:33:56.215 [nioEventLoopGroup-13-1] DEBUG io.moquette.broker.MQTTConnection -- Received MQTT message, type: PUBLISH
20:33:56.216 [nioEventLoopGroup-13-1] ERROR io.moquette.broker.NewNettyMQTTHandler -- Error processing protocol message: PUBLISH
java.lang.NullPointerException: Cannot invoke "String.hashCode()" because the return value of "io.moquette.broker.SessionCommand.getSessionId()" is null
        at io.moquette.broker.PostOffice.routeCommand(PostOffice.java:632)
        at io.moquette.broker.MQTTConnection.processPublish(MQTTConnection.java:427)
        at io.moquette.broker.MQTTConnection.handleMessage(MQTTConnection.java:80)
        at io.moquette.broker.NewNettyMQTTHandler.channelRead(NewNettyMQTTHandler.java:58)
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:379)
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:365)
....
```

The disconnection feature works like this (traces):

```
======> MQTT Clients: ({:id Temperature_Inside, :address 127.0.0.1, :port 54641})
======> MQTT Advanced Broker disconnecting .........
20:05:45.234 [main] DEBUG com.dkdhub.mqtt_broker.AdvancedBroker -- Will disconnect client Temperature_Inside
20:05:45.234 [main] DEBUG io.moquette.broker.SessionRegistry -- Disconnecting client: Temperature_Inside
20:05:45.235 [main] DEBUG io.moquette.broker.SessionRegistry -- Remove session state for client Temperature_Inside
20:05:45.235 [main] DEBUG io.moquette.broker.SessionRegistry -- Client Temperature_Inside successfully disconnected from broker
20:05:45.235 [main] DEBUG com.dkdhub.mqtt_broker.AdvancedBroker -- Disconnect result: true
======> MQTT Clients: ()
20:05:45.236 [nioEventLoopGroup-3-1] INFO io.moquette.broker.metrics.MQTTMessageLogger -- Channel Inactive
20:05:45.236 [nioEventLoopGroup-3-1] DEBUG io.moquette.broker.MQTTConnection -- Notifying connection lost event
20:05:45.237 [nioEventLoopGroup-3-1] DEBUG io.moquette.broker.SessionEventLoopGroup -- Routing cmd [CONN LOST] for session [Temperature_Inside] to event processor 7
20:05:45.237 [Session Executor 7] DEBUG io.moquette.broker.MQTTConnection -- Cleaning Temperature_Inside
20:05:45.237 [Session Executor 7] DEBUG io.moquette.interception.BrokerInterceptor -- Notifying unexpected MQTT client disconnection to interceptor CId=Temperature_Inside, username=null, interceptorId=3456
20:05:45.238 [pool-2-thread-1] DEBUG basic -- MQTT: client  #object[clojure.core$_PLUS_ 0x46f9cda1 clojure.core$_PLUS_@46f9cda1] Temperature_Inside #object[clojure.core$_PLUS_ 0x46f9cda1 clojure.core$_PLUS_@46f9cda1]  disconnected (lost)
```

We maintain a clojure library that wraps moquette, so the tests are written in clojure (actually this functionality is at trunk now and awaits for upstream updates), however they are way simple, so you may take a look by yourself - https://github.com/dkdhub/clj-mqtt-broker/blob/3d59a2244afec43ae2a3522ac9173ae683a9dc3d/test/basic.clj#L129 

Looking forward to hearing from you!

### DIFF
--- a/broker/src/main/java/io/moquette/broker/Server.java
+++ b/broker/src/main/java/io/moquette/broker/Server.java
@@ -637,8 +637,9 @@ public class Server {
      * Returns null if the broker is not started.
      * */
     public Collection<ClientDescriptor> listConnectedClients() {
-        if (sessions == null) {
-            return null;
+        if (!initialized) {
+            LOG.error("Moquette is not started, MQTT clients listing unavailable");
+            throw new IllegalStateException("Can't get clients list from a integration that is not yet started");
         }
         return sessions.listConnectedClients();
     }
--- a/broker/src/main/java/io/moquette/broker/Server.java
+++ b/broker/src/main/java/io/moquette/broker/Server.java
@@ -643,7 +643,7 @@ public class Server {
         return sessions.listConnectedClients();
     }
 
-    public boolean disconnectClient(final String clientID, boolean cleanSession) {
-        return sessions.dropConnection(clientID, cleanSession);
+    public boolean disconnectClient(final String clientID, boolean clearSession) {
+        return sessions.dropConnection(clientID, clearSession);
     }
 }
--- a/broker/src/main/java/io/moquette/broker/Server.java
+++ b/broker/src/main/java/io/moquette/broker/Server.java
@@ -643,7 +643,7 @@ public class Server {
         return sessions.listConnectedClients();
     }
 
-    public boolean disconnectClient(final String clientID) {
-        return sessions.dropConnection(clientID);
+    public boolean disconnectClient(final String clientID, boolean cleanSession) {
+        return sessions.dropConnection(clientID, cleanSession);
     }
 }
--- a/broker/src/main/java/io/moquette/broker/Server.java
+++ b/broker/src/main/java/io/moquette/broker/Server.java
@@ -642,8 +642,20 @@ public class Server {
         }
         return sessions.listConnectedClients();
     }
-
-    public boolean disconnectClient(final String clientID, boolean clearSession) {
-        return sessions.dropConnection(clientID, clearSession);
+    /**
+     * Force the disconnection of a client, closing the related session.
+     * @param clientId the name of the client to drop session.
+     */
+    public boolean disconnectClient(final String clientId) {
+        return sessions.dropSession(clientId);
+    }
+    
+    /**
+     * Force the disconnection of a client, closing the related session and removing any session state from
+     * the broker, such as subscriptions and queue.
+     * @param clientId the name of the client to drop session.
+     */
+    public boolean disconnectAndPurgeClientState(final String clientId) {
+        return sessions.dropAndPurgeSession(clientId);
     }
 }
--- a/broker/src/main/java/io/moquette/broker/Server.java
+++ b/broker/src/main/java/io/moquette/broker/Server.java
@@ -642,4 +642,8 @@ public class Server {
         }
         return sessions.listConnectedClients();
     }
+
+    public boolean disconnectClient(final String clientID) {
+        return sessions.dropConnection(clientID);
+    }
 }
--- a/broker/src/main/java/io/moquette/broker/Server.java
+++ b/broker/src/main/java/io/moquette/broker/Server.java
@@ -639,7 +639,7 @@ public class Server {
     public Collection<ClientDescriptor> listConnectedClients() {
         if (!initialized) {
             LOG.error("Moquette is not started, MQTT clients listing unavailable");
-            throw new IllegalStateException("Can't get clients list from a integration that is not yet started");
+            throw new IllegalStateException("Can't get clients list from a Server that is not yet started");
         }
         return sessions.listConnectedClients();
     }
--- a/broker/src/main/java/io/moquette/broker/Server.java
+++ b/broker/src/main/java/io/moquette/broker/Server.java
@@ -634,12 +634,8 @@ public class Server {
 
     /**
      * Return a list of descriptors of connected clients.
-     * Returns null if the broker is not started.
      * */
     public Collection<ClientDescriptor> listConnectedClients() {
-        if (sessions == null) {
-            return null;
-        }
         return sessions.listConnectedClients();
     }
 
--- a/broker/src/main/java/io/moquette/broker/Server.java
+++ b/broker/src/main/java/io/moquette/broker/Server.java
@@ -647,15 +647,15 @@ public class Server {
      * @param clientId the name of the client to drop session.
      */
     public boolean disconnectClient(final String clientId) {
-        return sessions.dropSession(clientId);
+        return sessions.dropSession(clientId, false);
     }
-    
+
     /**
      * Force the disconnection of a client, closing the related session and removing any session state from
      * the broker, such as subscriptions and queue.
      * @param clientId the name of the client to drop session.
      */
     public boolean disconnectAndPurgeClientState(final String clientId) {
-        return sessions.dropAndPurgeSession(clientId);
+        return sessions.dropSession(clientId, true);
     }
 }
--- a/broker/src/main/java/io/moquette/broker/Server.java
+++ b/broker/src/main/java/io/moquette/broker/Server.java
@@ -634,8 +634,12 @@ public class Server {
 
     /**
      * Return a list of descriptors of connected clients.
+     * Returns null if the broker is not started.
      * */
     public Collection<ClientDescriptor> listConnectedClients() {
+        if (sessions == null) {
+            return null;
+        }
         return sessions.listConnectedClients();
     }
 }
--- a/broker/src/main/java/io/moquette/broker/Server.java
+++ b/broker/src/main/java/io/moquette/broker/Server.java
@@ -636,6 +636,10 @@ public class Server {
      * Return a list of descriptors of connected clients.
      * */
     public Collection<ClientDescriptor> listConnectedClients() {
+        if (!initialized) {
+            LOG.error("Moquette is not started, MQTT clients listing unavailable");
+            throw new IllegalStateException("Can't get clients list from a Server that is not yet started");
+        }
         return sessions.listConnectedClients();
     }
 
--- a/broker/src/main/java/io/moquette/broker/SessionRegistry.java
+++ b/broker/src/main/java/io/moquette/broker/SessionRegistry.java
@@ -46,13 +46,15 @@ public class SessionRegistry {
          * Releases any held resources. Must be called when the EnqueuedMessage is no
          * longer needed.
          */
-        public void release() {}
+        public void release() {
+        }
 
         /**
          * Retains any held resources. Must be called when the EnqueuedMessage is added
          * to a store.
          */
-        public void retain() {}
+        public void retain() {
+        }
     }
 
     public static class PublishedMessage extends EnqueuedMessage {
@@ -210,7 +212,7 @@ public class SessionRegistry {
         //verify if subscription still satisfy read ACL permissions
         for (Subscription existingSub : session.getSubscriptions()) {
             final boolean topicReadable = authorizator.canRead(existingSub.getTopicFilter(), username,
-                                                               session.getClientID());
+                session.getClientID());
             if (!topicReadable) {
                 subscriptionsDirectory.removeSubscription(existingSub.getTopicFilter(), session.getClientID());
             }
@@ -311,6 +313,27 @@ public class SessionRegistry {
             .collect(Collectors.toList());
     }
 
+    boolean dropConnection(final String clientID) {
+        LOG.debug("Disconnecting client: {}", clientID);
+        if (clientID == null) return false;
+
+        final Session client = pool.get(clientID);
+
+        if (client != null) {
+
+            client.closeImmediately();
+            client.cleanUp();
+            purgeSessionState(client);
+            pool.remove(clientID);
+
+            LOG.debug("Client {} successfully disconnected from broker", clientID);
+            return true;
+        }
+
+        LOG.debug("Client {} not found, nothing disconnected", clientID);
+        return false;
+    }
+
     private Optional<ClientDescriptor> createClientDescriptor(Session s) {
         final String clientID = s.getClientID();
         final Optional<InetSocketAddress> remoteAddressOpt = s.remoteAddress();
@@ -319,7 +342,7 @@ public class SessionRegistry {
 
     /**
      * Close all resources related to session management
-     * */
+     */
     public void close() {
         queueRepository.close();
     }
--- a/broker/src/main/java/io/moquette/broker/SessionRegistry.java
+++ b/broker/src/main/java/io/moquette/broker/SessionRegistry.java
@@ -313,7 +313,7 @@ public class SessionRegistry {
             .collect(Collectors.toList());
     }
 
-    boolean dropConnection(final String clientId) {
+    boolean dropConnection(final String clientId, boolean cleanSession) {
         LOG.debug("Disconnecting client: {}", clientId);
         if (clientId == null) {
             return false;
@@ -325,8 +325,10 @@ public class SessionRegistry {
             return false;
         }
 
-       client.closeImmediately();
-       purgeSessionState(client);
+        client.closeImmediately();
+        if (cleanSession) {
+            purgeSessionState(client);
+        }
 
        LOG.debug("Client {} successfully disconnected from broker", clientId);
        return true;
--- a/broker/src/main/java/io/moquette/broker/SessionRegistry.java
+++ b/broker/src/main/java/io/moquette/broker/SessionRegistry.java
@@ -313,7 +313,14 @@ public class SessionRegistry {
             .collect(Collectors.toList());
     }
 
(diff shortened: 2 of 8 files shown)


source-c:
@andsel may I ask you to take a look at this please? Or to assign this request to someone who may address this?

Small note in regards to CONNECT+PUBLISH issue: the root cause is that the implementation mainly relies on NettyUtils.clientID(channel) call, that requires channel to be associated with the connection. At the moment of CONNECT the new association creation is implemented by firing out the `FutureTask` instance and since immediate publishing is not restricted by MQTT spec - the message can not be processed and should be ignored. It is enough to return `PostOffice.RouteResult.failed` for the case. That oneline fix is embedded into this request.

Thank you.

andsel:
Hi @source-c 
thanks for your contribution.  To ease the review, discussion an tracking of the issues I would ask a couple things.
Please:
- split the NPE PR, it's already covered by #743 so document well how to reproduce the problem in that PR
- remove the `0.17-SNAPSHOT-1` changes

broker/src/main/java/io/moquette/broker/SessionRegistry.java
```diff
@@ -46,13 +46,15 @@ public abstract static class EnqueuedMessage {
...
          */
-        public void release() {}
+        public void release() {
+        }
 
         /**
          * Retains any held resources. Must be called when the EnqueuedMessage is added
          * to a store.
          */
-        public void retain() {}
```
andsel, line 55: Avoid reformatting, keep the change to bare minimal, please

distribution/pom.xml
```diff
@@ -5,7 +5,7 @@
         <relativePath>../pom.xml</relativePath>
         <artifactId>moquette-parent</artifactId>
         <groupId>io.moquette</groupId>
-        <version>0.17-SNAPSHOT</version>
+        <version>0.17-SNAPSHOT-1</version>
```
andsel: Avoid the to change the version.

source-c:
Thank you, @andsel. Both PRs were modified in accordance to your comments.

andsel:
Thanks @source-c  I'll review in the next days, thank you for contributing to the project. I'll comment on them, if I have any doubts.

andsel:
@source-c #743 has been merged on `main` so you can rebase this, please

source-c:
Hi there! @andsel, I see ubuntu-18.04 have were all failed with no run, twice. Is there a chance for this commit to be merged to the `main` without that?

andsel:
Hi @source-c yes it's possible, but I have to review it and had no time in the past weeks :-( I'll try to do it during next weekend

source-c:
> Hi @source-c yes it's possible, but I have to review it and had no time in the past weeks :-( I'll try to do it during next weekend

Oh, fine then. Thank you!

andsel:
Hi @source-c 
I tried to understand the original statement problem, the one that report an NPE on command routing.
From the [comment](https://github.com/moquette-io/moquette/pull/744#issuecomment-1530227154) is not clear why the `publish` should be processed before the `connect` terminates his association.
In Moquette all MQTT messages are wrapped as commands and executed by an event loop.
If in one TCP connection a client send a `CONNECT` and a `PUBLISH` the two are serialized, so the connect correctly associates the `clientId` to the Netty's channel.

I tried to reproduce with following (run it with jbang), but haven't luck to reproduce your issue.
```java
///usr/bin/env jbang "$0" "$@" ; exit $?
//DEPS org.eclipse.paho:org.eclipse.paho.client.mqttv3:1.2.5
package io.moquette.testembedded;

import org.eclipse.paho.client.mqttv3.MqttClient;
import org.eclipse.paho.client.mqttv3.MqttClientPersistence;
import org.eclipse.paho.client.mqttv3.MqttException;
import org.eclipse.paho.client.mqttv3.persist.MemoryPersistence;

public class TestClient {
    public static void main(String[] args) throws MqttException {
        MqttClientPersistence dataStore = new MemoryPersistence();
        MqttClient client = new MqttClient("tcp://localhost:1883", "TestClient", dataStore);

        client.connect();
        client.publish("/temperature", new byte[] {0x14}, 0, false);
        client.disconnect();
        client.close();
    }
}
```

Please, could you expand your explanation of the problem?

source-c:
Seems, you cannot reproduce that because of implementation differences:

org.eclipse.paho.client.mqttv3.MqttClient:
```
aClient.connect(options, null, null).waitForCompletion(getTimeToWait());
```
when paho.mqtt.client does just:
```
self.connect_async(host, port, keepalive, bind_address, bind_port, clean_start, properties)
```

you see there is no await at pythonic client (which is used very broadly).
So, when one uses such approach:
```
client = mqtt.Client(client_id=cid)
client.connect(mqttBroker)
res = client.publish(topic, value)
```
both `CONNECT` and `PUBLISH` may (and will) go almost simultaniously.
On the other hand we have association creation at broker, that yields `FutureTask` called by `CONNECT` that came.
For the next packet containing `PUBLISH`, broker searches for client id (at `processPublish`) using `getClientId()` which is in turn just a wrapper around the code:
```
NettyUtils.clientID(channel);
```
and this channel in turn is yet not completely initialized.
As the result we are getting an NPE there.

andsel:
I've tried also with Java Paho Async client and HiveMQ async, but neither of the two exposes such behavior. 
Reading the [spec fro MQTT 3.1.1](http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html) it doesn't prohibit such to send other control packets before having received the `CONN_ACK`, requirement `[MQTT-3.1.4-5]`

```
Clients are allowed to send further Control Packets immediately after sending a CONNECT Packet; 
Clients need not wait for a CONNACK Packet to arrive from the Server. 
If the Server rejects the CONNECT, it MUST NOT process any data sent by the Client after the 
CONNECT Packet [MQTT-3.1.4-5].
```

although it describes this **non normative comment**:
```
Clients typically wait for a CONNACK Packet, However, if the Client exploits its freedom to send Control Packets before it receives a CONNACK, it might simplify the Client implementation as it does not have to police the connected state. The Client accepts that any data that it sends before it receives a CONNACK packet from the Server will not be processed if the Server rejects the connection.
```

So it seems that if the broker receives the PUB before sending the CONN_ACK, and the connection is accepted, then it has to process the PUB as if it's a valid.

I think that to implement such behavior we should move the assignment of `clientId` to channel, which happens in the session loop while processing the connect https://github.com/moquette-io/moquette/blob/ec4e84a666ab4b8cd4bf9431da3ea6c3b5fe6fc1/broker/src/main/java/io/moquette/broker/MQTTConnection.java#L217 , to he point before enqueuing the command: https://github.com/moquette-io/moquette/blob/ec4e84a666ab4b8cd4bf9431da3ea6c3b5fe6fc1/broker/src/main/java/io/moquette/broker/MQTTConnection.java#L183


In this case:
- if the `CONNECT` result with a positive result, then the `PUBLISH` is routed to the correct session loop and given the serialization of the commands, when the session loop start to process the `PUBLISH` all the session data is setup and both the session and the connection have valid data.
- if  the `CONNECT` result with a negative result, and closes the channel, when the session loop encounter any command (PUB, SUB etc) related to that channel/clientId, it doesn't have a channel/session created and *should* log an error or warning.

broker/src/main/java/io/moquette/broker/SessionRegistry.java
```diff
@@ -311,6 +313,27 @@ Collection<ClientDescriptor> listConnectedClients() {
...
+            pool.remove(clientID);
+
+            LOG.debug("Client {} successfully disconnected from broker", clientID);
+            return true;
+        }
+
+        LOG.debug("Client {} not found, nothing disconnected", clientID);
+        return false;
+    }
+
```
andsel:
```suggestion
    boolean dropConnection(final String clientId) {
        LOG.debug("Disconnecting client: {}", clientId);
        if (clientId == null) {
            return false;
        }

        final Session client = pool.get(clientId);
        if (client == null) {
            LOG.debug("Client {} not found, nothing disconnected", clientId);
            return false;
        }

       client.closeImmediately();
       purgeSessionState(client);

       LOG.debug("Client {} successfully disconnected from broker", clientId);
       return true;
    }

```

andsel (changes requested):
Left a request to change the layout of the code, plus I would ask to cover this with a unit test that verifies the behaviors:
- dropping a connection for a non existing session.
- dropping a connection for an existing session.


However it's not clear to me why  SessionRegistry.closeSession is not enough for your use case.

BTW in the suggestion i removed
```java
client.cleanUp();
pool.remove(clientId);
```

because are implicit in the call to `purgeSessionState`

distribution/pom.xml
source-c: Yeah, this change has been unrolled already.

broker/src/main/java/io/moquette/broker/SessionRegistry.java
source-c: Agree.

source-c:
> I think that to implement such behavior we should move the assignment of clientId to channel, which happens in the session loop while processing the connect
> 
> [moquette/broker/src/main/java/io/moquette/broker/MQTTConnection.java](https://github.com/moquette-io/moquette/blob/ec4e84a666ab4b8cd4bf9431da3ea6c3b5fe6fc1/broker/src/main/java/io/moquette/broker/MQTTConnection.java#L217)
> 
> Line 217 in [ec4e84a](https://github.com/moquette-io/moquette/commit/ec4e84a666ab4b8cd4bf9431da3ea6c3b5fe6fc1)
>  NettyUtils.clientID(channel, clientId); 
> , to he point before enqueuing the command:
> 
> [moquette/broker/src/main/java/io/moquette/broker/MQTTConnection.java](https://github.com/moquette-io/moquette/blob/ec4e84a666ab4b8cd4bf9431da3ea6c3b5fe6fc1/broker/src/main/java/io/moquette/broker/MQTTConnection.java#L183)
> 
> Line 183 in [ec4e84a](https://github.com/moquette-io/moquette/commit/ec4e84a666ab4b8cd4bf9431da3ea6c3b5fe6fc1)
>  final String sessionId = clientId; 
> 
> In this case:
> 
> - if the CONNECT result with a positive result, then the PUBLISH is routed to the correct session loop and given the serialization of the commands, when the session loop start to process the PUBLISH all the session data is setup and both the session and the connection have valid data.
> - if the CONNECT result with a negative result, and closes the channel, when the session loop encounter any command (PUB, SUB etc) related to that channel/clientId, it doesn't have a channel/session created and should log an error or warning.

Of course, implementing that in such a way follows the spec more precisely. But it doesn't neglect that empty clientId case should be either be covered at the routing. At least to identify a case of collision/race if any. Also, I suppose changing that behaviour is to be done separately to avoid mess, does it?

source-c:
> However it's not clear to me why SessionRegistry.closeSession is not enough for your use case.

There are a lot use cases when forcing disconnect is used under hard pressure and in accordance to security policies. Some of use cases already discussed at #747 BTW.

andsel:
> changing that behaviour is to be done separately to avoid mess, does it?

Yes, it's better separate in another PR

andsel:
> There are a lot use cases when forcing disconnect 

I understand, but it's not clear to me if the intended effect of the new method should be like a standard close; which shutdown the MQTTConnection and clean the Session status (means topic subscriptions and queues)  if needed (clean session = false).
Or does it has to remove all session status also if the session was a non clean session?
In the second case every reference to the session is brutally removed, while in the first is like a normal disconnection flow.

source-c:
>  In the second case every reference to the session is brutally removed, while in the first is like a normal disconnection flow.

Unfortunatelly, second case is most prefered. We should remove all the session internals disregading its state. The state might be not clean, the channel might not be operational, server (broker) might be overwhelmed by unwanted (sometimes broken) requests. Forcing session drop (i.e. kicking off the client) we perserve server stability among other reasons, trying to keep it operational for all other clients.

source-c:
However you are right - this should be decided by end user, so I remade cleanup contract and now session cleanup is to be specified by outer parameter explicitly.

broker/src/main/java/io/moquette/broker/Server.java
```diff
@@ -643,4 +642,8 @@ public Collection<ClientDescriptor> listConnectedClients() {
         }
         return sessions.listConnectedClients();
     }
+
+    public boolean disconnectClient(final String clientID, boolean clearSession) {
+        return sessions.dropConnection(clientID, clearSession);
+    }
```
```suggestion
    /**
     * Force the disconnection of a client, closing the related session.
     * @param clientId the name of the client to drop session.
     */
    public boolean disconnectClient(final String clientId) {
        return sessions.dropSession(clientId);
    }
    
    /**
     * Force the disconnection of a client, closing the related session and removing any session state from
     * the broker, such as subscriptions and queue.
     * @param clientId the name of the client to drop session.
     */
    public boolean disconnectAndPurgeClientState(final String clientId) {
        return sessions.dropAndPurgeSession(clientId);
    }
```

broker/src/main/java/io/moquette/broker/SessionRegistry.java
```diff
@@ -311,6 +313,27 @@ Collection<ClientDescriptor> listConnectedClients() {
...
+        client.closeImmediately();
+        if (clearSession) {
+            purgeSessionState(client);
+        }
+
+       LOG.debug("Client {} successfully disconnected from broker", clientId);
+       return true;
+    }
+
```
andsel:
```suggestion
   /**
    * Close the connection bound to the session for the clintId. If removeSessionState is provided
    * remove any session state like queues and subscription from broker memory.
    *
    * @param clientId the name of the client to drop the session.
    * @param removeSessionState boolean flag to request the removal of session state from broker.
    */ 
    boolean dropAndPurge(final String clientId, boolean removeSessionState) {
        LOG.debug("Disconnecting client: {}", clientId);
        if (clientId == null) {
            return false;
        }

        final Session client = pool.get(clientId);
        if (client == null) {
            LOG.debug("Client {} not found, nothing disconnected", clientId);
            return false;
        }

        client.closeImmediately();
        if (removeSessionState) {
            purgeSessionState(client);
        }

       LOG.debug("Client {} successfully disconnected from broker", clientId);
       return true;
    }

```

andsel (changes requested):
HI @source-c , I'm not a big fan of boolean flag parameter to customize the behavior of a method, I prefer to create 2 different methods that better express the intention.
On public API I avoid the feature flag, on internal if it generates more work and code duplication I could accept.

We are almost ready to go, please write a unit test to cover introduced functionality or let me know if you want I wrote it :-) adding some commits to this PR.

source-c:
> We are almost ready to go, please write a unit test to cover introduced functionality or let me know if you want I wrote it :-) adding some commits to this PR.

May I ask you to do it yourself please? It seems like we have quite different styling views, so I have no doubts your implementations will be much more clear than mine. :)
Thank you!

PR #748, opened 2023-05-01
Attempt at a method to kick an active client.

kazetsukaimiko:
Counterpart to #747 , this is an attempt at exposing a method to kick an active client. This is for use with clients that are exhibiting unacceptable behavior such as spamming.

If there are other ideas for how to address this, please let me know.


kazetsukaimiko:
Ah, it looks like #744 takes a shot at this as well. That may be a better implementation for the logging / etc.

PR #758, opened 2023-06-21
Fixed #633 adding subscriptions

hylkevds:
Converted subscription from HashSet to ArrayList
   
Copying a Set is much slower than copying an ArrayList. Since the slow part in the code is the copy step, using an ArrayList is much faster overall. Since the code using the subscriptions only iterates over them, the change from Set to List makes no real difference.

Fixes #633


andsel (dismissed):
LGTM

Before merging please add a line to `Changelog` file

hylkevds:
> LGTM
> Before merging please add a line to `Changelog` file

I always forget to update my Changelogs :)
I've updated it, PR needs a new approve now due to that change.

andsel (approved):

PR #764, opened 2023-07-31
Uses the expiry interval from connect properties

andsel:
- Retrieve the session expiry interval from Connect property SESSION_EXPIRY_INTERVAL
- Added unit test to verify that CONNECT's SESSION_EXPIRY_INTERVAL property is used to set session expiration timeout in seconds
- Fallback to global expiry if one is not provided (if session is not clean)

### DIFF
--- a/broker/src/main/java/io/moquette/broker/SessionRegistry.java
+++ b/broker/src/main/java/io/moquette/broker/SessionRegistry.java
@@ -22,6 +22,7 @@ import io.moquette.broker.subscriptions.Topic;
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.Unpooled;
 import io.netty.handler.codec.mqtt.MqttConnectMessage;
+import io.netty.handler.codec.mqtt.MqttProperties;
 import io.netty.handler.codec.mqtt.MqttQoS;
 import io.netty.handler.codec.mqtt.MqttVersion;
 import org.slf4j.Logger;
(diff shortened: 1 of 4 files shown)


broker/src/main/java/io/moquette/broker/SessionRegistry.java
```diff
@@ -298,9 +299,22 @@ private Session createNewSession(MqttConnectMessage msg, String clientId) {
...
         final MqttVersion mqttVersion = Utils.versionFromConnect(msg);
+        if (mqttVersion != MqttVersion.MQTT_5) {
+            // in MQTT3 cleanSession = true means expiryInterval=0 else infinite
+            expiryInterval = clean ? 0 : globalExpirySeconds;
+        } else {
+            final MqttProperties.MqttProperty<Integer> expiryIntervalProperty =
+                (MqttProperties.MqttProperty<Integer>) msg.variableHeader().properties()
+                    .getProperty(MqttProperties.MqttPropertyType.SESSION_EXPIRY_INTERVAL.value());
+            if (expiryIntervalProperty != null) {
+                expiryInterval = expiryIntervalProperty.value();
```
hylkevds: For server stability it is probably a good idea to allow server operators to set a maximum allowed SESSION_EXPIRY_INTERVAL. For public, high-volume servers it is simply not feasible to allow random clients to create (effectively) non-expiring sessions, and we all know clients can't be trusted to properly set their intervals :)

andsel: Good point, some cap value configurable by settings :+1

andsel: Solved with 945ac33385fc7aa5889161d80dace1b2ee9a14e4

hylkevds (approved):
Looks good.

PR #774, opened 2023-08-28
Fix: Typo in log with an invalid message

j-madden:


MikeDombo (approved):

broker/src/main/java/io/moquette/broker/NettyUtils.java
```diff
@@ -83,7 +83,7 @@ public static String userName(Channel channel) {
 	public static MqttMessage validateMessage(Object message) throws IOException, ClassCastException {
 		MqttMessage msg = (MqttMessage) message;
 		if (msg.decoderResult() != null && msg.decoderResult().isFailure()) {
-			throw new IOException("invalid massage", msg.decoderResult().cause());
+			throw new IOException("invalid message", msg.decoderResult().cause());
```
andsel:
```suggestion
			throw new IOException("Invalid message", msg.decoderResult().cause());
```

andsel (dismissed):

PR #775, opened 2023-09-18
feat: peer cert as username

MikeDombo:
Rebase of https://github.com/moquette-io/moquette/pull/675

This PR adds enables authenticating clients based on their peer certificates without changing the existing authentication interface. This functionality is enabled by the addition of a new configuration option which passes the client certificate PEM in place of `username`. The EMQx broker also has the same configuration option.

There is also a very small change to behavior with this PR. Previously, clients that did not provide a password were immediately rejected if anonymous sessions were disabled. As far as I can tell, the spec allows username without a password. The code will now allow this and pass the username and empty password to the authenticator.

Why is this needed?
The existing auth interfaces are insufficient when clients need to be identified based on their client certificate. The current interfaces only support username, password, and client ID without any ability for authenticators and authorizers to use the client certificate (if provided) for authN/authZ.

By passing in the client certificate as username, this allows the authorizer to uniquely identify clients and authenticate and authorize based on this cryptographically secure identity.

Close #675

### DIFF
--- a/broker/src/main/java/io/moquette/broker/MQTTConnection.java
+++ b/broker/src/main/java/io/moquette/broker/MQTTConnection.java
@@ -16,8 +16,9 @@
 package io.moquette.broker;
 
 import io.moquette.BrokerConstants;
-import io.moquette.broker.subscriptions.Topic;
 import io.moquette.broker.security.IAuthenticator;
+import io.moquette.broker.security.PemUtils;
+import io.moquette.broker.subscriptions.Topic;
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.ByteBufHolder;
 import io.netty.channel.Channel;
@@ -44,16 +45,21 @@ import io.netty.handler.codec.mqtt.MqttSubscribeMessage;
 import io.netty.handler.codec.mqtt.MqttUnsubAckMessage;
 import io.netty.handler.codec.mqtt.MqttUnsubscribeMessage;
 import io.netty.handler.codec.mqtt.MqttVersion;
+import io.netty.handler.ssl.SslHandler;
 import io.netty.handler.timeout.IdleStateHandler;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.io.IOException;
 import java.net.InetSocketAddress;
+import java.security.cert.Certificate;
+import java.security.cert.CertificateEncodingException;
 import java.util.List;
 import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
+import javax.net.ssl.SSLPeerUnverifiedException;
 
 import static io.moquette.BrokerConstants.INFLIGHT_WINDOW_SIZE;
 import static io.netty.channel.ChannelFutureListener.CLOSE_ON_FAILURE;
@@ -396,25 +402,46 @@ final class MQTTConnection {
     }
 
     private boolean login(MqttConnectMessage msg, final String clientId) {
-        // handle user authentication
+        String userName = null;
+        byte[] pwd = null;
+
         if (msg.variableHeader().hasUserName()) {
-            byte[] pwd = null;
+            userName = msg.payload().userName();
+            // MQTT 3.1.2.9 does not mandate that there is a password - let the authenticator determine if it's needed
             if (msg.variableHeader().hasPassword()) {
                 pwd = msg.payload().passwordInBytes();
-            } else if (!brokerConfig.isAllowAnonymous()) {
-                LOG.info("Client didn't supply any password and MQTT anonymous mode is disabled CId={}", clientId);
+            }
+        }
+
+        if (brokerConfig.isPeerCertificateAsUsername()) {
+            try {
+                // Use peer cert as username
+                SslHandler sslhandler = (SslHandler) channel.pipeline().get("ssl");
+                if (sslhandler != null) {
+                    Certificate[] certificateChain = sslhandler.engine().getSession().getPeerCertificates();
+                    userName = PemUtils.certificatesToPem(certificateChain);
+                }
+            } catch (SSLPeerUnverifiedException e) {
+                LOG.debug("No peer cert provided. CId={}", clientId);
+            } catch (CertificateEncodingException | IOException e) {
+                LOG.warn("Unable to decode client certificate. CId={}", clientId);
+            }
+        }
+
+        if (userName == null || userName.isEmpty()) {
+            if (brokerConfig.isAllowAnonymous()) {
+                return true;
+            } else {
+                LOG.info("Client didn't supply any credentials and MQTT anonymous mode is disabled. CId={}", clientId);
                 return false;
             }
-            final String login = msg.payload().userName();
-            if (!authenticator.checkValid(clientId, login, pwd)) {
-                LOG.info("Authenticator has rejected the MQTT credentials CId={}, username={}", clientId, login);
-                return false;
-            }
-            NettyUtils.userName(channel, login);
-        } else if (!brokerConfig.isAllowAnonymous()) {
-            LOG.info("Client didn't supply any credentials and MQTT anonymous mode is disabled. CId={}", clientId);
+        }
+
+        if (!authenticator.checkValid(clientId, userName, pwd)) {
+            LOG.info("Authenticator has rejected the MQTT credentials CId={}, username={}", clientId, userName);
             return false;
         }
+        NettyUtils.userName(channel, userName);
         return true;
     }
 
--- a/broker/src/main/java/io/moquette/broker/MQTTConnection.java
+++ b/broker/src/main/java/io/moquette/broker/MQTTConnection.java
@@ -414,18 +414,8 @@ final class MQTTConnection {
         }
 
         if (brokerConfig.isPeerCertificateAsUsername()) {
-            try {
-                // Use peer cert as username
-                SslHandler sslhandler = (SslHandler) channel.pipeline().get("ssl");
-                if (sslhandler != null) {
-                    Certificate[] certificateChain = sslhandler.engine().getSession().getPeerCertificates();
-                    userName = PemUtils.certificatesToPem(certificateChain);
-                }
-            } catch (SSLPeerUnverifiedException e) {
-                LOG.debug("No peer cert provided. CId={}", clientId);
-            } catch (CertificateEncodingException | IOException e) {
-                LOG.warn("Unable to decode client certificate. CId={}", clientId);
-            }
+            // Use peer cert as username
+            userName = readClientProvidedCertificates(clientId);
         }
 
         if (userName == null || userName.isEmpty()) {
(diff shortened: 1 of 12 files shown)


MikeDombo:
@andsel rebased PR ready for your review. The peer cert behavior is behind a BrokerConfiguration flag which is disabled by default.

broker/src/main/java/io/moquette/broker/security/PemUtils.java
```diff
@@ -0,0 +1,96 @@
+/*
+ * Copyright (c) 2022 The original author or authors
```
andsel:
```suggestion
 * Copyright (c) 2023 The original author or authors
```

broker/src/main/java/io/moquette/broker/security/PemUtils.java
```diff
@@ -0,0 +1,96 @@
...
+     * Copyright (c) 2000 - 2021 The Legion of the Bouncy Castle Inc. (https://www.bouncycastle.org)
+     * SPDX-License-Identifier: MIT
+     *
+     * <p>A generic PEM writer, based on RFC 1421
+     * From: https://javadoc.io/static/org.bouncycastle/bcprov-jdk15on/1.62/org/bouncycastle/util/io/pem/PemWriter.html</p>
+     */
+    public static class PemWriter extends BufferedWriter {
+        private static final int LINE_LENGTH = 64;
+        private final char[] buf = new char[LINE_LENGTH];
+        /**
```
andsel, line 53:
```suggestion

        /**
```

broker/src/main/java/io/moquette/broker/security/PemUtils.java
```diff
@@ -0,0 +1,96 @@
...
+         * Base constructor.
+         *
+         * @param out output stream to use.
+         */
+        public PemWriter(Writer out) {
+            super(out);
+        }
+        /**
```
andsel, line 62:
```suggestion

        /**
```

broker/src/main/java/io/moquette/broker/security/PemUtils.java
```diff
@@ -0,0 +1,96 @@
...
+         *
+         * @param type  key type.
+         * @param bytes encoded string
+         * @throws IOException IO Exception
+         */
+        public void writeObject(String type, byte[] bytes) throws IOException {
+            writePreEncapsulationBoundary(type);
+            writeEncoded(bytes);
+            writePostEncapsulationBoundary(type);
+        }
```
andsel, line 73:
```suggestion
        }
        
```

broker/src/main/java/io/moquette/broker/security/PemUtils.java
```diff
@@ -0,0 +1,96 @@
...
+                    if ((i + index) >= bytes.length) {
+                        break;
+                    }
+                    buf[index] = (char) bytes[i + index];
+                    index++;
+                }
+                this.write(buf, 0, index);
+                this.newLine();
+            }
+        }
```
andsel, line 89:
```suggestion
        }
        
```

broker/src/main/java/io/moquette/broker/security/PemUtils.java
```diff
@@ -0,0 +1,96 @@
...
+        private void writePreEncapsulationBoundary(String type) throws IOException {
+            this.write("-----BEGIN " + type + "-----");
+            this.newLine();
+        }
```
andsel, line 94:
```suggestion
        }
        
```

broker/src/main/java/io/moquette/broker/security/PemUtils.java
```diff
@@ -0,0 +1,96 @@
...
+import java.security.cert.CertificateEncodingException;
+import java.util.Base64;
+
+public final class PemUtils {
+    private static final String certificateBoundaryType = "CERTIFICATE";
+
+    public static String certificatesToPem(Certificate... certificates)
+        throws CertificateEncodingException, IOException {
+        try (StringWriter str = new StringWriter();
+             PemWriter pemWriter = new PemWriter(str)) {
```
andsel:
```suggestion
        try (StringWriter str = new StringWriter();
               PemWriter pemWriter = new PemWriter(str)) {
```

broker/src/main/java/io/moquette/broker/security/PemUtils.java
```diff
@@ -0,0 +1,96 @@
...
+            for (Certificate certificate : certificates) {
+                pemWriter.writeObject(certificateBoundaryType, certificate.getEncoded());
+            }
+            pemWriter.close();
```
andsel, line 37: But `pemWriter` isn't closed automatically by the `try` statement?

broker/src/main/java/io/moquette/broker/MQTTConnection.java
```diff
@@ -396,25 +402,46 @@ private void abortConnectionV5(MqttConnectReturnCode returnCode, ConnAckProperti
...
+                SslHandler sslhandler = (SslHandler) channel.pipeline().get("ssl");
+                if (sslhandler != null) {
+                    Certificate[] certificateChain = sslhandler.engine().getSession().getPeerCertificates();
+                    userName = PemUtils.certificatesToPem(certificateChain);
+                }
+            } catch (SSLPeerUnverifiedException e) {
+                LOG.debug("No peer cert provided. CId={}", clientId);
+            } catch (CertificateEncodingException | IOException e) {
+                LOG.warn("Unable to decode client certificate. CId={}", clientId);
+            }
```
andsel: This code could be extracted in a method named `readClientProvidedCertificates(channel)`

andsel (changes requested):
The idea seems good, however I would add an integration test with a real client that uses TLS connection and the server that validates the client identity using the provided certificate.

andsel:
> If the authorizer allows the operation, then a malicious client could exploit this to publish (or subscribe) to topics it should not normally be authorized on.

If the authorizer assumes that the operation is coming from the old client and denies the operation, then it is possible for operations which should be allowed to be be silently rejected (as Moquette does not disconnect clients after authZ failures).


Please could you describe better?
Then, do you think that if authz fails, Moquette should close the connection? I think this is a weakness in Moquette, so could be a separate issue (in case file one issue for that)

broker/src/main/java/io/moquette/broker/security/PemUtils.java
MikeDombo: It is closed automatically yes, however it needs to be closed before calling `str.toString()` or else the `str` will not contain the correct contents due to the buffered nature of `BufferedWriter` which `PemWriter` extends.

broker/src/main/java/io/moquette/broker/MQTTConnection.java
MikeDombo: Updated.

MikeDombo:
Thank you for the review @andsel.

I've updated the simple parts for formatting and extracting method.

> I would add an integration test with a real client that uses TLS connection and the server that validates the client identity using the provided certificate.

~~Are there any existing integration tests that you can point to which use either TLS or better yet, mutual TLS with client certs?~~ Nevermind, I found ServerIntegrationSSLClientAuthTest.

As for the authorizer bit of the description, I took that from the previous PR and it is possible that it no longer applies due to the significant session changes from 0.15 to 0.17. The idea was that an implementor of `IAuthorizatorPolicy` only gets the client ID and no other identifying information, so if the authorizer is caching information based on the client ID, then a malicious client connecting with the same client ID would inherit the permissions from the trusted client.

I will update the description to simplify since this is really just about providing a nicer way to offer TLS mutual auth for devices with client certificates.

As a separate point to this PR, yes I do think that unauthorized actions should kick the client (at least as an option); this is what AWS IoT Core does.

MikeDombo:
Added more testing with real TLS.

broker/src/test/java/io/moquette/integration/ServerIntegrationSSLClientAuthTest.java
```diff
@@ -186,16 +192,15 @@ protected void startServer(String dbPath) throws IOException {
         sslProps.put(IConfig.PERSISTENCE_ENABLED_PROPERTY_NAME, "true");
         sslProps.put(BrokerConstants.NEED_CLIENT_AUTH, "true");
         sslProps.put(IConfig.ENABLE_TELEMETRY_NAME, "false");
-        m_server.startServer(sslProps);
+        m_server.startServer(new MemoryConfig(sslProps), null, null, authenticator, null);
     }
 
-    @BeforeEach
```
andsel, line 192: I wouldn't miss the automatic call of common fixture setup. Instead, given that the authenticate with certificate is a new test case, we could add a new TestCase class, that contains only the 2 new added tests and common code with this class could be exported in an Abstract class. But the `BeforeEach` setup method autocall is something that should be preserved.

andsel (changes requested):
Thanks a lot @MikeDombo for this. I've a  concern about setup method, if you could separate would be great.
Add also a line in the Changelog please 🙏 

Thanks for the effort in sustaining the project.

broker/src/test/java/io/moquette/integration/ServerIntegrationSSLClientAuthTest.java
MikeDombo: Thank you for the reviews. Updated tests and extracted `ServerIntegrationSSLClientAuthBase` and put back the before each calls.

broker/src/test/java/io/moquette/integration/ServerIntegrationSSLClientAuthCertAsUsernameTest.java
```diff
@@ -0,0 +1,145 @@
+/*
+ * Copyright (c) 2012-2018 The original author or authors
```
andsel:
```suggestion
 * Copyright (c) 2012-2023 The original author or authors
```

broker/src/test/java/io/moquette/integration/ServerIntegrationSSLClientAuthCertAsUsernameTest.java
```diff
@@ -0,0 +1,145 @@
...
+        Properties sslProps = new Properties();
+        sslProps.put(IConfig.SSL_PORT_PROPERTY_NAME, "8883");
+        sslProps.put(IConfig.JKS_PATH_PROPERTY_NAME, "src/test/resources/serverkeystore.jks");
+        sslProps.put(IConfig.KEY_STORE_PASSWORD_PROPERTY_NAME, "passw0rdsrv");
+        sslProps.put(IConfig.KEY_MANAGER_PASSWORD_PROPERTY_NAME, "passw0rdsrv");
+        sslProps.put(IConfig.DATA_PATH_PROPERTY_NAME, dbPath);
+        sslProps.put(IConfig.PERSISTENCE_ENABLED_PROPERTY_NAME, "true");
+        sslProps.put(BrokerConstants.NEED_CLIENT_AUTH, "true");
+        sslProps.put(IConfig.ENABLE_TELEMETRY_NAME, "false");
+        sslProps.put(IConfig.PEER_CERTIFICATE_AS_USERNAME, "true");
```
andsel: These are almost the same set of properties defined in `ServerIntegrationSSLClientAuthTest` ([code ref](https://github.com/moquette-io/moquette/pull/775/files#diff-da4f6800f723568f557325a1e75c9cac9f684e9081a2f755ca4955be66d81376R62-R70)) so could be extracted in a method that prepare and return Properties instance in the abstract class.

andsel (changes requested):
We are really close, just one refactoring step and then the PR is perfect to be merged, I think.

broker/src/test/java/io/moquette/integration/ServerIntegrationSSLClientAuthCertAsUsernameTest.java
MikeDombo: Thank you, updated.

andsel (approved):
LGTM

PR #776, opened 2023-09-18
chore: keep java 8 compatibility

MikeDombo:
Maintain java 8 compatibility by just switching `Optional.isEmpty` to `!Optional.isPresent`

### DIFF
--- a/broker/src/main/java/io/moquette/broker/subscriptions/CTrie.java
+++ b/broker/src/main/java/io/moquette/broker/subscriptions/CTrie.java
@@ -2,10 +2,8 @@ package io.moquette.broker.subscriptions;
 
 import java.util.ArrayList;
 import java.util.Collections;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Optional;
-import java.util.Set;
 
 public class CTrie {
 
@@ -35,7 +33,7 @@ public class CTrie {
         Token token = topic.headToken();
         while (!topic.isEmpty()) {
             Optional<INode> child = inode.mainNode().childOf(token);
-            if (child.isEmpty()) {
+            if (!child.isPresent()) {
                 break;
             }
             topic = topic.exceptHeadToken();
--- a/pom.xml
+++ b/pom.xml
@@ -8,8 +8,8 @@
         <target.version>1.8</target.version>
         <junit.version>5.7.0</junit.version>
         <slf4j.version>1.7.36</slf4j.version>
-        <maven.compiler.source>11</maven.compiler.source>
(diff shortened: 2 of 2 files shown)


Kmrshubham335:
Could you let me know if the issue has been resolved or not for the contribution?

MikeDombo:
> Could you let me know if the issue has been resolved or not for the contribution?

Hi, I'm not sure what you're asking about. What issue are you referring to? This change allows Moquette to build with Java 8 again, which worked until recently when a change introduced a requirement for Java 11.

MikeDombo:
@andsel Thank you very much for reviewing my other PR. This one is also ready for review. I'm not sure on your Java version compatibility thoughts, but this is the only 1 line of code that requires Java 11 and the Java 8 version isn't really any worse (IMO).

andsel:
Java 8 is almost 10 years old. Last week Java 21 come out, I personally would push forward, for 0.18 would be nice to be not more older than JDK 17.

andsel:
Why still support JDK 8, as per [report](https://newrelic.com/resources/report/2023-state-of-the-java-ecosystem#toc-overview) the state of usage in production in 2023 for JDK 8  moved from 46% (2022) to 33% , so it's decreasing yoy, why still support it?

MikeDombo:
Right, I can certainly understand that. Of course you can do what you want, my only argument would be that raising the minimum version is not providing any benefit to the code. You may implement new features that rely on newer Java versions, but I'd just suggest to keep the minimum version the same until those new features are actually used.

This is your project so of course it is up to you, but I did want to offer my viewpoint and the simple fix (for the current state of what Java features are used).

andsel:
Moquette is tested on JDK 11 and 17, JDK 8 is not anymore supported on CI, so we could remain JDK 8 compatible, but we aren't testing it.
@MikeDombo maybe you have a stronger feeling about this, from your experience, Moquette is widely used with JDK 8?

MikeDombo:
I do not have any data to support it, but we use Moquette in Greengrass and Greengrass has a minimum supported version of JDK 8. The data from New Relic shows still 30% for Java 8 which seems pretty high still; and their data is primarily from cloud services too where it is easier to change the Java version. We're using Moquette in semi-embedded scenarios where people are unlikely to ever change the Java version if what they have is working.

Like I said in my previous reply, I feel that Java 8 should remain supported until you need (or maybe not need, but you at least want) a feature from a newer Java version which makes the code/project better.

andsel:
@MikeDombo  you convinced me :-)

andsel (approved):
LGTM

MikeDombo:
Ok awesome, thank you!

PR #778, opened 2023-09-18
fix: remove unused log placeholder

MikeDombo:
Super small change, just cleaning up a log line with an unused placeholder.


MikeDombo:
@andsel Thank you very much for reviewing my other PR. This one is also ready for review, and is very simple :)

andsel (approved):
LGTM

andsel:
Have to understand why 🔴 CI only in JDK 11 on ubuntu 10.04. Maybe it's transient.

PR #786, opened 2023-11-02
internalPublish now can send to specific clientId

maxwang789:
1.  internalPublish now can send to specific client 
2. add config to generate unique ClientID for each client

The original request is that one request will be sent to topic "/request", but the response should be sent back on another topic "/response".
With these modifications, the embedding broker could distinguish all connections. 

Correct me if I'm wrong.

### DIFF
--- a/broker/src/main/java/io/moquette/broker/MQTTConnection.java
+++ b/broker/src/main/java/io/moquette/broker/MQTTConnection.java
@@ -528,7 +528,7 @@ final class MQTTConnection {
                 LOG.debug("NOT processing disconnect {}, not bound.", clientID);
                 return null;
             }
-            if (protocolVersion == MqttVersion.MQTT_5.protocolLevel()) {
+            if (isProtocolVersion5()) {
                 MqttReasonCodeAndPropertiesVariableHeader disconnectHeader = (MqttReasonCodeAndPropertiesVariableHeader) msg.variableHeader();
                 if (disconnectHeader.reasonCode() != MqttReasonCodes.Disconnect.NORMAL_DISCONNECT.byteValue()) {
                     // handle the will
@@ -547,6 +547,10 @@ final class MQTTConnection {
             LOG.trace("dispatch disconnection userName={}", userName);
             return null;
         });
+    }
+
+    boolean isProtocolVersion5() {
+        return protocolVersion == MqttVersion.MQTT_5.protocolLevel();
     }
 
     PostOffice.RouteResult processSubscribe(MqttSubscribeMessage msg) {
@@ -833,5 +837,22 @@ final class MQTTConnection {
 
     public void bindSession(Session session) {
         bindedSession = session;
+    }
+
+    /**
+     * Invoked internally by broker to disconnect a client and close the connection
+     * */
+    void brokerDisconnect() {
+        final MqttMessage disconnectMsg = MqttMessageBuilders.disconnect().build();
+        channel.writeAndFlush(disconnectMsg)
+            .addListener(ChannelFutureListener.CLOSE);
+    }
+
+    void brokerDisconnect(MqttReasonCodes.Disconnect reasonCode) {
+        final MqttMessage disconnectMsg = MqttMessageBuilders.disconnect()
+            .reasonCode(reasonCode.byteValue())
+            .build();
+        channel.writeAndFlush(disconnectMsg)
+            .addListener(ChannelFutureListener.CLOSE);
     }
 }
--- a/broker/src/main/java/io/moquette/broker/PostOffice.java
+++ b/broker/src/main/java/io/moquette/broker/PostOffice.java
@@ -32,6 +32,7 @@ import io.netty.handler.codec.mqtt.MqttSubAckPayload;
 import io.netty.handler.codec.mqtt.MqttSubscribeMessage;
 import io.netty.handler.codec.mqtt.MqttTopicSubscription;
 import io.netty.util.ReferenceCountUtil;
+import org.apache.commons.codec.binary.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -276,6 +277,21 @@ class PostOffice {
                                         MQTTConnection mqttConnection) {
         // verify which topics of the subscribe ongoing has read access permission
         int messageID = messageId(msg);
+        final Session session = sessionRegistry.retrieve(clientID);
+
+        if (mqttConnection.isProtocolVersion5()) {
+            for (MqttTopicSubscription topicFilter : msg.payload().topicSubscriptions()) {
+                if (isSharedSubscription(topicFilter.topicName())) {
+                    final String shareName = extractShareName(topicFilter.topicName());
+                    if (!validateShareName(shareName)) {
+                        // this is a malformed packet, MQTT-4.13.1-1, disconnect it
+                        LOG.info("{} used an invalid shared subscription name {}, disconnecting", clientID, shareName);
+                        session.disconnectFromBroker();
+                        return;
+                    }
+                }
+            }
+        }
         List<MqttTopicSubscription> ackTopics = authorizator.verifyTopicsReadAccess(clientID, username, msg);
         MqttSubAckMessage ackMessage = doAckMessageFromValidateFilters(ackTopics, messageID);
 
@@ -292,7 +308,6 @@ class PostOffice {
         }
 
         // add the subscriptions to Session
-        Session session = sessionRegistry.retrieve(clientID);
         session.addSubscriptions(newSubscriptions);
 
         // send ack message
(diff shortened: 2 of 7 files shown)


hylkevds:
That seems to be a feature that inherently violates the MQTT spec. Every client that has a subscription to a given Topic must get every (QoS dependent) message sent to this Topic.

The MQTT 5 spec adds a request/response mechanism that explains how to send a message to a specific client, by negotiating a private Topic for this client.

maxwang789:
> That seems to be a feature that inherently violates the MQTT spec. Every client that has a subscription to a given Topic must get every (QoS dependent) message sent to this Topic.
> 
> The MQTT 5 spec adds a request/response mechanism that explains how to send a message to a specific client, by negotiating a private Topic for this client.

Yes, MQTT v5 add new mechanism. But it needs publisher and subscriber support new fields and new behaviors.
Sometime old equipment just uses MQTT as transport layer, it won't upgrade to MQTTv5.


And this modification is intended to be used in embedded server only, it won't change mqtt protocol or other behavior outside broker.

broker/src/main/java/io/moquette/BrokerConstants.java
```diff
@@ -77,6 +77,8 @@ public final class BrokerConstants {
     public static final String ALLOW_ANONYMOUS_PROPERTY_NAME = IConfig.ALLOW_ANONYMOUS_PROPERTY_NAME;
     public static final String REAUTHORIZE_SUBSCRIPTIONS_ON_CONNECT = "reauthorize_subscriptions_on_connect";
     public static final String ALLOW_ZERO_BYTE_CLIENT_ID_PROPERTY_NAME = "allow_zero_byte_client_id";
+
+    public static final String UNIQUE_CLIENT_ID="unique_client_id";
```
hylkevds: Please use the same formatting as other code.
(Not just on this line)

broker/src/main/java/io/moquette/broker/PostOffice.java
```diff
@@ -534,7 +534,7 @@ private RoutingResults publish2Subscribers(ByteBuf payload, Topic topic, MqttQoS
         final BatchingPublishesCollector collector = new BatchingPublishesCollector(sessionLoops);
 
         for (final Subscription sub : topicMatchingSubscriptions) {
-            if (filterTargetClients == NO_FILTER || filterTargetClients.contains(sub.getClientId())) {
+            if (Objects.isNull(filterTargetClients)||filterTargetClients.isEmpty()|| filterTargetClients.contains(sub.getClientId())) {
```
hylkevds: Why did you change this line?

broker/src/main/java/io/moquette/broker/PostOffice.java
```diff
@@ -659,12 +659,24 @@ static MqttQoS lowerQosToTheSubscriptionDesired(Subscription sub, MqttQoS qos) {
      *            the result of the enqueuing operation to session loops.
      */
     public RoutingResults internalPublish(MqttPublishMessage msg) {
+        return internalPublish(msg,Collections.emptySet());
```
hylkevds: This should use the `NO_FILTER` constant, not the `emptySet()`.

broker/src/main/java/io/moquette/broker/Server.java
```diff
@@ -556,14 +551,17 @@ private <T, U> T loadClass(String className, Class<T> intrface, Class<U> constru
      * @return
      */
     public RoutingResults internalPublish(MqttPublishMessage msg, final String clientId) {
+        return internalPublish(msg,new HashSet<>(Collections.singletonList(clientId)));
```
hylkevds: This changes the semantics of `clientId`.
> @param clientId the id of the sending integration.

In other words, `clientId` is the _sender_, not the _receiver_.

broker/src/test/java/io/moquette/integration/ServerIntegrationPahoCanPublishOnReadBlockedTopicTest.java
```diff
@@ -45,14 +46,16 @@
...
+import java.util.Collections;
+import java.util.HashSet;
 import java.util.Properties;
 
 import static io.moquette.broker.ConnectionTestUtils.EMPTY_OBSERVERS;
 import static java.nio.charset.StandardCharsets.UTF_8;
-import static org.junit.jupiter.api.Assertions.assertNotNull;
-import static org.junit.jupiter.api.Assertions.assertNull;
-import static org.junit.jupiter.api.Assertions.fail;
+import static org.junit.jupiter.api.Assertions.*;
```
hylkevds: Don't use star imports.

broker/src/test/java/io/moquette/integration/ServerIntegrationPahoCanPublishOnReadBlockedTopicTest.java
```diff
@@ -166,7 +220,7 @@ public void shouldNotInternalPublishOnReadBlockedSubscriptionTopic() throws Exce
             // it's OK, the subscribed should fail with error code 128
         }
 
-        m_server.internalPublish(message, "INTRLPUB");
+        m_server.internalPublish(message, new HashSet<>());
```
hylkevds: This line should not need to be changed.

broker/src/main/java/io/moquette/broker/MQTTConnection.java
```diff
@@ -193,11 +193,16 @@ PostOffice.RouteResult processConnect(MqttConnectMessage msg) {
             }
 
             // Generating client id.
-            clientId = UUID.randomUUID().toString().replace("-", "");
+            clientId = genClientId();
             serverGeneratedClientId = true;
             LOG.debug("Client has connected with integration generated id: {}, username: {}", clientId, username);
         } else {
-            serverGeneratedClientId = false;
+            if(brokerConfig.isUniqueClientId()){
```
hylkevds: What is the reason for this code? ClientIds in MQTT are already unique.

hylkevds (changes requested):
I've added some questions and change requests.

broker/src/main/java/io/moquette/BrokerConstants.java
maxwang789: I will change it to FORCE_UNIQUE_CLIENT_ID_PROPERTY_NAME

hylkevds: The comment was not about the name of the property, but about the fact that the spaces around the equals-sign (`=`) are missing.

The fact that the client id is already unique per MQTT spec is a separate issue. I don't think this property is needed.

broker/src/main/java/io/moquette/broker/MQTTConnection.java
maxwang789: NO, according to my test result.
1. If the client id is empty, it will generate a new clientId with new UUID
2. If the client is not empty, it just accepts original clientId. It may cause duplicate clientIds.

broker/src/main/java/io/moquette/broker/PostOffice.java
maxwang789: In previous logic, the empty filter means no matched.
It will filter out all clients and do nothing.
I just change the logic, no filter means everyone, it will send to every client without any filter.

hylkevds: The previous logic worked correctly, with simpler logic. If the goal is to send to everyone, pass the NO_FILTER constant. If the goal is to send to a specific list of clients, use a list.

There is no need to change this.

broker/src/main/java/io/moquette/broker/PostOffice.java
maxwang789: OK, I will change to NO_FILTER.
But I thought NO_FILTER should be pointed to Collections.EmptySet also because the intent of the variable means empty set and it should not be modified. 
A normal HashSet may be changed by mistake and cause unexpected behavior.

broker/src/main/java/io/moquette/broker/MQTTConnection.java
hylkevds: Reconnecting with the same client id will cause the previous connection with the same client id to be disconnected. Client IDs are unique, as specified in the MQTT specification.
Your code makes it impossible for a client to reconnect with the same client id, and receive all the messages that were not delivered yet due to the lost connection. This is a key part of the MQTT specification.

broker/src/main/java/io/moquette/broker/Server.java
maxwang789: oh, I may misunderstand the meaning of clientId here because the original clientId was discarded.
Should I introduce a new method such as internalPublishTo ?

broker/src/main/java/io/moquette/broker/PostOffice.java
hylkevds: NO_FILTER has different semantics from `Collections.EmptySet`. That's why NO_FILTER is not set to `Collections.EmptySet`. (`Collections.EmptySet` itself is a constant, so that's a very bad candidate for using as a filter contant)

Passing an empty list will cause the message to be sent to no-one. Which is consistent with the semantics of the method call.

PR #817, opened 2024-02-11
Refactoring: converted loop into stream processing

andsel:
## Release notes

[rn:skip]

## What does this PR do?

Converted a loop with various stages (subscription add and filtering to send retained) into fluent stream map and filter


hylkevds (approved):
Very nice, exactly the same size, but a lot more readable :)

andsel:
Thank's @hylkevds. Latest development of Moquette regard about to make it MQTT5 compliant. I've just completed the adaptation for subscriptions, if you want give it a try from the `main` branch , would be great, thanks 🙏

PR #832, opened 2024-04-28
Fixed #831: changed several cases where a shared readerIndex was changed

### LINKED ISSUE #831: Shrared buffer issues due to readerIndex
ByteBufs are shared between many sessions and threads. Because if this it is important that none of these session change the `readerIndex` of the shared buffer, and when changing the readerIndex can not be avoided, a `duplicate()` must be made.

In most cases this can be fixed by using the `.getX()` methods instead of the `.readX()` versions.

A PR is on its way.

hylkevds:
DebugUtils.payload2Str set the readerIndex to the readableBytes, witch was simply wrong. 

Utils.readBytesAndRewind was unused and thus removed. Though it changed the readerIndex, it also reset it, but this use would still have not been thread-safe.

SegmentedPersistentQueueSerDes use of readBytes (that changes the readerIndex) was changed to getBytes which does not change the readerIndex.

Closes #831

### DIFF
--- a/broker/src/main/java/io/moquette/broker/DebugUtils.java
+++ b/broker/src/main/java/io/moquette/broker/DebugUtils.java
@@ -23,10 +23,7 @@ import java.nio.charset.StandardCharsets;
 public final class DebugUtils {
 
     public static String payload2Str(ByteBuf content) {
-        final int readerPin = content.readableBytes();
-        final String result = content.toString(StandardCharsets.UTF_8);
-        content.readerIndex(readerPin);
-        return result;
+        return content.toString(StandardCharsets.UTF_8);
     }
 
     private DebugUtils() {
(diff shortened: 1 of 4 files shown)


andsel:
Hi @hylkevds thanks for working on this fix.

> DebugUtils.payload2Str set the readerIndex to the readableBytes, witch was simply wrong.

You are right, should be `readerIndex()`, if the buffer isn't shared.

All changes look good, but would be nice to have this covered by a unit test, because or code wasn't tested or the coverage of these lines was not so complete and the bug didn't pop up.

hylkevds:
The writePayload one is testable, added a test for it.

andsel (approved):
LGTM

andsel:
Thank's for contributing the fix @hylkevds

Issue #734, opened 2023-03-03
[Question] Potential memory leak


basimons:
Hello,

I don't know if this is the right place to ask, if not, please let me know, I'll transfer this to stackoverflow.

We are using Moquette for our IOT data, and after applying some stateless transformation on the payload we put it in kafka. This all works as expected and is very performant, thank you for that!

However, we've been noticing that there is a memory leak in our service. I'm posting it here, as we also have a similar service who does basically the same, but listens to IOT data over http.

Our memory usage can be seen here:
![image](https://user-images.githubusercontent.com/49156770/222756373-1b61ab62-2776-4a52-9f8e-d788d1e4a058.png)
Where the yellow line is the DirectBuffer part and the red line is the tenured gen.

The memory according to kubernetes (`kubernetes.memory.working_set`) is also steadily increasing, with roughly the same speed (indicating that it is probably the byte buffer). 
![image](https://user-images.githubusercontent.com/49156770/222756198-4b15931f-b562-4690-afcd-2421f4ec5195.png)

Initially after exploring the heap dump, we didn't find anything strange. We also used jemalloc to track off heap memory usage, but that also is not pointing to anything in particular.

After doing some more reading, I ran this OQL query: `SELECT x AS ByteBuffer, x.capacity AS Capacity, x.limit AS Limit, x.mark AS Mark, x.position AS Position FROM java.nio.DirectByteBuffer x WHERE ((x.capacity > (1024 * 1024)) and (x.cleaner != null))`, using MAT, on the heap dump. The output of this seams to point to a potential issue as we get roughly 400 `java.nio.DirectByteBuffer`'s with a size of `4194304` bytes. Am I correct in assuming there are 400 different ByteBufs of roughly 4mB, or am I wrong and there are just a bunch of ByteBufs pointing to the same byte[] underneath and there is no issue at all?

See output here:
![image](https://user-images.githubusercontent.com/49156770/222769002-3ac018f1-6ff5-4650-97cb-d2663b9890d6.png)

Does anybody know if we might be using the Moquette library in a wrong way, that could cause such a thing? Because we are not using any direct byte buffers by ourselves, we only use a ByteBuf once when we create a `MqttPublishMessage`, which gets released by the `Server#internalPublish` method. If not, what would our next best step be, to figure out what is causing this?

We'd greatly appreciate some insight/tips/help, as we've been looking at it for a while, but can't seem to figure out what the cause is.

Thanks in advance!

hylkevds:
[...]

PR #833, opened 2024-04-29
Fixed SegmentedQueues not being cleaned up on session purge

hylkevds:
Segmented queues were closed, but not cleared on session close. This resulted in segments not being returned to the pool, and thus disk-space not being released. The queues were also not removed from the QueuePool, resulting in a memory leak.

Relates #734

### DIFF
--- a/broker/src/main/java/io/moquette/broker/unsafequeues/QueuePool.java
+++ b/broker/src/main/java/io/moquette/broker/unsafequeues/QueuePool.java
@@ -389,9 +389,10 @@ public class QueuePool {
         final LinkedList<SegmentRef> segmentRefs = queueSegments.remove(queueName);
         SegmentRef segmentRef = segmentRefs.pollLast();
         segmentsAllocationLock.lock();
+        LOG.debug("Purging segments for queue {}", queueName);
         try {
             while (segmentRef != null) {
-                LOG.debug("Consumed tail segment {} from queue {}", segmentRef, queueName);
+                LOG.debug("Purging segment {} from queue {}", segmentRef, queueName);
                 recycledSegments.add(segmentRef);
                 segmentRef = segmentRefs.pollLast();
             }
--- a/broker/src/main/java/io/moquette/broker/unsafequeues/QueuePool.java
+++ b/broker/src/main/java/io/moquette/broker/unsafequeues/QueuePool.java
@@ -384,6 +384,23 @@ public class QueuePool {
         }
     }
 
+    void purgeQueue(String name) {
+        final QueueName queueName = new QueueName(name);
+        final LinkedList<SegmentRef> segmentRefs = queueSegments.remove(queueName);
+        SegmentRef segmentRef = segmentRefs.pollLast();
+        segmentsAllocationLock.lock();
+        try {
+            while (segmentRef != null) {
+                LOG.debug("Consumed tail segment {} from queue {}", segmentRef, queueName);
+                recycledSegments.add(segmentRef);
+                segmentRef = segmentRefs.pollLast();
+            }
+        } finally {
+            segmentsAllocationLock.unlock();
+        }
+        queues.remove(queueName);
+    }
+
     /**
      * Free mapped files
      * */
--- a/ChangeLog.txt
+++ b/ChangeLog.txt
@@ -1,4 +1,5 @@
 Version 0.19-SNAPSHOT
+   [fix] Fixed SegmentedQueues not being cleaned up on session purge. (#833)
 
 Version 0.18:
    [fix] Update Netty to 4.1.116 and H2 2.1.214
--- a/broker/src/main/java/io/moquette/broker/unsafequeues/Queue.java
+++ b/broker/src/main/java/io/moquette/broker/unsafequeues/Queue.java
@@ -157,6 +157,15 @@ public class Queue {
     }
 
     /**
+     * Close the Queue and release all resources.
+     */
+    public void close() {
+        queuePool.purgeQueue(name);
+        headSegment = null;
+        tailSegment = null;
+    }
+
+    /**
      * Read next message or return null if the queue has no data.
      * */
     public Optional<ByteBuffer> dequeue() throws QueueException {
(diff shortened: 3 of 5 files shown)


broker/src/main/java/io/moquette/broker/SessionRegistry.java
```diff
@@ -483,15 +482,15 @@ private void purgeSessionState(Session session) {
...
         remove(session.getClientID());
-
+        sessionsRepository.delete(session.getSessionData());
         subscriptionsDirectory.removeSharedSubscriptionsForClient(session.getClientID());
     }
 
     void remove(String clientID) {
         final Session old = pool.remove(clientID);
         if (old != null) {
+            unsubscribe(old);
```
andsel: Are we sure that `pool`'s session is the same instance of current session?

andsel (changes requested):
Hi @hylkevds thanks for this PR. I would also add unit test on the `unsafequeues.Queue` or `unsafequeues.QueuePool` to verify that a queue purge or close give back the segments to be reused.

broker/src/main/java/io/moquette/broker/SessionRegistry.java
hylkevds: The only situation where it could happen that there are two sessions with the same clientID is when a client re-connects with cleanSession=true. In that case the old session is purged before the new session is created.

I can change the code to only remove from the pool if the session is the same as the one in the pool.

That just lead me to a different potential issue, with the `sessionExpirationService` timer. I think that timer could trigger right as a client is reconnecting, and as it is based on the `clientID`, it could kill a session that was just made. Shouldn't `removeExpiredSession` check if the session really is expired, and not reconnected? And probably do the cleanup on the session thread?

hylkevds:
> Hi @hylkevds thanks for this PR. I would also add unit test on the `unsafequeues.Queue` or `unsafequeues.QueuePool` to verify that a queue purge or close give back the segments to be reused.

I'll look into making some tests for this.

broker/src/main/java/io/moquette/broker/SessionRegistry.java
andsel: > I think that timer could trigger right as a client is reconnecting, and as it is based on the clientID, it could kill a session that was just made. Shouldn't removeExpiredSession check if the session really is expired, and not reconnected? And probably do the cleanup on the session thread?

Why do you think that that case could happen? Looking at the code, when a session disconnect it's also untracked from the session expiration service. When it reconnects it's tracked again, reinserting the reference into the pool managed by the session expiration service.
In this way all session tracked by the expiration service are session present in the session registry, and so can't be the case of a duplication.

hylkevds: The session expiration service only contains _disconnected_ sessions that have _cleanSession=false_. Sessions that are connected don't need to be tracked since they can't expire, and sessions with cleanSession=true are cleaned up on disconnect.

Imagine a disconnected session with cleanSession=false. The expiration service finds it in the queue and removes it for cleaning, but there is a time interval between it being removed from the queue and it being actually cleaned. Either because there is a large batch of sessions to expire, or because the thread is put in hold. In that time the user reconnects, and a new session is made that inherits the subscriptions and clientID of the old session. Now the expiration thread is reactivated and kills the thread with the stored clientID...

andsel: @hylkevds you are right! Very good explanation of the case. I think this aspect should be kept out of this PR, and an issue tracking the problem should opened so be solved. Thanks for pointing it out 👍

hylkevds:
@andsel: I added a test to verify the queue cleanup and page-file re-use.

broker/src/main/java/io/moquette/broker/unsafequeues/QueuePool.java
```diff
@@ -384,6 +384,23 @@ public Queue getOrCreate(String queueName) throws QueueException {
         }
     }
 
+    void purgeQueue(String name) {
+        final QueueName queueName = new QueueName(name);
+        final LinkedList<SegmentRef> segmentRefs = queueSegments.remove(queueName);
+        SegmentRef segmentRef = segmentRefs.pollLast();
+        segmentsAllocationLock.lock();
```
andsel, line 391: I would add a debug line here, like
```java
LOG.debug("Purging segments for queue {}", queueName);
```

broker/src/main/java/io/moquette/broker/unsafequeues/QueuePool.java
```diff
@@ -384,6 +384,23 @@ public Queue getOrCreate(String queueName) throws QueueException {
...
+        try {
+            while (segmentRef != null) {
+                LOG.debug("Consumed tail segment {} from queue {}", segmentRef, queueName);
```
andsel: Would rewrite this log as

```suggestion
                LOG.debug("Purging segment {} from queue {}", segmentRef, queueName);
```

andsel (changes requested):
Hi @hylkevds the changes are almost good, just a couple of nitpicks on the log messages, then the PR is done.

broker/src/main/java/io/moquette/broker/unsafequeues/QueuePool.java
hylkevds: No problem, I added both logging lines.

broker/src/main/java/io/moquette/broker/SessionRegistry.java
```diff
@@ -490,19 +489,19 @@ private void purgeSessionState(Session session) {
...
-
+        remove(session);
+        sessionsRepository.delete(session.getSessionData());
         subscriptionsDirectory.removeSharedSubscriptionsForClient(session.getClientID());
     }
 
-    void remove(String clientID) {
-        final Session old = pool.remove(clientID);
-        if (old != null) {
+    void remove(Session session) {
```
andsel: I would keep the fixes for session management out of this PR, so that's more clear what this PR does and simplify to understand the change history, reducing the scope of the PR

andsel (changes requested):
Please keep this PR with only changes related to the persistent queues, that would help a lot the review and the change history of the project. 
Please create a follow up PR for the issue related to session management.

broker/src/main/java/io/moquette/broker/SessionRegistry.java
hylkevds: I've split the commit containing the SessionRegistry changes into a separate branch. PR coming soon...

andsel (approved):
LGTM

andsel:
Thanks @hylkevds for your contribution to the project!

PR #834, opened 2024-04-29
Reduced memory use on Sessions by removing handled items from inflightTimeouts

hylkevds:
Reduced memory use on busy Sessions by removing handled items from inflightTimeouts

Items in the inflightTimeouts DelayQueue were only ever removed when they timed out. But in normal operation the related messages would have been handled long before that. There should only ever be a number equal to the number of inflightSlots in the queue, but the queue would grow to the maximum number of messages ever handled in a 5 second interval. This made each session take much more memory than needed.

- Relates #734

### DIFF
--- a/broker/src/main/java/io/moquette/broker/Session.java
+++ b/broker/src/main/java/io/moquette/broker/Session.java
@@ -192,6 +192,7 @@ class Session {
 
     public void processPubRec(int pubRecPacketId) {
         // Message discarded, make sure any buffers in it are released
+        cleanFromInflight(pubRecPacketId);
         SessionRegistry.EnqueuedMessage removed = inflightWindow.remove(pubRecPacketId);
         if (removed == null) {
             LOG.warn("Received a PUBREC with not matching packetId");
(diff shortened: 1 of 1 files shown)


broker/src/main/java/io/moquette/broker/Session.java
```diff
@@ -343,6 +345,7 @@ private boolean inflightHasSlotsAndConnectionIsUp() {
 
     void pubAckReceived(int ackPacketId) {
         // TODO remain to invoke in somehow m_interceptor.notifyMessageAcknowledged
+        inflightTimeouts.removeIf(d -> d.packetId == ackPacketId);
```
andsel: Please extract this on a method such as `cleanFromInflight`

```java
private void cleanFromInflight(int packetId) {
    inflightTimeouts.removeIf(d -> d.packetId == packetId);
}     
```

andsel (changes requested):
The idea is correct, just a nitpick to make it more expressive.

broker/src/main/java/io/moquette/broker/Session.java
hylkevds: Wasn't sure which form you'd prefer, since it was only one line.
Commit Updated.

andsel (approved):
LGTM

PR #835, opened 2024-04-29
Fixed ExpirationService not shutting down Executor

hylkevds:
When shutting down the server the Executor Threads in ScheduledExpirationService instances were not terminated, resulting in a hanging server that would not shut down.

### DIFF
--- a/broker/src/main/java/io/moquette/broker/scheduler/ScheduledExpirationService.java
+++ b/broker/src/main/java/io/moquette/broker/scheduler/ScheduledExpirationService.java
@@ -25,13 +25,14 @@ public class ScheduledExpirationService<T extends Expirable> {
     private final ScheduledFuture<?> expiredEntityTask;
     private final Clock clock;
     private final Consumer<T> action;
+    private final ScheduledExecutorService actionsExecutor;
 
     private final Map<String, ExpirableTracker<T>> expiringEntitiesCache = new HashMap<>();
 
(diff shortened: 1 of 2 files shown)


broker/src/main/java/io/moquette/broker/scheduler/ScheduledExpirationService.java
```diff
@@ -25,13 +25,14 @@ public class ScheduledExpirationService<T extends Expirable> {
     private final ScheduledFuture<?> expiredEntityTask;
     private final Clock clock;
     private final Consumer<T> action;
+    ScheduledExecutorService actionsExecutor;
```
andsel: Given that `actionsExecutor` is assigned only once in the constructor, this can be 

```suggestion
    private final ScheduledExecutorService actionsExecutor;
```

andsel (changes requested):
The idea is correct, just a nit pick on the the visibility of the newly introduced field.

broker/src/main/java/io/moquette/broker/scheduler/ScheduledExpirationService.java
hylkevds: Oops, fixed :)

andsel (approved):
LGTM

andsel:
Thank's @hylkevds for fixing it!

PR #836, opened 2024-04-30
Fixed empty shared subscriptions list remaining forever

hylkevds:
Even for clients that never made shared subscriptions, upon session clean, an empty shared subscriptions list was created and never removed.

Relates #734


andsel (approved):
LGTM

andsel:
Thanks @hylkevds for your contribution.

PR #837, opened 2024-04-30
Optimised page file opening

hylkevds:
Instead of mapping the entire page file every time a segment is re-used, we map the page file once, and re-use the mapped buffer for each segment.
The cache uses WeakReferences to ensure that a page file that is no longer used does not stay mapped.

### DIFF
--- a/broker/src/main/java/io/moquette/broker/unsafequeues/PagedFilesAllocator.java
+++ b/broker/src/main/java/io/moquette/broker/unsafequeues/PagedFilesAllocator.java
@@ -1,12 +1,17 @@
 package io.moquette.broker.unsafequeues;
 
 import java.io.IOException;
+import java.lang.ref.WeakReference;
 import java.nio.MappedByteBuffer;
 import java.nio.channels.FileChannel;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.StandardOpenOption;
+import java.util.HashMap;
+import java.util.Map;
 import java.util.Properties;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Default implementation of SegmentAllocator. It uses a series of files (named pages) and split them in segments.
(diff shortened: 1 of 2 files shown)


hylkevds:
It seems the 200ms timeout on creating subscriptions is sometimes not enough on the GitHub CI environment:
https://github.com/moquette-io/moquette/blob/2092693c5d9d5bcb0468631416d064c3707ffa32/broker/src/test/java/io/moquette/testclient/Client.java#L211

Do you want me to create a PR to increase it to 300?

broker/src/main/java/io/moquette/broker/unsafequeues/PagedFilesAllocator.java
```diff
@@ -36,11 +46,30 @@ interface AllocationListener {
         this.segmentSize = segmentSize;
         this.lastPage = lastPage;
         this.lastSegmentAllocated = lastSegmentAllocated;
-        this.currentPage = openRWPageFile(this.pagesFolder, this.lastPage);
+        this.currentPage = getPageFile(this.lastPage);
+    }
+
+    private MappedByteBuffer getPageFile(int pageId) throws QueueException {
```
andsel: I would rename, get<something> is confusing, a getter with parameter. Better something more explicit, I would propose 
```suggestion
    private MappedByteBuffer openOrRetrievePageFile(int pageId) throws QueueException {
```

andsel (changes requested):
The idea is good, I left a rename comment. 
Plus I would ask to add a line in changelog 🙏

broker/src/main/java/io/moquette/broker/unsafequeues/PagedFilesAllocator.java
hylkevds: Good point about not making getters with parameters, or getters that do more complicated stuff than just getting...
Fixed it!

hylkevds: Also added the changelog entry.

andsel (approved):
LGTM

andsel:
Thank's @hylkevds for your contribution!

PR #842, opened 2024-06-13
Fixes 841: Incorrect reference used in compareAndSet in CTrie.cleanTomb.

### LINKED ISSUE #841: High CPU usage when run Ctrie.insert function
### Expected behavior
I found that the cpu usage of mqtt broker is high after running for one month，the cpu usage increase from 9% to 95%
The number of CPU cores of the server is 24U 
I found that the ctrie tree is very large when the cpu usage is high, the tree contains a large number of tomb nodes (TNode)
In this case, a large number of CPU resources are consumed by running function: CNode.anychidrenmatch.

1、Why do a large number of TNode nodes exist in the system and are not cleared?

2、Hope that the TNode node can be automatically cleared or a thread can be used to automatically clear the TNode when the TNode is not used.

3、What the following scenarios does the system not automatically clear TNodes?       

 if (cnode instanceof TNode) {
            // this inode is a tomb, has no clients and should be cleaned up
            // Because we implemented cleanTomb below, this should be rare, but possible
            // Consider calling cleanTomb here too
            return Action.OK;
        }

### Actual behavior
when the usage is 5%, the number of TNodes is 151, but when the cpu usage is up to 95% , the number of TNodes  is 7,065,591
The TNodes are not cleared, and  the number is been increasing over time.

### Moquette MQTT version
0.16
https://github.com/moquette-io/moquette/archive/refs/tags/v0.16.tar.gz

1.8

hylkevds:
cleanTomb used compareAndSet to update a reference, but incorrectly re-fetched the 'original' instead of using the version that was used to make the copy. The result was that in case of a conflict, the changes of one thread would be overwritten by another thread.

If cleanTomb failes when called from the remove method, the node is already replaced by a TNode, and the subsequent call will
re-try to clean the TNode. The node removed by cleanTomb may already have been replaced with a live node by another thread, so
cleanTomb checks if the removed node actually was the intended node before committing the results.

- bug
- Closes #841

### DIFF
--- a/ChangeLog.txt
+++ b/ChangeLog.txt
@@ -1,4 +1,5 @@
 Version 0.18-SNAPSHOT:
+   [fix] Incorrect reference used in compareAndSet in CTrie.cleanTomb. (#841)
    [feature] Implement response-information property for request-response flow. (#840)
    [fix] Optimised page file opening for disk-based queues. (#837)
    [feature] Manage payload format indicator property, when set verify payload format. (#826)
--- a/broker/src/main/java/io/moquette/broker/subscriptions/CNode.java
+++ b/broker/src/main/java/io/moquette/broker/subscriptions/CNode.java
@@ -99,9 +99,9 @@ class CNode implements Comparable<CNode> {
         }
     }
 
-    public void remove(INode node) {
+    public INode remove(INode node) {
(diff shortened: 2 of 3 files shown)


andsel (approved):
LGTM

Great catch @hylkevds a BIG 👍 thank's for the fix

PR #843, opened 2024-06-17
Increased timeout to fix build failure on slow GitHub build systems

hylkevds:
Builds are sometimes failing on the GH build systems due to a timeout. This PR increases those timeouts from 200ms to 300 ms and puts the actual value of the timeouts in a constant for easier access.

### DIFF
--- a/broker/src/test/java/io/moquette/testclient/Client.java
+++ b/broker/src/test/java/io/moquette/testclient/Client.java
@@ -13,12 +13,10 @@
  *
  * You may elect to redistribute this code under either of these licenses.
  */
-
 package io.moquette.testclient;
 
 import io.moquette.BrokerConstants;
 import io.netty.bootstrap.Bootstrap;
-import io.netty.buffer.ByteBuf;
 import io.netty.channel.*;
 import io.netty.channel.nio.NioEventLoopGroup;
 import io.netty.channel.socket.SocketChannel;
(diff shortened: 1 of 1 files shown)


broker/src/test/java/io/moquette/testclient/Client.java
```diff
@@ -52,6 +50,9 @@ public interface ICallback {
     }
 
     private static final Logger LOG = LoggerFactory.getLogger(Client.class);
+
+    private static final int TIMEOUT_DURATION_MS = 300;
```
andsel: I would suggest to create a `Duration` and then change the usage:
```suggestion
    private static final Duration TIMEOUT_DURATION = Duration.ofMillis(300);
```

broker/src/test/java/io/moquette/testclient/Client.java
```diff
@@ -195,7 +196,7 @@ public MqttSubAckMessage subscribe(String topic1, MqttQoS qos1, String topic2, M
             .addSubscription(qos2, topic2)
             .build();
 
-        return doSubscribeWithAckCasting(subscribeMessage, 200, TimeUnit.MILLISECONDS);
+        return doSubscribeWithAckCasting(subscribeMessage, TIMEOUT_DURATION_MS, TimeUnit.MILLISECONDS);
```
andsel:
```suggestion
        return doSubscribeWithAckCasting(subscribeMessage, TIMEOUT_DURATION.toMillis(), TimeUnit.MILLISECONDS);
```

andsel (changes requested):
This is fine, just a suggestion on the constant to transform in a `Duration` which is more explicit express what is it.

hylkevds:
> This is fine, just a suggestion on the constant to transform in a `Duration` which is more explicit express what is it.

Good idea!
I did change a few int parameters to long, to avoid casting.
It's a pity that `CountDownLatch.await()` doesn't accept a Duration, that would have been even cleaner :)

andsel (approved):
LGTM

PR #872, opened 2024-11-23
Fix/resolve flaky error in tests

andsel:
## Release notes
[rn:skip]

## What does this PR do?
Fixes some flaky test acting on 3 things:
- drop from custom low level client the callback mechanism and the queue, but resort just in the queued messages an wait blocking on it.
- updates the Publish collect based on Paho client to release the latch only after having copied the data, to avoid nulls due to read ordering
From 
```java
latch.countDown();
receivedTopic = topic;
receivedMessage = message;
```
to
```java
receivedTopic = topic;
receivedMessage = message;
latch.countDown();
```
- in HiveMQ client usages, register the listener to subscribe before the verification and not during the verification. This makes the code less linear, but it's the intended use of the API.

### DIFF
--- a/broker/src/test/java/io/moquette/integration/mqtt5/PayloadFormatIndicatorTest.java
+++ b/broker/src/test/java/io/moquette/integration/mqtt5/PayloadFormatIndicatorTest.java
@@ -23,12 +23,10 @@ import com.hivemq.client.mqtt.mqtt5.Mqtt5BlockingClient;
 import com.hivemq.client.mqtt.mqtt5.exceptions.Mqtt5PubAckException;
 import com.hivemq.client.mqtt.mqtt5.exceptions.Mqtt5PubRecException;
 import com.hivemq.client.mqtt.mqtt5.message.publish.Mqtt5PayloadFormatIndicator;
-import com.hivemq.client.mqtt.mqtt5.message.publish.Mqtt5PublishResult;
 import com.hivemq.client.mqtt.mqtt5.message.publish.puback.Mqtt5PubAckReasonCode;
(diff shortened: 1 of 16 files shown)


broker/src/test/java/io/moquette/integration/mqtt5/PayloadFormatIndicatorTest.java
```diff
@@ -64,19 +73,23 @@ public String clientName() {
...
+        assertThat(subAck.getReasonCodes()).contains(Mqtt5SubAckReasonCode.GRANTED_QOS_1);
+        LOG.info("SUBACK received");
 
         Mqtt5BlockingClient publisher = createPublisherClient();
         publisher.publishWith()
             .topic("temperature/living")
             .payload("18".getBytes(StandardCharsets.UTF_8))
             .payloadFormatIndicator(Mqtt5PayloadFormatIndicator.UTF_8)
-            .qos(MqttQos.AT_MOST_ONCE)
+            .qos(MqttQos.AT_LEAST_ONCE)
```
andsel: Switch to QoS1 so that we are sure the client receives and bubble up to test code the publish

broker/src/test/java/io/moquette/integration/mqtt5/SubscriptionOptionsTest.java
```diff
@@ -54,9 +58,9 @@ static class PublishCollector implements IMqttMessageListener {
 
         @Override
         public void messageArrived(String topic, MqttMessage message) throws Exception {
-            latch.countDown();
             receivedTopic = topic;
             receivedMessage = message;
+            latch.countDown();
```
andsel, line 59: Switch the order so that the latch release happens only after the data has been copied, else the client can read null values.

PR #899, opened 2025-03-27
Update moquette-broker coordinates to resolve the 401 unauthorized error from Jitpack

dskarzh:
This PR corrects instructions in `README.md` on how to embed moquette in other projects

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,9 +40,9 @@ In repositories section, add:
 In dependencies section add:
 ```
 <dependency>
-  <groupId>com.github.moquette-io</groupId>
+  <groupId>com.github.moquette-io.moquette</groupId>
   <artifactId>moquette-broker</artifactId>
-  <version>0.18.0</version>
+  <version>v0.18.0</version>
 </dependency>
 ```
 
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ In dependencies section add:
 <dependency>
   <groupId>com.github.moquette-io.moquette</groupId>
   <artifactId>moquette-broker</artifactId>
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -40,9 +40,9 @@ In repositories section, add:
 In dependencies section add:
 ```
 <dependency>
-  <groupId>com.github.moquette-io</groupId>
+  <groupId>com.github.moquette-io.moquette</groupId>
   <artifactId>moquette-broker</artifactId>
-  <version>0.18.0</version>
+  <version>v0.18.0</version>
```
andsel: Ahh it's the tag name and Moquette always used `vM.M.P` format. I think it's ugly to use 
```
<version>v0.18.0</version>
```

If you also feel the same, I think I should switch tag policy to remove the `v` from the tag.
In past I used that to distinguish tags from version branches.

Let me know what do you think

andsel (commented):
Left a comment to know your opinion on the version name `v0.18.0` vs `0.18.0`

README.md
dskarzh: I completely agree that the `v` prefix is not only ugly but also unusual in the Java ecosystem.

andsel: Created tag `0.18.0` https://github.com/moquette-io/moquette/tree/0.18.0 so you could drop the `v` from the `version` tag.

dskarzh: Thanks. That works, though group ID still missing `.moquette`.  I updated the PR to omit `v` prefix.

andsel (approved):
LGTM

andsel:
Thanks @dskarzh for your contribution!

PR #900, opened 2025-03-28
MQTTv3.1.1 supports a zero-byte clientId during authentication

l2063610646:
Bug:
According to the MQTT v3.1.1 protocol specification, MQTT clients can send connect messages without specifying the client ID. This is not a feature exclusive to MQTTV5, **it should also exist in MQTTv3.1.1**.

I think brokers should generate random clientId instead of failing authentication.


andsel (approved):
LGTM

PR #910, opened 2025-05-28
Added metrics framework and implementation

hylkevds:
- enhancement

## Release notes
   [feature] Added metrics framework and Prometheus implementation.
     - Enable by setting `metrics_provider_class` to `MetricsProviderPrometheus`.
     - Metrics endpoint can be found on http://localhost:9400/metrics by default.
     - port can be changed with `metrics_endpoint_port`. Set to 0 to disable http endpoint.

## What does this PR do?

To gain insights into the state of running software, the software can expose metrics, which are then gathered by a metrics framework.
A popular framework for this is [Prometheus](https://prometheus.io/), but there are others.
This PR adds the ability for Moquette to expose metrics for such system to gather, and provides an implementation for Prometheus specifically.

By default, metrics are not gathered, and the system is only a shim that has minimal impact on performance. The Proteus implementation is a separate project that can be added to the classpath if desired. This minimises the impact for users that do not require metrics and does not change any dependencies for those use-cases.

- [x] I have made corresponding changes to the documentation
- [x] ~~I have made corresponding change to the default configuration files (and/or docker env variables)~~

1. Start moquette with `metrics_provider_class=MetricsProviderPrometheus` 
1. navigate to http://localhost:9400/metrics


andsel:
Hi @hylkevds thanks for this. It will take some time to me for review, but I'll do, just little bit slow.

broker/src/main/java/io/moquette/broker/SessionEventLoop.java
```diff
@@ -12,25 +13,33 @@ final class SessionEventLoop extends Thread {
 
     private final BlockingQueue<FutureTask<String>> taskQueue;
     private final boolean flushOnExit;
+    private final int queueId;
+    /**
+     * Allows a task to fetch the id of the session queue that is executing it.
+     */
+    private static final ThreadLocal<Integer> threadQueueId = new ThreadLocal<>();
```
andsel, line 22: Why use a thread local when this information is present and can be safely stored and retrieved by a field of the SessionEventLoop instance?

andsel (commented):
Thank's @hylkevds for thus PR.
I've given a first review step, in the client part of the code, where Moquette uses this.
It seems fair, but my big concerns is why using a custom API when we could leverage wide spread used metric APIs and implementations, like for example Dropwizard Metrics which defines client APIs and provide various implementations, like the one for Prometheus?

hylkevds:
> Thank's @hylkevds for thus PR. I've given a first review step, in the client part of the code, where Moquette uses this. It seems fair, but my big concerns is why using a custom API when we could leverage wide spread used metric APIs and implementations, like for example Dropwizard Metrics which defines client APIs and provide various implementations, like the one for Prometheus?

My main concern here is to not add _any_ dependencies on _any_ metrics libraries.
On one hand, there are too many to count and everyone has their favourite. Given that one major use case of Moquette is being embedded in other software (including on Android), whatever metrics system we choose, will be the wrong one for most users.
On the other hand, this will keep moquette as small as possible, with as few dependencies as possible.

So instead of choosing any, I made a minimal shim, with the actual metrics collection being done by a separate plugin that could trivially be written for any metrics system that matches the system already used by whatever Moquette is embedded into.
The default (null) implementation doesn't consume any resources, so won't have a performance impact either.

broker/src/main/java/io/moquette/broker/SessionEventLoop.java
hylkevds: It _is_ a field in the SessionEventLoop instance, but we can't get to it from where we need it. The tasks that the event loop executes don't get the information about which event loop they are being executed on.

Currently it's only being used at one point (tracking the number of messages handled by each event loop), but I expect more will be added in the future.

hylkevds:
Elaborating a bit more about the "custom API" thing: Metrics gathering has to be wired directly into the code. Regardless of which metrics library we choose, we, as authors of moquette, are the ones that define which metrics are gathered as Count, which as Histogram, etc.
The "normal" way to define the metrics would be to spread all the code that is currently in MetricsProviderPrometheus.java all over the Moquette code.
The only thing this Interface does is pull those bits of metrics-gathering code out of the Moquette classes into a separate class.

As far as I can tell, there is not yet a nice "SLF4J-API" equivalent for metrics. Dropwizard Metrics, OpenTelemetry and Proteus are all the equivalent of JUL, Log4J or Logback. They can expose metrics in each others formats, like the logging frameworks can write log files in different formats, but if we choose one, but the embedding application uses another, then there are now two frameworks active, that both expose their own metrics endpoint.

Even if there were a Metrics-API library, it might still be a good idea to make a central metrics-gathering class for moquette, to avoid spreading `if (metrics)` checks and metrics definitions all over Moquette.

andsel:
> My main concern here is to not add any dependencies on any metrics libraries.

~~I generally agree, but some dependencies are needed. Dropwizard Metrics (or Micrometer) are an API that has various implementations, it's more like a Logging API provided by Log4J or SLF4J.

I understand your concern, but Moquette would just depend on the API not the implementation.~~

----

After reading you last comment, could agree with you. I'll move forward in reviewing

broker/src/main/java/io/moquette/broker/SessionEventLoop.java
andsel: In future if we need more information to be shared on the a thread base, without moving around parameters, we could create a sort context.
I tend to prefer to have explicit params in functions to move these contextual information other then something like implicit context like the Thread Local, but maybe is personal taste, I'm not against until we abuse it.

broker/src/main/java/io/moquette/logging/MetricsProvider.java
```diff
@@ -0,0 +1,76 @@
...
+ * http://www.opensource.org/licenses/apache2.0.php
+ *
+ * You may elect to redistribute this code under either of these licenses.
+ */
+package io.moquette.logging;
+
+import io.moquette.broker.config.IConfig;
+
+/**
+ * Interface that a metrics implementation must implement.
```
andsel:
```suggestion
 * Interface that a metrics implementation must implement.
 * It mainly defines methods that are used to track Moquette metrics.
```

broker/src/main/java/io/moquette/logging/MetricsProvider.java
```diff
@@ -0,0 +1,76 @@
...
+ */
+public interface MetricsProvider {
+
+    public void init(IConfig config);
+
+    public void stop();
+
+    /**
+     * Notify the metrics provider about the number and size of session queues. This will
+     * be called once.
```
andsel: If this has to be called one, could we provide an abstract class for implementors that grant the single executions. Something like:
```java
private boolean executed = false;

public void initSessionQueues(int queueCount, int queueSize) {
  if (executed) {
    throw new IllegalStateException("initSessioQueues must be called only once and was already executed");
  }
  ...
}

```

broker/src/main/java/io/moquette/logging/MetricsManager.java
```diff
@@ -0,0 +1,69 @@
...
+ *
+ * The Eclipse Public License is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * The Apache License v2.0 is available at
+ * http://www.opensource.org/licenses/apache2.0.php
+ *
+ * You may elect to redistribute this code under either of these licenses.
+ */
+package io.moquette.logging;
```
andsel: This is not logging but metering, so I would move into package `io.moquette.metrics`

broker/src/main/java/io/moquette/logging/MetricsProvider.java
```diff
@@ -0,0 +1,76 @@
+/*
+ * Copyright (c) 2012-2018 The original author or authors
```
andsel:
```suggestion
 * Copyright (c) 2012-2025 The original author or authors
```

broker/src/main/java/io/moquette/logging/MetricsManager.java
```diff
@@ -0,0 +1,69 @@
+/*
+ * Copyright (c) 2012-2018 The original author or authors
```
andsel:
```suggestion
 * Copyright (c) 2012-2025 The original author or authors
```

broker/src/main/java/io/moquette/logging/MetricsProviderNull.java
```diff
@@ -0,0 +1,75 @@
+/*
+ * Copyright (c) 2012-2018 The original author or authors
```
andsel:
```suggestion
 * Copyright (c) 2012-2025 The original author or authors
```

metrics_prometheus/src/main/java/io/moquette/metrics/prometheus/MetricsProviderPrometheus.java
```diff
@@ -0,0 +1,201 @@
+/*
+ * Copyright (c) 2025 The original author or authors
```
andsel:
```suggestion
 * Copyright (c) 2012-2025 The original author or authors
```

metrics_prometheus/src/test/java/io/moquette/metrics/prometheus/HttpHelper.java
```diff
@@ -0,0 +1,101 @@
+/*
+ * Copyright (c) 2025 The original author or authors
```
andsel:
```suggestion
 * Copyright (c) 2012-2025 The original author or authors
```

metrics_prometheus/src/test/java/io/moquette/metrics/prometheus/MetricsProviderPrometheusTest.java
```diff
@@ -0,0 +1,198 @@
...
+ */
+public class MetricsProviderPrometheusTest {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(MetricsProviderPrometheusTest.class.getName());
+
+    Server m_server;
+    IMqttClient clientListener;
+    IMqttClient clientPublisher;
+    MessageCollector messagesCollector;
+    IConfig m_config;
```
andsel:
```suggestion
    IConfig config;
```

metrics_prometheus/src/test/java/io/moquette/metrics/prometheus/MetricsProviderPrometheusTest.java
```diff
@@ -0,0 +1,198 @@
...
+import org.slf4j.LoggerFactory;
+
+/**
+ * Tests the Prometheus Metrics Provider.
+ */
+public class MetricsProviderPrometheusTest {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(MetricsProviderPrometheusTest.class.getName());
+
+    Server m_server;
```
andsel:
```suggestion
    Server server;
```

I would avoid to use `m_`, it's a heritage from the past.

metrics_prometheus/src/test/resources/log4j.properties
```diff
@@ -0,0 +1,44 @@
+#log4j.rootLogger=ERROR, stdout, file
```
andsel: I know in other parts we use such log4J prop files, but we should keep to the minimum to test this module, to avoid confusion and keep it simple.

metrics_prometheus/src/test/java/io/moquette/metrics/prometheus/MessageCollector.java
```diff
@@ -0,0 +1,109 @@
+/*
+ * Copyright (c) 2012-2016 The original author or authors
```
andsel:
```suggestion
 * Copyright (c) 2012-2025 The original author or authors
```

metrics_prometheus/src/test/java/io/moquette/metrics/prometheus/MetricsProviderPrometheusTest.java
```diff
@@ -0,0 +1,198 @@
+/*
+ * Copyright (c) 2025 The original author or authors
```
andsel:
```suggestion
 * Copyright (c) 2012-2025 The original author or authors
```

metrics_prometheus/src/test/java/io/moquette/metrics/prometheus/MessageCollector.java
```diff
@@ -0,0 +1,109 @@
...
+     */
+    @Override
+    public void deliveryComplete(IMqttDeliveryToken token) {
+//        try {
+//            token.waitForCompletion(1_000);
+//            m_messages.offer(new ReceivedMessage(token.getMessage(), token.getTopics()[0]));
+//        } catch (MqttException e) {
+//            e.printStackTrace();
+//        }
+    }
```
andsel:
```suggestion
    public void deliveryComplete(IMqttDeliveryToken token) {
    }
```

I would avoid to copy commented code.

metrics_prometheus/src/test/java/io/moquette/metrics/prometheus/IntegrationUtils.java
```diff
@@ -0,0 +1,90 @@
+/*
+ * Copyright (c) 2012-2018 The original author or authors
```
andsel:
```suggestion
 * Copyright (c) 2012-2025 The original author or authors
```

broker/src/main/java/io/moquette/logging/MetricsManager.java
```diff
@@ -0,0 +1,69 @@
...
+
+    private static final Logger LOG = LoggerFactory.getLogger(MetricsManager.class);
+
+    /**
+     * To ensure there is always an implementation ready, we initialise with a
+     * null provider, and replace on init.
+     */
+    private static MetricsProvider metricsProvider = new MetricsProviderNull();
+
+    public static void init(IConfig config) {
```
andsel: Is it possible to cover with test this with a test resource file in `test/resources/META-INF/services` ?

broker/src/main/java/io/moquette/logging/MetricsManager.java
```diff
@@ -0,0 +1,69 @@
...
+                break;
+            }
+        }
+        if (usedProvider == null) {
+            LOG.info("No MetricsProvider configured, or no matching found, using NULL provider. Available providers: {}", foundProviders);
+        } else {
+            metricsProvider = usedProvider;
+        }
+        metricsProvider.init(config);
+    }
```
andsel:
```suggestion
   public static void init(IConfig config) {
        ServiceLoader<MetricsProvider> loader = ServiceLoader.load(MetricsProvider.class);
        String classname = config.getProperty(METRICS_PROVIDER_CLASS, "");

        List<MetricsProvider> foundProviders = new ArrayList<>();
        loader.forEach(foundProviders::add);
        
        Optional<MetricsProvider> usedProviderOpt = foundProviders.stream()
            .filter(provider -> providerMatchClassname(provider, classname))
            .findFirst();
        if (usedProviderOpt.isPresent()) {
            MetricsProvider usedProvider = usedProviderOpt.get();
            LOG.info("Using configured MetricsProvider: {}", usedProvider.getClass().getName());
            metricsProvider = usedProvider;
        } else {
            LOG.info("No MetricsProvider configured, or no matching found, using NULL provider. Available providers: {}",
                foundProviders.stream().map(p -> p.getClass().getName()).collect(Collectors.toList()));
        }
        metricsProvider.init(config);
    }

    private static boolean providerMatchClassname(MetricsProvider provider, String classname) {
        return !StringUtil.isNullOrEmpty(classname) && provider.getClass().getName().endsWith(classname);
    }
```


Try to make more explicit the intention of search for a provider with same class name defined in `classname`

andsel (changes requested):
Thanks @hylkevds for all your hard work on this!

I've left some comments and a suggestion to improve readability of the MetricsManager init method

broker/src/main/java/io/moquette/logging/MetricsManager.java
hylkevds: Good question, I'll try.
Though it will be tested when the Prometheus implementation is built and tested.

hylkevds:
Your "init is only called once" comment actually touches on an important detail I wanted your opinion on:
Currently the MetricsManager uses a static reference to a MetricsProvider. This means there can be only one MetricsProvider, even if one were to try to start multiple Moquette instances in the same VM. The tests were also very messy in this regard, since they create all kinds of classes without calling the cleanup functions.

Do you think this is an issue, and if it is, what is your preferred solution? 
In FROST I use the settings object for passing these semi-global objects.

hylkevds:
I rewrote the initialisation, now each instance of Server gets its own MetricsProvider.

andsel:
Good! @hylkevds if you could resolve the conflicts then we are good to merge :-)

hylkevds:
I've also managed to add tests using `test/resources/META-INF/services`, seems to work just fine!
Will now resolve the conflicts.

broker/src/test/java/io/moquette/metrics/MetricsProviderTest.java
```diff
@@ -0,0 +1,192 @@
...
+package io.moquette.metrics;
+
+import io.moquette.broker.config.IConfig;
+import java.util.Arrays;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * A metrics provider used for testing.
+ */
+public class MetricsProviderTest extends AbstractMetricsProvider {
```
andsel: Given that `<Class>Test` is a pattern used to name the test suites, to avoid confusion with this, that's a test fixture class, I would rename it `MetricsProviderMock` or `MetricsProviderDouble`

andsel (changes requested):
Left just a note on the naming, then I think we are done with this PR :-D

broker/src/test/java/io/moquette/metrics/MetricsProviderTest.java
hylkevds: Yes, while naming it my inspiration drew blank. Fixed now!

andsel (approved):
LGTM

andsel:
Thanks a lot @hylkevds for all this hard work! :-D

PR #915, opened 2025-07-27
Changed the remove method to work of a session instead of a clientID

hylkevds:
The client may have started a new session before the old one is purged, therefore the purge must check if the session being removed should actually be removed and is not a new session.

By basing the remove method on the session instead of the client id, it can ensure only the correct session is removed from the pool.

### DIFF
--- a/broker/src/main/java/io/moquette/broker/SessionRegistry.java
+++ b/broker/src/main/java/io/moquette/broker/SessionRegistry.java
@@ -242,9 +242,8 @@ public class SessionRegistry {
     private void removeExpiredSession(ISessionsRepository.SessionData expiredSession) {
         final String expiredAt = expiredSession.expireAt().map(Instant::toString).orElse("UNDEFINED");
         LOG.debug("Removing session {}, expired on {}", expiredSession.clientId(), expiredAt);
-        remove(expiredSession.clientId());
+        remove(pool.get(expiredSession.clientId()));
         sessionsRepository.delete(expiredSession);
-
         subscriptionsDirectory.removeSharedSubscriptionsForClient(expiredSession.clientId());
     }
 
@@ -510,20 +509,20 @@ public class SessionRegistry {
             throw new SessionCorruptedException("Session has already changed state: " + session);
         }
 
-        unsubscribe(session);
-        remove(session.getClientID());
-
+        remove(session);
+        sessionsRepository.delete(session.getSessionData());
         subscriptionsDirectory.removeSharedSubscriptionsForClient(session.getClientID());
     }
 
-    void remove(String clientID) {
-        final Session old = pool.remove(clientID);
-        if (old != null) {
+    void remove(Session session) {
+        String clientID = session.getClientID();
+        if (pool.remove(clientID, session)) {
             metricsProvider.removeOpenSession();
+            unsubscribe(session);
             // remove from expired tracker if present
             sessionExpirationService.untrack(clientID);
             loopsGroup.routeCommand(clientID, "Clean up removed session", () -> {
-                old.cleanUp();
+                session.cleanUp();
                 return null;
             });
         }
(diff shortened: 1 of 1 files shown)


andsel:
@hylkevds do you know when this happen in production? For example is this a valid case that manifest the problem?
1. `client_1` is connected with name `session_1`
2. another instance of a client, name it `client_2` connects with same name `session_1`
3. `client_1` is expected to disconnect and its session should be removed without interfering with new session object created for `client_2`.


If that's the case, is it possible to write an integration test to cover it?

hylkevds:
> @hylkevds do you know when this happen in production? For example is this a valid case that manifest the problem?
>     1. `client_1` is connected with name `session_1`
>     2. another instance of a client, name it `client_2` connects with same name `session_1`
>     3. `client_1` is expected to disconnect and its session should be removed without interfering with new session object created for `client_2`.

That specific case is covered. When client_1 is still connected, when client_2 connects with the same ClientId, client_1 will be disconnected, and depending on the clean_session setting, the old Session will be cleaned up before the new one is created.

Though the call to `sessionsRepository.delete(session.getSessionData());` is currently missing here!
Looking through it again, I think this was the main reason I changed it. As `sessionsRepository.delete(session.getSessionData());` requires a session, and we want to avoid multiple calls to pool.get(clientId).

The tricky issue would be when `client_1` makes a session with cleanSession=false (or with a timeout in MQTT5) and the timed cleanup happens right as `client_2` connects, and a race-condition makes the session cleaner execute after the new session was made.

The correct solution would be to make the cleanup happen on the SessionLoop, and add a simple check to see if the session is actually connected. A connected session should never be removed.

andsel (commented):
I think that if it's a race condition then has to be managed with proper synchronization strategy.

hylkevds:
Yes, hence why we should move the cleanup to the SessionLoop. Though the currently missing `sessionsRepository.delete(session.getSessionData());` is the bigger issue.
I'll see what I can come up with.

hylkevds:
I've had another look and made two additional changes:
- the session-timeout purge now happens on the SessionLoop, to make sure the purge can not happen at the same time (on a different thread) as a session re-open.
- I added an extra check on the session state, in case the cleanup get scheduled after a re-open. This is important, since the cleanup also removes all subscriptions.

Regarding tests, unfortunately, race conditions are nearly impossible to reliably test, since they depend on the exact interleaving of threads...

broker/src/main/java/io/moquette/broker/SessionRegistry.java
```diff
@@ -241,11 +241,18 @@ public SessionCreationResult(Session session, CreationModeEnum mode, boolean alr
...
-        sessionsRepository.delete(expiredSession);
-
-        subscriptionsDirectory.removeSharedSubscriptionsForClient(expiredSession.clientId());
+        final String clientId = expiredSession.clientId();
+        loopsGroup.routeCommand(clientId, "PurgeSession", () -> {
+            LOG.debug("Removing session {}, expired on {}", clientId, expiredAt);
+            final Session session = pool.get(clientId);
+            boolean success = session.assignState(SessionStatus.DISCONNECTED, SessionStatus.DESTROYED);
+            if (!success) {
+                remove(session);
```
andsel, line 250: I think that if the session status isn't anymore `DISCONNECTED` while it's going to be removed because expired, it means it was already cleaned by some  other client that connected with same clientId. This happened between the enqueuing of the `PurgeSession` command and the command execution. 
However, this confuse me, the PurgeSession is operating on a session that already changed its state from what it expect (so it has loose control on it, because of racing with some other thread) and does a destructive operation: remove + delete from session repo.

andsel (changes requested):
Hi @hylkevds , sorry for this long review cycle it fallen back in the TODOs.
Left a doubt on the implementation.

PR #917, opened 2025-08-10
[WIP] Switch SLF4J backend to log4j2

andsel:
## Checklist

- [ ] My code follows the style guidelines of this project
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have updated the Changelog if it's a feature or a fix that has to be reported

## Author's Checklist

- [ ] Move the configuration to xml from properties, but accept both formats.

## Related issues

-

### DIFF
--- a/broker/pom.xml
+++ b/broker/pom.xml
@@ -32,10 +32,23 @@
         </dependency>
 
         <dependency>
-            <!-- This is used to use log4j12 implementation -->
-            <groupId>org.slf4j</groupId>
-            <artifactId>slf4j-reload4j</artifactId>
-            <version>${slf4j.version}</version>
+            <groupId>org.apache.logging.log4j</groupId>
+            <artifactId>log4j-api</artifactId>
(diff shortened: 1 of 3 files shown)


hylkevds:
Interesting, do you have a specific reason for switching?

andsel:
> do you have a specific reason for switching?

Log4j 1.x is more than outdated, also if use reload4j. However with log4j2 we could employ MDC to simplify the moving of some log context information plus we could imagine to implement a log file per session with the routing filter.

hylkevds:
Ideally, the broker would only have a dependency on a logging api, not an implementation.
Currently, when embedding, a maven-exclude has to be added to get rid of the logging implementation if it doesn't fit with the implementation of the embedding application.

We could make the broker only depend on the API (either SLF4J-API or log4j-api) and have a separate maven subproject that makes moquette a runable application by adding the full logging implementation...

andsel:
This is a good point!

PR #919, opened 2025-08-25
Extended test for access rights on shared subscriptions

hylkevds:
Looking over the subscription-adding code I noticed that for shared subscriptions the client received a rejection message, when they are not allowed to read the topic, but the subscription is still added to the topic tree regardless!
This test extension seems to confirm my suspicion.
This is probably a security vulnerability... @andsel you may want to enable [private vulnerability reporting](https://docs.github.com/en/code-security/security-advisories/working-with-repository-security-advisories/configuring-private-vulnerability-reporting-for-a-repository) on the repository.

### DIFF
--- a/broker/src/test/java/io/moquette/integration/mqtt5/SharedSubscriptionTest.java
+++ b/broker/src/test/java/io/moquette/integration/mqtt5/SharedSubscriptionTest.java
@@ -70,7 +70,9 @@ public class SharedSubscriptionTest extends AbstractSubscriptionIntegrationTest 
         // stop already started broker instance
         stopServer();
 
-        final IAuthorizatorPolicy policy = new DeclarativeAuthorizatorPolicy.Builder().build();
+        final IAuthorizatorPolicy policy = new DeclarativeAuthorizatorPolicy.Builder()
+                .writeTo(Topic.asTopic("measures/temp"), null, "publisher")
+                .build();
         startServer(dbPath, policy);
 
         // Connect the client to newly started broker
@@ -85,6 +87,14 @@ public class SharedSubscriptionTest extends AbstractSubscriptionIntegrationTest 
             "Granted qos list must be the same cardinality of the subscribe request");
         assertEquals(MqttQoS.FAILURE.value(), grantedQoSes.iterator().next(),
             "Not readable topic should reflect also in shared subscription");
+        
+        // Test if the client does not reveive data on denied Topic.
+        Mqtt5BlockingClient publisherClient = createPublisherClient();
+        publish(publisherClient, "measures/temp", MqttQos.AT_MOST_ONCE);
+
+        MqttMessage received = lowLevelClient.receiveNextMessage(Duration.ofSeconds(1));
+        assertNull(received, "Client should not have reveived a message on rejected shared topic");
+
     }
 
 
--- a/broker/src/test/java/io/moquette/integration/mqtt5/SharedSubscriptionTest.java
+++ b/broker/src/test/java/io/moquette/integration/mqtt5/SharedSubscriptionTest.java
@@ -66,7 +66,7 @@ public class SharedSubscriptionTest extends AbstractSubscriptionIntegrationTest 
     }
 
     @Test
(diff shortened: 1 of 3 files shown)


hylkevds:
The second commit is the minimal change required to fix the issue.

It moves the access-check forward in the `subscribeClientToTopics` method, and changes the sharedSubscription loop to go over the `ackTopics` list instead of the original message, with a QoS check like in the non-shared subscription loop.

andsel:
Thank's @hylkevds I'll review it

broker/src/test/java/io/moquette/integration/mqtt5/SharedSubscriptionTest.java
```diff
@@ -85,6 +87,14 @@ public void givenATopicNotReadableWhenAClientSubscribeSharedThenReceiveSubackWit
             "Granted qos list must be the same cardinality of the subscribe request");
         assertEquals(MqttQoS.FAILURE.value(), grantedQoSes.iterator().next(),
             "Not readable topic should reflect also in shared subscription");
+        
+        // Test if the client does not reveive data on denied Topic.
+        Mqtt5BlockingClient publisherClient = createPublisherClient();
+        publish(publisherClient, "measures/temp", MqttQos.AT_MOST_ONCE);
+
+        MqttMessage received = lowLevelClient.receiveNextMessage(Duration.ofSeconds(1));
+        assertNull(received, "Client should not have reveived a message on rejected shared topic");
```
andsel:
```suggestion
        assertNull(received, "Client should not have received a message on rejected shared topic");
```

broker/src/test/java/io/moquette/integration/mqtt5/SharedSubscriptionTest.java
```diff
@@ -85,6 +87,14 @@ public void givenATopicNotReadableWhenAClientSubscribeSharedThenReceiveSubackWit
             "Granted qos list must be the same cardinality of the subscribe request");
         assertEquals(MqttQoS.FAILURE.value(), grantedQoSes.iterator().next(),
             "Not readable topic should reflect also in shared subscription");
+        
+        // Test if the client does not reveive data on denied Topic.
```
andsel:
```suggestion
        // Test if the client does not receive data on denied Topic.
```

andsel (changes requested):
Thank's @hylkevds for your contribution. 

I've left some nitpicks, plus I would like to discuss with you a couple of changes:
1. rename the test method to `givenNonReadableTopicWhenClientSubscribeToTheSharedThenReceiveSubackWithNegativeResponseAndNoMessageIsDispatched`
2. In method `subscribeClientToTopics` rename `ackTopics`to `ackingSubsriptions`, this should increase readability.
3. Extract `sub -> sub.qualityOfService() != FAILURE)` in an static helper method like 

  ```java
     private boolean notFailure(MqttTopicSubscription sub) {
         return sub.qualityOfService() != FAILURE;
     }
  ``` 

4. Extract `sub -> SharedSubscriptionUtils.isSharedSubscription(sub.topicName())` into another helper method like:

   ```java
    private boolean isShared(MqttTopicSubscription sub) {
         return SharedSubscriptionUtils.isSharedSubscription(sub.topicName());
     }
   ```

In this way the code:

```java
sharedSubscriptions = ackTopics.stream()
                .filter(sub -> sub.qualityOfService() != FAILURE)
                .filter(sub -> SharedSubscriptionUtils.isSharedSubscription(sub.topicName()))
                .map(SharedSubscriptionData::fromMqttSubscription)
                .collect(Collectors.toList());
```

should become some more pleasant:

```java
haredSubscriptions = ackTopics.stream()
                .filter(this::notFailure)
                .filter(this::isShared)
                .map(SharedSubscriptionData::fromMqttSubscription)
                .collect(Collectors.toList());
```

Let me know about the last 2 suggestion points, that are just code maniac things :-D

hylkevds:
Sure, no problem. It's your project, you can be as specific about code style as you want :)

I've moved the `isShared` method in the `SharedSubscriptionUtils` class, so they are next to each other. I've made `isNoFailure` method static, since it can be static.

I added the changes as a third commit to the PR.

andsel (approved):
LGTM

andsel:
Thanks @hylkevds for all your contribution to the project! :-)

PR #931, opened 2026-01-31
Subscription handling cleanup

hylkevds:
While working on a subscription related feature I had to carefully review the way subscriptions are passed through Moquette, and did some massive cleanup and simplification.

Subscription information was being converted back and forth between several classes,
resulting in much unneeded object creation. But in the end it would always end up as Subscription.
- This removes all the intermediate Objects, simplifying the code.
- This includes the removal of the SharedSubscription class, since in the end, those are also turned into normal Subscriptions, making the extra class meaningless.
- The logic contained many superfluous if/else checks for subscriptionIdentifiers caused by the wrapping and unwrapping of Subscription data.
- All methods that would take several parameters with subscription information now simply accept a subscription.
- Subscriptions are (already) registered on the Session, so when an unsubscribe is received, the subscription can be fetched from the Session. This allows the subscription-remove functions to be greatly simplified.
- To simplify and speed up fetching the subscription from the Session, it is now stored in a Map instead of a Set.

These changes will greatly improve the maintainability of the code, and probably increase efficiency.

## Checklist
- [x] ~~I have made corresponding changes to the documentation~~
- [x] ~~I have made corresponding change to the default configuration files (and/or docker env variables)~~


hylkevds:
Note that this PR is on top of PR #919, since I discovered that problem at the start of this work.

andsel:
Hi @hylkevds I'll check it, sorry but I'm pretty loaded and this PR is not so trivial.

hylkevds:
No rush, I know how life is. I had wanted to get this done half a year ago, but I also just didn't find the time for it.
You should also prioritize 919, not this one.
This PR is far from trivial, and it doesn't actually fix or add any functionality. It just makes the logic a lot easier to follow, and probably improves performance though I've not measured that.

broker/src/main/java/io/moquette/broker/subscriptions/CTrieSubscriptionDirectory.java
```diff
@@ -54,13 +49,14 @@ public void init(ISubscriptionsRepository subscriptionsRepository) {
 
         for (Subscription subscription : this.subscriptionsRepository.listAllSubscriptions()) {
             LOG.debug("Re-subscribing {}", subscription);
-            ctrie.addToTree(SubscriptionRequest.buildNonShared(subscription));
+            // TODO: This must call all registered InterceptHandler.onSubscribe
```
andsel: @hylkevds why should call the `onSubscribe`?  `onSubscribe` callback is intended to be invoked when a client effectively send a `SUBSCRIBE` to a series of topic filter. This `init` method, just reconstructs the directory from the data provided by the subscription's repository (the storage) so aren't real `SUBSCRIBE` commands.

ChangeLog.txt
```diff
@@ -1,4 +1,5 @@
 Version 0.19-SNAPSHOT
+   [Cleanup] Cleaned up subscription handling.
```
andsel:
```suggestion
   [cleanup] Cleaned up subscription handling. (#931)
```

broker/src/main/java/io/moquette/broker/subscriptions/Subscription.java
```diff
@@ -73,6 +74,19 @@ public Topic getTopicFilter() {
         return topicFilter;
     }
 
+    /**
+     * Returns the original topic string the client sent, including the
+     * $share/sharename/ part for shared subscriptions.
+     *
+     * @return the original topic string.
+     */
+    public String getOriginalTopicWithSharename() {
```
andsel:
```suggestion
    public String getOriginalTopicFilterWithSharename() {
```

broker/src/main/java/io/moquette/broker/subscriptions/ShareName.java
```diff
@@ -55,4 +61,9 @@ public String toString() {
             "shareName='" + shareName + '\'' +
             '}';
     }
+
+    @Override
+    public int compareTo(ShareName o) {
+        return shareName.compareTo(o.shareName);
```
andsel, line 68: If `shareName` is null this is a NPE. 
If we know that it can't ever be `null` we need to use a null check assertion in the constructor, or alternatively if `null` is same as `EMPTY_SHARENAME` replace null `shareName` with `EMPTY_SHARENAME`  in the constructor and remove null checks around.

broker/src/main/java/io/moquette/broker/subscriptions/CTrieSubscriptionDirectory.java
```diff
@@ -107,76 +103,50 @@ private static List<Subscription> selectSubscriptionsWithHigherQoSForEachSession
...
     @Override
-    public void addShared(String clientId, ShareName name, Topic topicFilter, MqttSubscriptionOption option) {
-        SubscriptionRequest shareSubRequest = SubscriptionRequest.buildShared(name, topicFilter, clientId, option);
-        addSharedSubscriptionRequest(shareSubRequest);
+    public void addShared(Subscription sub) {
+        addSharedSubscriptionRequest(sub);
     }
 
-    private void addSharedSubscriptionRequest(SubscriptionRequest shareSubRequest) {
+    private void addSharedSubscriptionRequest(Subscription shareSubRequest) {
```
andsel: `addSharedSubscriptionRequest` -> `addShared`

broker/src/main/java/io/moquette/broker/subscriptions/CTrieSubscriptionDirectory.java
```diff
@@ -107,76 +103,50 @@ private static List<Subscription> selectSubscriptionsWithHigherQoSForEachSession
...
+        subscriptionsRepository.addNewSubscription(sub);
         return notExistingSubscription;
     }
 
     @Override
-    public void addShared(String clientId, ShareName name, Topic topicFilter, MqttSubscriptionOption option) {
-        SubscriptionRequest shareSubRequest = SubscriptionRequest.buildShared(name, topicFilter, clientId, option);
-        addSharedSubscriptionRequest(shareSubRequest);
+    public void addShared(Subscription sub) {
+        addSharedSubscriptionRequest(sub);
```
andsel: This can be inlined?

broker/src/main/java/io/moquette/broker/subscriptions/CTrieSubscriptionDirectory.java
```diff
@@ -107,76 +103,50 @@ private static List<Subscription> selectSubscriptionsWithHigherQoSForEachSession
...
-    @Override
-    public boolean add(String clientId, Topic filter, MqttSubscriptionOption option, SubscriptionIdentifier subscriptionId) {
-        SubscriptionRequest subRequest = SubscriptionRequest.buildNonShared(clientId, filter, option, subscriptionId);
-        return addNonSharedSubscriptionRequest(subRequest);
-    }
-
-    private boolean addNonSharedSubscriptionRequest(SubscriptionRequest subRequest) {
-        boolean notExistingSubscription = ctrie.addToTree(subRequest);
-        subscriptionsRepository.addNewSubscription(subRequest.subscription());
+    private boolean addNonSharedSubscriptionRequest(Subscription sub) {
```
andsel: I would remove all the `Request` postfix, given that the argument type changed, maybe didn't have much meaning also when they were created. So propose to change `addNonSharedSubscriptionRequest` to `addNonShared` (the argument already speaks of what we are adding.

broker/src/main/java/io/moquette/broker/PostOffice.java
```diff
@@ -400,49 +379,52 @@ public void subscribeClientToTopics(MqttSubscribeMessage msg, String clientID, S
         ackingSubsriptions = updateWithMaximumSupportedQoS(ackingSubsriptions);
         MqttSubAckMessage ackMessage = doAckMessageFromValidateFilters(ackingSubsriptions, messageID);
 
-        final List<SharedSubscriptionData> sharedSubscriptions;
         final Optional<SubscriptionIdentifier> subscriptionIdOpt;
 
         if (mqttConnection.isProtocolVersion5()) {
-            sharedSubscriptions = ackingSubsriptions.stream()
+            try {
```
andsel: Is this moved here for "fail fast" path ?

broker/src/main/java/io/moquette/broker/PostOffice.java
```diff
@@ -400,49 +379,52 @@ public void subscribeClientToTopics(MqttSubscribeMessage msg, String clientID, S
...
+            List<Subscription> sharedSubscriptions = ackingSubsriptions.stream()
                 .filter(PostOffice::isNoFailure)
                 .filter(SharedSubscriptionUtils::isSharedSubscription)
-                .map(SharedSubscriptionData::fromMqttSubscription)
+                .map((s) -> new Subscription(
+                            clientID,
+                            Topic.asTopic(SharedSubscriptionUtils.extractFilterFromShared(s.topicFilter())),
+                            s.option(),
+                            new ShareName(SharedSubscriptionUtils.extractShareName(s.topicFilter())), subscriptionIdOpt))
                 .collect(Collectors.toList());
```
andsel: This can be a construction static method, to it could become:
```suggestion
                .map(PostOffice::buildSharedFrom(s, clientID, subscriptionIdOpt))
.collect(Collectors.toList());
```

```java
private static Subscription buildSharedFrom(Subscription s, String clientID, Optional<SubscriptionIdentifier> subscriptionIdOpt) {
    return new Subscription(clientID,
                            Topic.asTopic(SharedSubscriptionUtils.extractFilterFromShared(s.topicFilter())),
                            s.option(),
                            new ShareName(SharedSubscriptionUtils.extractShareName(s.topicFilter())), 
                            subscriptionIdOpt);
}   
```

broker/src/main/java/io/moquette/broker/Session.java
```diff
@@ -142,16 +142,27 @@ public String getClientID() {
...
+    }
+
+    public void addSubscription(Subscription sub) {
+        String topic = sub.getOriginalTopicWithSharename();
+        subscriptions.put(topic, sub);
     }
 
     public void addSubscriptions(List<Subscription> newSubscriptions) {
-        subscriptions.addAll(newSubscriptions);
+        for (Subscription sub:newSubscriptions) {
```
andsel:
```suggestion
        for (Subscription sub : newSubscriptions) {
```

broker/src/main/java/io/moquette/broker/subscriptions/CTrieSubscriptionDirectory.java
```diff
@@ -54,13 +49,14 @@ public void init(ISubscriptionsRepository subscriptionsRepository) {
...
         }
 
-        for (SharedSubscription shared : subscriptionsRepository.listAllSharedSubscription()) {
+        for (Subscription shared : subscriptionsRepository.listAllSharedSubscription()) {
             LOG.debug("Re-subscribing shared {}", shared);
-            ctrie.addToTree(SubscriptionRequest.buildShared(shared.getShareName(), shared.topicFilter(),
-                shared.clientId(), MqttSubscriptionOption.onlyFromQos(shared.requestedQoS())));
+            // TODO: This must call all registered InterceptHandler.onSubscribe
+            ctrie.addToTree(shared);
         }
```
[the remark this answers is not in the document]
andsel: @hylkevds I think this observation is right. I think that we can address in a follow-up PR, because requires also unit test to verify the error, and this PR is already big and maybe this change if out of context with the PR.

broker/src/main/java/io/moquette/broker/PostOffice.java
```diff
@@ -400,49 +379,52 @@ public void subscribeClientToTopics(MqttSubscribeMessage msg, String clientID, S
...
 
-            Optional<SharedSubscriptionData> invalidSharedSubscription = sharedSubscriptions.stream()
-                .filter(subData -> !SharedSubscriptionUtils.validateShareName(subData.name.toString()))
+            Optional<Subscription> invalidSharedSubscription = sharedSubscriptions.stream()
+                .filter(subData -> !SharedSubscriptionUtils.validateShareName(subData.getShareName().toString()))
```
andsel: `subData` now can became simply `sub`

broker/src/main/java/io/moquette/broker/Session.java
```diff
@@ -142,16 +142,27 @@ public String getClientID() {
         return data.clientId();
     }
 
-    public List<Subscription> getSubscriptions() {
-        return new ArrayList<>(subscriptions);
+    public Collection<Subscription> getSubscriptions() {
+        return subscriptions.values();
```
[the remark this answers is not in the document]
andsel: @hylkevds I think it's right

broker/src/main/java/io/moquette/broker/SessionRegistry.java
```diff
@@ -356,7 +356,7 @@ private void reactivateSubscriptions(Session session, String username) {
             final boolean topicReadable = authorizator.canRead(existingSub.getTopicFilter(), username,
                 session.getClientID());
             if (!topicReadable) {
-                subscriptionsDirectory.removeSubscription(existingSub.getTopicFilter(), session.getClientID());
+                subscriptionsDirectory.removeSubscription(existingSub);
             }
```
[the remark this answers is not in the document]
andsel: @hylkevds I think it's right, please could you check that?

broker/src/main/java/io/moquette/broker/SessionRegistry.java
```diff
@@ -365,7 +365,7 @@ private void reactivateSubscriptions(Session session, String username) {
 
     private void unsubscribe(Session session) {
         for (Subscription existingSub : session.getSubscriptions()) {
-            subscriptionsDirectory.removeSubscription(existingSub.getTopicFilter(), session.getClientID());
+            subscriptionsDirectory.removeSubscription(existingSub);
         }
```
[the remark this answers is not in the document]
andsel: @hylkevds Also here I think it's right. I recall that I created pretty big unit tests, but maybe this is more an integration one that's missed. WDYT ?

andsel (changes requested):
Hi @hylkevds thank's a TON for your monumental work here, the spirit of the original code was to make think smoother but at the end created a little bit of mess that you correctly fixed.

Now that we don't have anymore a specialized `SharedSubscription` but only the `Subscription` which can be shared or not, I think that in every method that is something like "*sharedSubscription*(Subscription)" we have to put a defensive check that effectively the provided `Subscription` parameter is of shared kind. This to avoid of potential leak of non shared Subscription in the shared Subscription paths. WDYT?

I left suggestion comments and also spotted that Copilot spotted maybe real problems, please could you check those?

I've create #934 as follow-up work on this, to avoid making the context of this PR too big. Feel free to add even more tasks to that issue, maybe all the other Copilot suggestions if you think that fixing in this PR is out-of-context or make it too big.


Great work 👏

broker/src/main/java/io/moquette/broker/Session.java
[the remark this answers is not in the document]
hylkevds: Indeed. The performance cost of making the copy are mostly irrelevant, since the method is only used when re-initialising a server after a restart, or when cleaning up a session.
I'll also change the method header back to returning a List.

broker/src/main/java/io/moquette/broker/PostOffice.java
```diff
@@ -400,49 +379,52 @@ public void subscribeClientToTopics(MqttSubscribeMessage msg, String clientID, S
...
                 .findFirst();
```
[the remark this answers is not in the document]
hylkevds: I think the actual correct fix would be to make sure an empty share name can never happen. En empty share name is bound to cause other issues too.
We should track this as a separate issue, since it already exists in the current version and is not specific to the PR.

broker/src/main/java/io/moquette/broker/subscriptions/CTrieSubscriptionDirectory.java
hylkevds: Because an embedded server should also be brought back up to the state of all subscriptions. For instance, FROST-Server only sends messages on topics that actually have subscriptions. But to know a subscription exists, it registers an InterceptHandler so it knows about subscribe and unsubscribe events.

You could argue that the embedding server should save and restore its own state, but that just leads to the double storing of data, with the significant chance of these two states diverging due to race-conditions during shutdown. It's much more reliable to have Moquette holding the master storage, and the embedding server following that.

An alternative would be to have a separate method on the InterceptHandler for subscriptions restored from storage.

broker/src/main/java/io/moquette/broker/SessionRegistry.java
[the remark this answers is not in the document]
hylkevds: It's correct that there is a problem. There already was a problem, and now it slightly changed :)
I think the old case never handled shared subscription at all, when re-activating a subscription.
The `session.removeSubscription(existingSub);` was already missing in the original version...

Maybe we should simplify the interface of ISubscriptionsDirectory, and have that isShared() check there?

broker/src/main/java/io/moquette/broker/SessionRegistry.java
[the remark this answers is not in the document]
hylkevds: This `unsubscribe(Session)` method is only used when purging a session. So we'd need an integration tests that
- Creates a shared subscription with a persistent session.
- Reconnects with the same clientId, but requesting a clean session.

That test currently doesn't exist... And I'm pretty sure the problem already existed before as a result. I'll add the `hasShareName` check.

broker/src/main/java/io/moquette/broker/subscriptions/ShareName.java
hylkevds: I just noticed validateShareName already does the null and empty check. But I'll add a non-null check regardless, to ensure we can't accidentally add a null sharename in the future.

broker/src/main/java/io/moquette/broker/subscriptions/CTrieSubscriptionDirectory.java
hylkevds: Yep, totally. Will do

broker/src/main/java/io/moquette/broker/PostOffice.java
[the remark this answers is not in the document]
hylkevds: On second look, `ShareName#toString` is indeed the wrong method. Should be `ShareName#getShareName`. That is a bit confusing indeed.
Fixed now.

broker/src/main/java/io/moquette/broker/subscriptions/CTrieSubscriptionDirectory.java
andsel: Maybe Moquette could expose an API to query the subscription storage, so that FROST-Server queries Moquette's subscription storage. In that case I should ask your help to shape the possible queries, the most obvious is "list all subscriptions", but maybe you have other use cases. Then let the asynch listener mechanism to receive the updates when happens, instead of polling this API to aligns continuously its internal state.
WDYT?

hylkevds: Yes, some dedicated methods to inspect the current topics would be even better. It would also allow for occasionally checking if the state of the embedding server still matches that of the broker.
I'll add it to our subscription TODO issue (#934) and create a wishlist for discussion.

andsel:
@hylkevds let me know when this PR is ready for another review :-)

hylkevds:
I think I've addressed all comments, except for removing the TODO comment on lime 57 of [CTrieSubscriptionDirectory.java](https://github.com/moquette-io/moquette/pull/931/files/ee0cf5aa456c8a89ae358307d1abb1974c48d990#diff-ecc72b469fa877814f4c90b5e5d683132edb3a5db2167bfc2c8224479a228cfb)

hylkevds:
The TODO comments are now also gone

broker/src/main/java/io/moquette/broker/subscriptions/CTrieSubscriptionDirectory.java
```diff
@@ -107,76 +101,54 @@ private static List<Subscription> selectSubscriptionsWithHigherQoSForEachSession
...
-    public boolean add(String clientId, Topic filter, MqttSubscriptionOption option, SubscriptionIdentifier subscriptionId) {
-        SubscriptionRequest subRequest = SubscriptionRequest.buildNonShared(clientId, filter, option, subscriptionId);
-        return addNonSharedSubscriptionRequest(subRequest);
-    }
-
-    private boolean addNonSharedSubscriptionRequest(SubscriptionRequest subRequest) {
-        boolean notExistingSubscription = ctrie.addToTree(subRequest);
-        subscriptionsRepository.addNewSubscription(subRequest.subscription());
+    public boolean add(Subscription sub) {
+        if (sub.hasShareName()) {
```
andsel: Would you mind to add checker methods like in https://github.com/andsel/moquette/pull/2 ?

andsel (changes requested):
@hylkevds maybe a nitpick, but please consider to add helper checker methods on subscription type, like assertion methods.

broker/src/main/java/io/moquette/broker/subscriptions/CTrieSubscriptionDirectory.java
hylkevds: Hmm, personally I dislike hiding the throwing of exceptions away like that, especially if the method name doesn't make it very obvious that its only goal is to throw an exception sometimes... It doesn't make the code more readable nor easier to follow.

Of course it's your project, so if you insist I'll change it :)

hylkevds: Dinner and some Yoga later...
I see the point that those `if ... then ... throw new x` are very ugly, but a method should either encapsulate significant complexity, or be re-useable. How about this further development of the pattern:

```java
public class Exceptions {
    public static final void illegalArgumentIf(boolean predicate, String message) {
        if (predicate) {
            throw new IllegalArgumentException(message);
        }
    }
}
```
That is re-usable, and calling 
```java
Exceptions.illegalArgumentIf(sub.hasShareName(), "Adding a shared subscription using the non-shared method.");
```
makes it very clear what the intention is.
It also open the option of making a version that accepts a lambda instead of a boolean, for more complex checks.

We can then also do the same for other exceptions. What do you think?

andsel: My initial consideration came from https://docs.oracle.com/javase/8/docs/api/java/util/Objects.html#requireNonNull-T- where in some methods, to assure a parameter is not null we use:
```java
public void aMethod(Object some) {
  Object.requireNonNull("some parameter can't be null");
  ...
}
```

Your idea is not bad, is a good brick to build other "check-style" method. I think that if the method is more constrained provides implicitly more context.

What do you say of renaming `checkIsShared` with `requireShared(subscription, message)`, it's very domain specific and carries all the meaning it needs.

hylkevds: Yes, `requireShared` is better. 
On that note, shouldn't it just be a method on the subscription itself? Then we don't need to pass the subscription to the method: `subscription.requireShared(message)`.

andsel: That's a nice suggestion, however the invariant is part of the client code, in this code `CTrieSubscriptionDirectory` and it's not proper business for the subscription itself. Just to limit the public API surface of the Subscription and limiting the scope.

hylkevds: Do you want to update your PR to this PR, or shall I add the changes as a separate commit to this PR, or do we do this completely separate?

andsel: You can add the changes as a separate commit to this PR 🙏

hylkevds: Done

andsel (approved):
LGTM

Thank you so so so much @hylkevds for your work on this! 👏

PR #932, opened 2026-03-14
Feature/topic rewriting

hylkevds:
This PR is the reason I made #931, since that PR makes this one a lot simpler :)
There is no rush for a detailed code review on this, but it may provide some context for the other one, and we may need to discuss if this is a feature that fits Moquette in the first place.

## What does this PR do?

Topic rewriting, meaning the server internally changes the topic that the client subscribes to, without the client noticing. The rewriting is done by a class implementing the TopicRewriter interface that can be set when starting the Broker.

The three main use cases are:

- Migrating to a new topic structure: When one wants to change the topic structure, but it's not feasible to update all clients in one go. The publishers can change to the new structure, and old clients have their subscriptions rewritten for the new structure, without them noticing. Of course this assumes the messages themselves don't change.
- Topic normalisation: When the same message is sent to different topics, these can be merged into one topic for the server, reducing load. For instance, in the OGC SensorThings API, a client can choose which details to receive when something changes on the server by adding `$select`. Client subscribing to `v1.1/Things?$select=id,name` would receive the same messages as those subscribing to  `v1.1/Things?$select=name,id`, but the embedding server needs to publish two messages, one for each topic. With topic rewriting, these topics can be merged into one topic, meaning the embedding server only needs to publish one message.
- Topic de-normalisation: If the topic structure is fixed, but want to add fine-grained access-control, meaning which exact messages a client receives depends on the user(group). Topic rewriting allows the subscriptions to be separated based on the user(group), and the embedding server can publish messages on those de-normalised topics based on this user(group) information.

I'm currently testing it with a [branch of FROST-Server](https://github.com/FraunhoferIOSB/FROST-Server/tree/develop-mqttRights) to see if it works as expected.

It currently only rewrites topic on subscriptions, not on publishes. If there is interest, I could implement that as well, though I currently don't need that.

## Checklist

- ~~[ ] I have made corresponding change to the default configuration files (and/or docker env variables)~~

## Related issues

- Requires #931

### DIFF
--- a/broker/src/main/java/io/moquette/broker/PostOffice.java
+++ b/broker/src/main/java/io/moquette/broker/PostOffice.java
@@ -57,6 +57,8 @@ import java.util.function.Consumer;
 import java.util.stream.Collectors;
 
 import static io.moquette.broker.Utils.messageId;
+import io.moquette.interception.TopicRewriter;
+import io.moquette.interception.TopicRewriterUnity;
 import io.moquette.metrics.MetricsManager;
 import io.moquette.metrics.MetricsProvider;
 import static io.netty.handler.codec.mqtt.MqttMessageIdVariableHeader.from;
@@ -206,6 +208,7 @@ class PostOffice {
     private final ISessionsRepository sessionRepository;
     private SessionRegistry sessionRegistry;
     private BrokerInterceptor interceptor;
+    private TopicRewriter topicRewriter = new TopicRewriterUnity();
     private final FailedPublishCollection failedPublishes = new FailedPublishCollection();
     private final SessionEventLoopGroup sessionLoops;
     private final Clock clock;
@@ -269,6 +272,10 @@ class PostOffice {
 
         this.retainedMessagesExpirationService = new ScheduledExpirationService<>(clock, this::cleanRetainedExpired);
         recreateRetainedExpires(retainedRepository);
+    }
+
+    public void setTopicRewriter(TopicRewriter topicRewriter) {
+        this.topicRewriter = topicRewriter;
     }
 
     private void cleanRetainedExpired(ExpirableTopic expirable) {
@@ -420,7 +427,9 @@ class PostOffice {
             .map(sub -> {
                 final Topic topic = new Topic(sub.topicFilter());
                 MqttSubscriptionOption option = sub.option();//MqttSubscriptionOption.onlyFromQos(sub.qualityOfService());
-                return new Subscription(clientID, topic, option, subscriptionIdOpt);
+                final Subscription subscription = new Subscription(clientID, topic, option, subscriptionIdOpt);
+                subscription.setTopicFilterInternal(topicRewriter.rewriteTopic(subscription));
+                return subscription;
             }).collect(Collectors.toList());
 
         final Set<Subscription> subscriptionToSendRetained = newSubscriptions.stream()
@@ -444,12 +453,14 @@ class PostOffice {
         }
     }
 
-    private static Subscription buildSharedSubscriptionFrom(MqttTopicSubscription s, String clientID, Optional<SubscriptionIdentifier> subscriptionIdOpt) {
-        return new Subscription(
+    private Subscription buildSharedSubscriptionFrom(MqttTopicSubscription s, String clientID, Optional<SubscriptionIdentifier> subscriptionIdOpt) {
+        final Subscription subscription = new Subscription(
             clientID,
             Topic.asTopic(SharedSubscriptionUtils.extractFilterFromShared(s.topicFilter())),
             s.option(),
             new ShareName(SharedSubscriptionUtils.extractShareName(s.topicFilter())), subscriptionIdOpt);
+        subscription.setTopicFilterInternal(topicRewriter.rewriteTopic(subscription));
+        return subscription;
     }
 
     private static boolean needToReceiveRetained(Utils.Couple<Boolean, Subscription> addedAndSub) {
@@ -519,7 +530,7 @@ class PostOffice {
     private void publishRetainedMessagesForSubscriptions(String clientID, Collection<Subscription> newSubscriptions) {
         Session targetSession = this.sessionRegistry.retrieve(clientID);
         for (Subscription subscription : newSubscriptions) {
-            final String topicFilter = subscription.getTopicFilter().toString();
+            final String topicFilter = subscription.getTopicFilterInternal().toString();
             final Collection<RetainedMessage> retainedMsgs = retainedRepository.retainedOnTopic(topicFilter);
 
             if (retainedMsgs.isEmpty()) {
@@ -929,7 +940,16 @@ class PostOffice {
         boolean isSessionPresent = targetSession != null;
         if (isSessionPresent) {
             LOG.debug("Sending PUBLISH message to active subscriber CId: {}, topicFilter: {}, qos: {}",
-                      sub.getClientId(), sub.getTopicFilter(), qos);
+                      sub.getClientId(), sub.getTopicFilterInternal(), qos);
+
+            if (sub.getTopicFilterClient().hasWildcard() && sub.isTopicRewritten()) {
+                // Topic contains a wildcard AND is rewritten. The interceptor that did the rewriting
+                // must tell us what topic the client expects.
+                topic = topicRewriter.rewriteTopicInverse(sub.getTopicFilterClient(), topic);
+            } else {
+                // Non-Wildcard or non-rewritten topic, we can use the client version.
+                topic = sub.getTopicFilterClient();
+            }
 
             metricsProvider.addMessage(SessionEventLoop.getThreadQueueId(), qos.value());
             Collection<? extends MqttProperties.MqttProperty> existingProperties = msg.variableHeader().properties().listAll();
@@ -941,7 +961,7 @@ class PostOffice {
             // If we are, the subscriber disconnected after the subscriptions tree selected that session as a
             // destination.
             LOG.debug("PUBLISH to not yet present session. CId: {}, topicFilter: {}, qos: {}", sub.getClientId(),
-                      sub.getTopicFilter(), qos);
+                      sub.getTopicFilterInternal(), qos);
         }
     }
 
--- a/broker/src/main/java/io/moquette/broker/PostOffice.java
+++ b/broker/src/main/java/io/moquette/broker/PostOffice.java
@@ -428,8 +428,7 @@ class PostOffice {
                 final Topic topic = new Topic(sub.topicFilter());
                 MqttSubscriptionOption option = sub.option();//MqttSubscriptionOption.onlyFromQos(sub.qualityOfService());
                 final Subscription tempSub = new Subscription(clientID, topic, option, subscriptionIdOpt);
-                final Subscription subscription = tempSub.withInternal(topicRewriter.rewriteTopic(tempSub));
-                return subscription;
+                return tempSub.withRewrittenTopic(topicRewriter.rewriteTopic(tempSub));
             }).collect(Collectors.toList());
 
         final Set<Subscription> subscriptionToSendRetained = newSubscriptions.stream()
@@ -459,8 +458,7 @@ class PostOffice {
             Topic.asTopic(SharedSubscriptionUtils.extractFilterFromShared(s.topicFilter())),
             s.option(),
             new ShareName(SharedSubscriptionUtils.extractShareName(s.topicFilter())), subscriptionIdOpt);
-        final Subscription subscription = tempSub.withInternal(topicRewriter.rewriteTopic(tempSub));
-        return subscription;
+        return tempSub.withRewrittenTopic(topicRewriter.rewriteTopic(tempSub));
     }
 
     private static boolean needToReceiveRetained(Utils.Couple<Boolean, Subscription> addedAndSub) {
@@ -530,7 +528,7 @@ class PostOffice {
     private void publishRetainedMessagesForSubscriptions(String clientID, Collection<Subscription> newSubscriptions) {
         Session targetSession = this.sessionRegistry.retrieve(clientID);
         for (Subscription subscription : newSubscriptions) {
-            final String topicFilter = subscription.getTopicFilterInternal().toString();
+            final String topicFilter = subscription.getTopicFilterRewritten().toString();
             final Collection<RetainedMessage> retainedMsgs = retainedRepository.retainedOnTopic(topicFilter);
 
             if (retainedMsgs.isEmpty()) {
@@ -940,7 +938,7 @@ class PostOffice {
         boolean isSessionPresent = targetSession != null;
         if (isSessionPresent) {
             LOG.debug("Sending PUBLISH message to active subscriber CId: {}, topicFilter: {}, qos: {}",
-                      sub.getClientId(), sub.getTopicFilterInternal(), qos);
+                      sub.getClientId(), sub.getTopicFilterRewritten(), qos);
 
             if (sub.getTopicFilterClient().hasWildcard() && sub.isTopicRewritten()) {
                 // Topic contains a wildcard AND is rewritten. The interceptor that did the rewriting
@@ -961,7 +959,7 @@ class PostOffice {
             // If we are, the subscriber disconnected after the subscriptions tree selected that session as a
             // destination.
             LOG.debug("PUBLISH to not yet present session. CId: {}, topicFilter: {}, qos: {}", sub.getClientId(),
-                      sub.getTopicFilterInternal(), qos);
+                      sub.getTopicFilterRewritten(), qos);
         }
     }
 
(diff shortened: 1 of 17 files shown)


broker/src/test/java/io/moquette/integration/ServerIntegrationTopicRewriteTest.java
```diff
@@ -0,0 +1,205 @@
...
+        server.stopServer();
+    }
+
+    /**
+     * On the server the sensor sub-tree was moved to sensors, but legacy clients should still receive messages on the old topic.
+     * Subscriber B subscribing to sensor/sensor1 should get messages sent to sensors/sensor1, as if they were sent to sensor/sensor1.
+     * Subscriber C subscribing to sensor/+ should get messages sent to sensors/+, as if they were sent to sensor/+.
+     */
+    @Test
+    public void checkSubscribersGetCorrectTopicNotifications() throws Exception {
```
andsel: I think that the test is actually testing many things:
1. `subscriberA` which subscribes to the rewritten topic get notified.
2. `subscriberB` which subscribes to the existing topic, continues to receives notifications also if the topic is rewritten.
3. `subscriberC` which subscribes with wildcards to the original topic continues to receive notification also after the topic is rewritten.
4. after a client subscribed to the original topic is unsubscribed from that, then doesn't receive any notification.
5. after a client subscribed to the rewritten topic is unsubscribed from that, then doesn't receive any notification.

I would like the test to be simple, so that it's better to split once for each case.

broker/src/test/java/io/moquette/integration/ServerIntegrationTopicRewriteTest.java
```diff
@@ -0,0 +1,205 @@
...
+
+        {
+            Awaitility.await().until(cbSubscriberA::isMessageReceived);
+            MqttMessage message = cbSubscriberA.retrieveMessage();
+            assertEquals(messageText2, new String(message.getPayload(), UTF_8));
+            assertEquals(1, message.getQos());
+            subscriberA.disconnect();
+        }
+        {
+            Thread.sleep(Durations.ONE_SECOND.toMillis());
```
andsel:
```suggestion
            Awaitility.await().atMost(Durations.ONE_SECOND).until(() -> !cbSubscriberB.isMessageReceived());
```

I think that also this wait can be rewritten with Awaitility.

andsel (changes requested):
Hi @hylkevds I think this idea of topic rewriting on the subscription/read side, is interesting. 
There are some things that I think we have to fix, but as first iteration let's start with the test code. Then after this, I'll explain some my concerns on the core part.
Thanks for proposing this! 👍

broker/src/test/java/io/moquette/integration/ServerIntegrationTopicRewriteTest.java
hylkevds: Awaitility now has something for that:
```
        Awaitility.await()
            .during(Durations.ONE_SECOND)
            .atMost(2, TimeUnit.SECONDS)
            .until(() -> !cbSubscriberC.isMessageReceived());
```

Note that the `atMost` must be significantly larger than the `during`, otherwise it still fails, for some reason. Probably because the polling isn't perfect, and the last poll may be scheduled to happen after the atMost kills the thread. So I guess the atMost must be at least about two poll-intervals longer than the during.

broker/src/test/java/io/moquette/integration/ServerIntegrationTopicRewriteTest.java
hylkevds: Indeed, this test is trivial to split into three, so I've done that now.

My main rationale for merging them is that the split test will take three times as long to execute, and most of that time is spent waiting... merging them makes that waiting happen in parallel.

andsel: Every test should test one use case, there is no reason to mess up different cases in the same.

broker/src/test/java/io/moquette/integration/ServerIntegrationTopicRewriteTest.java
```diff
@@ -0,0 +1,229 @@
...
+
+        {
+            Awaitility.await().until(cbSubscriberA::isMessageReceived);
+            MqttMessage message = cbSubscriberA.retrieveMessage();
+            assertNotNull(message, "MUST be a received message");
+            assertEquals(messageText1, new String(message.getPayload(), UTF_8));
+            assertEquals(1, message.getQos());
+        }
+        subscriberA.disconnect();
+    }
```
andsel:
```suggestion
        Awaitility.await().until(cbSubscriberA::isMessageReceived);
        MqttMessage message = cbSubscriberA.retrieveMessage();
        assertNotNull(message, "MUST be a received message");
        assertEquals(messageText1, new String(message.getPayload(), UTF_8));
        assertEquals(1, message.getQos());
        subscriberA.disconnect();
    }
```

I don't think there is good motivation why opening a new code block, except to reuse locally the `message` variable name.

andsel (changes requested):
Left a nitpick on the test.
After test are fixed I'll pass to the code part review.

broker/src/test/java/io/moquette/integration/ServerIntegrationTopicRewriteTest.java
hylkevds: Somehow I missed that code block when splitting the test into three. Fixed now.

broker/src/main/java/io/moquette/interception/TopicRewriter.java
```diff
@@ -0,0 +1,46 @@
...
+     * Reverse the rewrite for the given publish Topic, so that it matches the
+     * clientTopic. This is needed when a subscription that this topic matched
+     * (the clientTopic) has a wild-card and is rewritten.
+     *
+     * @param clientTopic The topic that the client originally subscribed to.
+     * This will have a wild-card in it.
+     * @param publishedTopic The topic that received a publish.
+     * @return The topic that the client would expect the publish to be on.
+     */
+    public Topic rewriteTopicInverse(Topic clientTopic, Topic publishedTopic);
```
andsel:
```suggestion
    Topic rewriteTopicInverse(Topic clientTopic, Topic publishedTopic);
```
No need to redundantly mark interface methods as public.

broker/src/main/java/io/moquette/interception/TopicRewriter.java
```diff
@@ -0,0 +1,46 @@
...
+ */
+public interface TopicRewriter {
+
+    /**
+     * Rewrite the topic for the given subscription.
+     *
+     * @param subscription The subscription to rewrite the Topic for.
+     * @return the rewritten topic.
+     */
+    public Topic rewriteTopic(Subscription subscription);
```
andsel:
```suggestion
    Topic rewriteTopic(Subscription subscription);
```

broker/src/main/java/io/moquette/broker/PostOffice.java
```diff
@@ -420,7 +427,9 @@ public void subscribeClientToTopics(MqttSubscribeMessage msg, String clientID, S
             .map(sub -> {
                 final Topic topic = new Topic(sub.topicFilter());
                 MqttSubscriptionOption option = sub.option();//MqttSubscriptionOption.onlyFromQos(sub.qualityOfService());
-                return new Subscription(clientID, topic, option, subscriptionIdOpt);
+                final Subscription subscription = new Subscription(clientID, topic, option, subscriptionIdOpt);
+                subscription.setTopicFilterInternal(topicRewriter.rewriteTopic(subscription));
+                return subscription;
```
andsel: `Subscription` class is intended to be immutable, so that this becomes:


```suggestion
                final Subscription subscription = new Subscription(clientID, topic, option, subscriptionIdOpt, opicRewriter.rewriteTopic(subscription));
                return subscription;
```

In this case the `topicFilterInternal` becomes final and received as constructor argument.

broker/src/main/java/io/moquette/broker/subscriptions/Subscription.java
```diff
@@ -30,7 +30,8 @@ public final class Subscription implements Serializable, Comparable<Subscription
     private static final long serialVersionUID = -3383457629635732794L;
     private final MqttSubscriptionOption option;
     final String clientId;
-    final Topic topicFilter;
+    final Topic topicFilterClient;
+    private Topic topicFilterInternal;
```
andsel: If this is the rewritten topic, a better named should help understand the meaning. This is the redirected filter if I understand correctly.

The subscriber subscribes to `/old/topic` and continue to subscribe to that.
The user define a redirection from `/new/topic` to `/old/topic`.
The publisher publishes on `/new/topic` and the message is dispatched to both `/new/topic` and this redirected topic `/old/topic`.

So could be named `redirectedTopicFilter`.

WDYT?

andsel (changes requested):
Left a couple  of comments on the immutability of `Subscription` and better naming for this rerouted or redirected prefix to the TopicFilter.

broker/src/main/java/io/moquette/broker/subscriptions/Subscription.java
hylkevds: The topic isn't always redirected, so `redirectedTopicFilter` may not actually be accurate. It may just be the same as `topicFilterClient`. Hence the names *Internal vs *Client, one is the one the client requested, the other is the one we use internally, rewritten or not.

broker/src/main/java/io/moquette/broker/PostOffice.java
hylkevds: Ok, I'll make the field final and make sure only the last updated one is stored.

Currently Subscription has a mix of protected and private fields, what's your opinion on that?

broker/src/main/java/io/moquette/interception/TopicRewriter.java
hylkevds: Fixed.
Hmm, maybe we can use Spotless for style rules like these.

broker/src/main/java/io/moquette/broker/subscriptions/Subscription.java
hylkevds: I've renamed the field, in a separate commit in case we want to change it again :)

broker/src/main/java/io/moquette/interception/TopicRewriter.java
andsel: Great idea, created #960 to track this.

broker/src/main/java/io/moquette/broker/PostOffice.java
andsel: The `protected` are used in same package classes, for example `clientId` is used in `DumpTreeVisitor`, `CTrie` and some tests. We can switch to private and leverage the getter methods if it's functional equivalent (the getters doesn't modify the field value in returning it). My idea is that if it's a final field, it can be accessed directly without getter, because the getter doesn't add any value, but that means that's an implicit interface.

At the moment we can leave as it is, later we could switch to getters and private modifier.

broker/src/main/java/io/moquette/broker/subscriptions/Subscription.java
andsel: What do you think of `topicFIlterRewritten` ?

broker/src/main/java/io/moquette/broker/PostOffice.java
```diff
@@ -420,7 +427,9 @@ public void subscribeClientToTopics(MqttSubscribeMessage msg, String clientID, S
             .map(sub -> {
                 final Topic topic = new Topic(sub.topicFilter());
                 MqttSubscriptionOption option = sub.option();//MqttSubscriptionOption.onlyFromQos(sub.qualityOfService());
-                return new Subscription(clientID, topic, option, subscriptionIdOpt);
+                final Subscription tempSub = new Subscription(clientID, topic, option, subscriptionIdOpt);
+                final Subscription subscription = tempSub.withRedirectedTopic(topicRewriter.rewriteTopic(tempSub));
+                return subscription;
```
andsel:
```suggestion
                return  tempSub.withRedirectedTopic(topicRewriter.rewriteTopic(tempSub));
```

broker/src/main/java/io/moquette/broker/PostOffice.java
```diff
@@ -444,12 +453,14 @@ public void subscribeClientToTopics(MqttSubscribeMessage msg, String clientID, S
...
-    private static Subscription buildSharedSubscriptionFrom(MqttTopicSubscription s, String clientID, Optional<SubscriptionIdentifier> subscriptionIdOpt) {
-        return new Subscription(
+    private Subscription buildSharedSubscriptionFrom(MqttTopicSubscription s, String clientID, Optional<SubscriptionIdentifier> subscriptionIdOpt) {
+        final Subscription tempSub = new Subscription(
             clientID,
             Topic.asTopic(SharedSubscriptionUtils.extractFilterFromShared(s.topicFilter())),
             s.option(),
             new ShareName(SharedSubscriptionUtils.extractShareName(s.topicFilter())), subscriptionIdOpt);
+        final Subscription subscription = tempSub.withRedirectedTopic(topicRewriter.rewriteTopic(tempSub));
+        return subscription;
```
andsel:
```suggestion
        return tempSub.withRedirectedTopic(topicRewriter.rewriteTopic(tempSub));
```

andsel (changes requested):
Thank's @hylkevds for addressing all my concerns.
I've left a couple of suggestion on inlining a couple of code parts and suggested (another) new name for topicFilterInternal, let me know your thoughts, and if you agree with the new name proposed go ahead and replace all "Internal" references with the new one.

hylkevds:
> Thank's @hylkevds for addressing all my concerns. I've left a couple of suggestion on inlining a couple of code parts and suggested (another) new name for topicFilterInternal, let me know your thoughts, and if you agree with the new name proposed go ahead and replace all "Internal" references with the new one.

I've updated the names. topicRewritten sounds like a fitting name indeed :)
I've also updated the changeLog with the API changes, and fixed the assign-and-return spots.

broker/src/main/java/io/moquette/interception/messages/InterceptSubscribeMessage.java
```diff
@@ -42,8 +42,12 @@ public MqttSubscriptionOption getOption() {
         return subscription.getOption();
     }
 
-    public String getTopicFilter() {
-        return subscription.getTopicFilter().toString();
+    public String getTopicFilterClient() {
+        return subscription.getTopicFilterClient().toString();
+    }
+
+    public String getTopicFilterInternal() {
```
andsel:
```suggestion
    public String getTopicFilterRewritten() {
```

andsel (changes requested):
Hi @hylkevds the PR is really close to the merge, left a comment on a missed rename.

broker/src/main/java/io/moquette/broker/PostOffice.java
```diff
@@ -929,7 +938,16 @@ private void publishToSession(ByteBuf payload, Topic topic, Subscription sub, Mq
...
+                      sub.getClientId(), sub.getTopicFilterRewritten(), qos);
+
+            if (sub.getTopicFilterClient().hasWildcard() && sub.isTopicRewritten()) {
+                // Topic contains a wildcard AND is rewritten. The interceptor that did the rewriting
+                // must tell us what topic the client expects.
+                topic = topicRewriter.rewriteTopicInverse(sub.getTopicFilterClient(), topic);
+            } else {
+                // Non-Wildcard or non-rewritten topic, we can use the client version.
+                topic = sub.getTopicFilterClient();
+            }
```
andsel, line 951: Hi @hylkevds could you help me understand better this piece, please.

This is related to the fact that `topicName` in publish message sent to a subscriber must contain a topic name that matches the topicFilter of the subscription.

## Topic filter without wildcards

Subscribes subscribes to `old/sensor1`, which the rewriter rewrites to `new/sensor1`. This means that on the subscriptions tree we register the subscription under the `new/sensor1` path. 
When a publish matching `new/sensor1` is received a new publish message is emitted with a topic name of the client, so `old/sensor1`.


## Topic filter with wildcards

Subscribes subscribes to `old/#`, which the rewriter rewrites to `new/#`. This means that on the subscriptions tree we register the subscription under the `new/#` path.
When a publish matching `new/#` is received, for example `new/sensor1`, a new publish message is emitted with a topic name as it matched the client's topic filter, so `old/sensor1`. In this case we can't user `topicFilterClient` because it would be `old/#` and we need a method in the rewriter that convert `new/sensor1` to `old/sensor1`.


Am I right?

hylkevds: Yes, that's exactly correct.
One main thing to do is better documentation, like adding what you just typed to the TopicRewriter interface.

hylkevds: I've added your description to the JavaDoc of TopicRewriter.

andsel (approved):
Thanks @hylkevds for all your effort on this :-) Good job!
LGTM

PR #940, opened 2026-05-15
Fixed metrics session mis-counting

hylkevds:
## What does this PR do?
Fix a session-accounting-mistake in the session-reopen code, causing the Metrics to miscount sessions. I was seeing negative session counts :)

Since the metrics code is not released yet, no need to update the changelog.

- [x] I have made corresponding changes to the documentation
- [x] I have made corresponding change to the default configuration files (and/or docker env variables)

### DIFF
--- a/broker/src/test/java/io/moquette/metrics/MetricsTest.java
+++ b/broker/src/test/java/io/moquette/metrics/MetricsTest.java
@@ -25,15 +25,17 @@ import io.netty.handler.codec.mqtt.MqttQoS;
 import java.io.IOException;
 import static java.nio.charset.StandardCharsets.UTF_8;
 import java.nio.file.Path;
+import java.time.Duration;
 import java.util.Properties;
-import java.util.logging.Level;
 import org.awaitility.Awaitility;
 import org.awaitility.Durations;
 import org.eclipse.paho.client.mqttv3.IMqttClient;
 import org.eclipse.paho.client.mqttv3.MqttClient;
 import org.eclipse.paho.client.mqttv3.MqttClientPersistence;
+import org.eclipse.paho.client.mqttv3.MqttConnectOptions;
 import org.eclipse.paho.client.mqttv3.MqttException;
 import org.eclipse.paho.client.mqttv3.persist.MqttDefaultFilePersistence;
+import org.hamcrest.core.Is;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.Assertions;
 import static org.junit.jupiter.api.Assertions.assertEquals;
@@ -107,15 +109,17 @@ public class MetricsTest {
         server.stopServer();
     }
 
+    private MetricsProviderMock validateMetricsProvider() {
+        // Setup, check if the mock metrics provider is used and is clean.
+        MetricsProvider metricsProvider = server.getMetricsProvider();
+        assertTrue(metricsProvider instanceof MetricsProviderMock, "MetricsProvider should be of type MetricsProviderMock, found " + metricsProvider.getClass());
+        MetricsProviderMock mp = (MetricsProviderMock) metricsProvider;
+        return mp;
+    }
+
     @Test
     public void testMetrics() throws MqttException {
-        MetricsProvider metricsProvider = server.getMetricsProvider();
-        MetricsProviderMock mp = null;
-        if (metricsProvider instanceof MetricsProviderMock) {
-            mp = (MetricsProviderMock) metricsProvider;
-        } else {
-            Assertions.fail("MetricsProvider should be of type MetricsProviderMock, found " + metricsProvider.getClass());
-        }
+        MetricsProviderMock mp = validateMetricsProvider();
 
         assertEquals(0, mp.getSessionCount());
         assertEquals(0, mp.getPublishCount());
@@ -144,12 +148,10 @@ public class MetricsTest {
         clientListener.disconnect();
         clientPublisher.disconnect();
 
-        try {
-            // Sleep shortly to give the metrics time to update.
-            Thread.sleep(100);
-        } catch (InterruptedException ex) {
-            // It's fine
-        }
+        Awaitility.await("Wait for the metrics to update.")
+            .pollInterval(Duration.ofMillis(10))
+            .atMost(Duration.ofMillis(100))
+            .until(() -> mp.getSessionCount(), Is.is(0));
 
         assertEquals(0, mp.getSessionCount());
         assertEquals(1, mp.getPublishCount());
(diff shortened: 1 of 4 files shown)


andsel:
There is a way to test it?

hylkevds:
Good question!
I think this should trigger when we start with a non-clean session, disconnect, and reconnect with a clean-session. Let me see if I can make a test for that.

hylkevds:
Yep, that worked.
I added a commit before the fix that has an extra test that should fail without the fix.

broker/src/test/java/io/moquette/metrics/MetricsTest.java
```diff
@@ -164,4 +164,75 @@ public void testMetrics() throws MqttException {
 
     }
 
+    @Test
+    public void testResopenNonCleanSessionWithClean() throws MqttException {
```
andsel:
```suggestion
    public void givenClosedNonCleanSessionWhenReconnectAsCleanThenSessionCountMetricsDoesntMissIt() throws MqttException {
```

broker/src/test/java/io/moquette/metrics/MetricsTest.java
```diff
@@ -164,4 +164,75 @@ public void testMetrics() throws MqttException {
...
+        MetricsProvider metricsProvider = server.getMetricsProvider();
+        MetricsProviderMock mp = null;
+        if (metricsProvider instanceof MetricsProviderMock) {
+            mp = (MetricsProviderMock) metricsProvider;
+        } else {
+            Assertions.fail("MetricsProvider should be of type MetricsProviderMock, found " + metricsProvider.getClass());
+        }
```
andsel: I would avoid conditions in tests:


```suggestion
        assertTrue(metricsProvider instanceof MetricsProviderMock, "MetricsProvider should be of type MetricsProviderMock, found " + metricsProvider.getClass());
        MetricsProviderMock mp = (MetricsProviderMock) metricsProvider;
```

broker/src/test/java/io/moquette/metrics/MetricsTest.java
```diff
@@ -164,4 +164,75 @@ public void testMetrics() throws MqttException {
...
+
+        clientListener.disconnect();
+        clientPublisher.disconnect();
+
+        try {
+            // Sleep shortly to give the metrics time to update.
+            Thread.sleep(100);
+        } catch (InterruptedException ex) {
+            // It's fine
+        }
```
andsel: Using Awaitility is safer way to sleep in tests:

```suggestion
        Awaitility.await("Sleep shortly to give the metrics time to update.")
            .pollDelay(Duration.ofMillis(50))
            .atMost(Duration.ofMillis(100))
            .until(() -> mp.getSessionCount() == 1);
```

broker/src/test/java/io/moquette/metrics/MetricsTest.java
```diff
@@ -164,4 +164,75 @@ public void testMetrics() throws MqttException {
...
```
andsel:
```suggestion
        Awaitility.await("Sleep shortly to give the metrics time to update.")
            .pollDelay(Duration.ofMillis(50))
            .atMost(Duration.ofMillis(100))
            .until(() -> mp.getSessionCount() == 0);
```

andsel (changes requested):
Hi @hylkevds thanks for proposing the fix, and in particular to cover it with test case. I've left some comments on this part, to improve the context and readability of the test.

In general I think the test scope is bigger than needed, in particular:
- is it necessary to also have a publisher and test that it published some data? (so maybe we can remove all the puvblish related things for the sake of testing session count metrics).
- we should clearly divide the test in phases, so that ipotetically a test should be of almost 4 calls. That simplify to understand which is the focus of the test:
  1. fixture setup
  2. exercise
  3. verify
  4. teardown

broker/src/test/java/io/moquette/metrics/MetricsTest.java
hylkevds: Good points.
I used the publisher as a timing mechanism, if the message comes through, then the system is ready for the next step. But the mock metrics provider can be polled easily, so that can be cleaned up. We do still need two connections, since disconnecting a non-clean session doesn't change any numbers in the metrics, so we can't test if the disconnect actually happened yet.

I changed the until() call a bit to `until(() -> mp.getSessionCount(), Is.is(2))` since that will show the actual value in the exception instead of just the statement that the condition was not met.

For the Prometheus test this is not quite that easy, since polling the metric involves doing an http call and parsing the response.

broker/src/test/java/io/moquette/metrics/MetricsTest.java
```diff
@@ -164,4 +166,64 @@ public void testMetrics() throws MqttException {
...
+        clientListener.disconnect();
+        clientPublisher.disconnect();
+
+        Awaitility.await("Sleep shortly to give the metrics time to update.")
+            .pollInterval(Duration.ofMillis(10))
+            .atMost(Duration.ofMillis(100))
+            .until(() -> mp.getSessionCount(), Is.is(1));
+
+        // There should now be one open session, the non-clean listener session.
+        assertEquals(1, mp.getSessionCount());
```
andsel: Maybe this is redundant given that's already verified by the Awaitility statement. The same in other Awaitility invocation.

andsel: I think all this piece can be extrated in a test helper method, like:
`fixturePrepareDisconnectedNonCleanSession`

broker/src/test/java/io/moquette/metrics/MetricsTest.java
```diff
@@ -164,4 +166,64 @@ public void testMetrics() throws MqttException {
...
+        clientListener.disconnect();
+        clientPublisher.disconnect();
+
+        Awaitility.await("Sleep shortly to give the metrics time to update.")
+            .pollInterval(Duration.ofMillis(10))
+            .atMost(Duration.ofMillis(100))
+            .until(() -> mp.getSessionCount(), Is.is(0));
+
+        // There should now be no open sessions.
+        assertEquals(0, mp.getSessionCount());
```
andsel: Is this part of the test related to the fix, so verify that the counters goes down when the client (with clean sessions) disconnects?

andsel (changes requested):
Hi, thank's for Awaitility fix.
Left some comment to better structure the test plus a question to understand better the scope of the test.

broker/src/test/java/io/moquette/metrics/MetricsTest.java
hylkevds: Yes, those are now redundant, I'll remove them.

broker/src/test/java/io/moquette/metrics/MetricsTest.java
hylkevds: That depends on whether you see this test as a very specific one just for this one bug, or as a more general test that checks if session counting works properly. The first place where the bug manifests is where the count goes from 1 back to 2, since the bug causes the count to not go up. After that, all counts are one lower than they should be, so the last one will be -1 with the bug.

That's also why I think moving half the test-script to a separate method is not a good idea. That part may just be set-up when only considering this very specific bug, but it does still test things: 
- Is the session counter increased when a new session opens.
- Is the session counter decreased when a clean session disconnects
- Is the session counter not decreased when a non-clean session disconnects.

And while we could make separate tests for each of these, that would be pointless, since those actions need to be part of this test any way. And since the order of tests is not fixed, chances are that this test runs first and thus a bug in one of those parts is triggered by this test first...

This test tests the behaviour of the entire system, not a single, stand-alone function. So each of the assertions in this test-script is equally important.

andsel: Godd point, that's fair 👍

andsel (commented):
@hylkevds I'm good with, just update the changelog and we are fine to merge.

broker/src/test/java/io/moquette/metrics/MetricsTest.java
hylkevds: I did move the MetricsProvider implementation validation to its own method, and used that in both tests in the file.

I also replaced the Thread.sleep() in the other test with an awaitility check.

hylkevds:
> @hylkevds I'm good with, just update the changelog and we are fine to merge.

When will I not forget to update the changelog...

Done now :)

andsel (approved):
LGTM when CI is 🟢

andsel:
Thanks @hylkevds for this contribution 💯

PR #941, opened 2026-06-06
Add Nginx Proxy Protocol Support & Fix Exceptions

norkts:
### Feature
- Add support for Nginx proxy_protocol
- Fix several abnormal connection exceptions

### Configuration
```
# Enable PROXY protocol support (disabled by default)
proxy_protocol.enabled true

# Auto-detect mode (supports both proxied and direct traffic, default)
proxy_protocol.expected false
```
### Behavior Description
1. enabled=true, expected=false (Default)
Auto detection mode: decode connection with PROXY header; process normal MQTT traffic directly if no PROXY header exists.
2. enabled=true, expected=true
Strict mode: only accept connections with valid PROXY header; reject connections without PROXY protocol.
3. enabled=false
Disable PROXY protocol parser, use standard native MQTT processing.


hiSandog:
Proxy protocol support is useful, but auto-detect mode needs careful tests because the first bytes decide whether the connection is PROXY or native MQTT. It would be good to cover direct MQTT traffic whose first byte is not a valid PROXY prefix, malformed/partial PROXY headers, and strict mode rejecting non-PROXY clients before MQTT parsing starts. That would also document how client IP attribution behaves in each mode.

PR #946, opened 2026-06-12
Expose client remote address in connect intercept

### LINKED ISSUE #867: Is there any way to get the ip of the client?
Is there any way to get the ip of the client?

HTHou:
## Summary

- Include the client remote socket address in `InterceptConnectMessage` for CONNECT interceptors.
- Expose remote address, client address, and client port accessors.
- Keep the existing one-argument connect interception API compatible without retaining the live Netty `Channel`.

## Test
- `./mvnw -pl broker -Dtest=BrokerInterceptorTest test`

Fixes #867

### DIFF
--- a/broker/src/main/java/io/moquette/interception/messages/InterceptConnectMessage.java
+++ b/broker/src/main/java/io/moquette/interception/messages/InterceptConnectMessage.java
@@ -16,19 +16,52 @@
 
 package io.moquette.interception.messages;
 
+import io.netty.channel.Channel;
 import io.netty.handler.codec.mqtt.MqttConnectMessage;
+
+import java.net.InetSocketAddress;
+import java.util.Optional;
 
 public class InterceptConnectMessage extends InterceptAbstractMessage {
 
     private final MqttConnectMessage msg;
+    private final Channel channel;
 
     public InterceptConnectMessage(MqttConnectMessage msg) {
+        this(msg, null);
+    }
+
+    public InterceptConnectMessage(MqttConnectMessage msg, Channel channel) {
         super(msg);
         this.msg = msg;
+        this.channel = channel;
     }
 
     public String getClientID() {
         return msg.payload().clientIdentifier();
+    }
+
+    public Optional<Channel> getChannel() {
+        return Optional.ofNullable(channel);
+    }
+
+    public Optional<InetSocketAddress> getRemoteAddress() {
+        if (channel != null && channel.remoteAddress() instanceof InetSocketAddress) {
+            return Optional.of((InetSocketAddress) channel.remoteAddress());
+        }
+        return Optional.empty();
+    }
+
+    public String getClientAddress() {
+        return getRemoteAddress()
+            .map(InetSocketAddress::getHostString)
+            .orElse(null);
+    }
+
+    public int getClientPort() {
+        return getRemoteAddress()
+            .map(InetSocketAddress::getPort)
+            .orElse(-1);
     }
 
     public boolean isCleanSession() {
--- a/broker/src/main/java/io/moquette/interception/messages/InterceptConnectMessage.java
+++ b/broker/src/main/java/io/moquette/interception/messages/InterceptConnectMessage.java
@@ -16,7 +16,6 @@
 
 package io.moquette.interception.messages;
 
-import io.netty.channel.Channel;
 import io.netty.handler.codec.mqtt.MqttConnectMessage;
 
 import java.net.InetSocketAddress;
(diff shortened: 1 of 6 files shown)


andsel:
Hi @HTHou thanks for you contribution. Given that `InterceptConnectMessage` is like a data class to move around some information, I wouldn't pass a channel why is a live object. I prefer to extract that information from the channel when the `InterceptConnectMessage`  is created, ideally as was proposed in https://github.com/moquette-io/moquette/issues/867#issuecomment-2492830421.

HTHou:
Thanks @andsel. I have updated the PR so that `InterceptConnectMessage` only stores the extracted `InetSocketAddress` and no longer retains or exposes the live Netty `Channel`.

The reason I initially exposed the `Channel` was to support another use case: allowing an interceptor or application to actively disconnect a client, for example when it is considered inactive by a custom policy. If exposing `Channel` directly is not appropriate, would it make sense to provide a controlled API for this, such as `disconnectClient(clientId)` or `closeConnection(clientId)`?

andsel:
I would prefer a clean API/SPI instance provided by Moquette to the implementors of listeners to cleanly interact with what to do with channel, like in this case drop a connection.

HTHou:
Thanks, that makes sense. I can address the controlled disconnect capability in a separate PR with a clean Moquette API/SPI for listener implementations. I will keep the scope of this PR limited to exposing the client's remote address through `InterceptConnectMessage`, without exposing the Netty `Channel`.

Could you please review this PR when you have a chance? Thanks!

broker/src/test/java/io/moquette/interception/BrokerInterceptorTest.java
```diff
@@ -123,6 +128,42 @@ public void testNotifyClientConnected() throws Exception {
         assertEquals(40, n.get());
     }
 
+    @Test
+    public void testNotifyClientConnectedIncludesRemoteAddress() throws Exception {
+        final CountDownLatch notified = new CountDownLatch(1);
+        final AtomicReference<InterceptConnectMessage> intercepted = new AtomicReference<>();
+        final BrokerInterceptor localInterceptor = new BrokerInterceptor(
+            Collections.<InterceptHandler>singletonList(new AbstractInterceptHandler() {
```
andsel: Maybe we can use `List.of` or do you need that the list is unmodifiable?

broker/src/test/java/io/moquette/interception/BrokerInterceptorTest.java
```diff
@@ -123,6 +128,42 @@ public void testNotifyClientConnected() throws Exception {
...
+                public void onSessionLoopError(Throwable error) {
+                    throw new RuntimeException(error);
+                }
+            }));
+
+        try {
+            final InetSocketAddress remoteAddress = new InetSocketAddress("127.0.0.1", 12345);
+            localInterceptor.notifyClientConnected(MqttMessageBuilders.connect().build(), remoteAddress);
+
+            assertTrue(notified.await(1, TimeUnit.SECONDS));
```
andsel: This can be replaced with Awaitility and the same condition.

andsel (changes requested):
Need some nitpicks, but I think the PR is close to be merged.

broker/src/test/java/io/moquette/interception/BrokerInterceptorTest.java
HTHou: Updated in 38dd93dc. I replaced the singleton list usages with `List.of(...)`. Thanks!

broker/src/test/java/io/moquette/interception/BrokerInterceptorTest.java
HTHou: Updated in 38dd93dc. I replaced the `CountDownLatch` with `Awaitility.await().until(() -> intercepted.get() != null)`. Thanks!

HTHou:
Thanks @andsel. I addressed both review comments in 38dd93dc: the test now uses `List.of(...)` and Awaitility, and `BrokerInterceptorTest` passes. Both review threads have been resolved. Could you please take another look when you have a chance?

broker/src/main/java/io/moquette/interception/messages/InterceptConnectMessage.java
```diff
@@ -18,19 +18,44 @@
...
     }
 
+    public Optional<InetSocketAddress> getRemoteAddress() {
+        return Optional.ofNullable(remoteAddress);
+    }
+
+    public String getClientAddress() {
+        return getRemoteAddress()
+            .map(InetSocketAddress::getHostString)
+            .orElse(null);
```
andsel: `getRemoteAddress()` wraps an instance with an Optional and here we unwrap it back. So I think that functional API are good, but in this case I think it's an overkill.

```suggestion
        return remoteAddress != null ? remoteAddress.getHostString() : null;
```

broker/src/main/java/io/moquette/interception/messages/InterceptConnectMessage.java
```diff
@@ -18,19 +18,44 @@
...
+    }
+
+    public int getClientPort() {
+        return getRemoteAddress()
+            .map(InetSocketAddress::getPort)
+            .orElse(-1);
```
andsel: Same overkill also here:

```suggestion
        return remoteAddress != null ? remoteAddress.getPort() : -1;
```

andsel (changes requested):
Hi @HTHou thanks a lot to have addressed my concerns.
Left a couple of nitpicks then we are good to merge 🙏

broker/src/main/java/io/moquette/interception/messages/InterceptConnectMessage.java
HTHou: Updated in 280bcf8b. `getClientAddress()` now directly checks `remoteAddress` and returns its host string, as suggested. Thanks!

broker/src/main/java/io/moquette/interception/messages/InterceptConnectMessage.java
HTHou: Updated in 280bcf8b. `getClientPort()` now directly checks `remoteAddress` and returns its port, as suggested. Thanks!

HTHou:
Thanks @andsel. I addressed the two latest suggestions in 280bcf8b and resolved both review threads. `BrokerInterceptorTest` still passes with all 7 tests. Could you please take another look when you have a chance?

andsel (approved):
LGTM

andsel:
Thanks @HTHou for your contribution 👍

PR #955, opened 2026-07-04
Updated Unsubscribe notification mirror subscribe notification

hylkevds:
Currently, the InterceptSubscribeMessage gets the subscription and the username, but the InterceptUnsubscribeMessage only gets topic, clientId and username. Probably because before the subscription cleanup the Subscription was not available at the point where the intercept message was sent.

Since unsubscribe now works on the Subscription, the InterceptUnsubscribeMessage can mirror the Subscribe equivalent, and have the full Subscription instead of just the topic, username and clientID.

- enhancement

### DIFF
--- a/broker/src/main/java/io/moquette/interception/Interceptor.java
+++ b/broker/src/main/java/io/moquette/interception/Interceptor.java
@@ -44,7 +44,7 @@ public interface Interceptor {
 
     void notifyTopicSubscribed(Subscription sub, String username);
 
-    void notifyTopicUnsubscribed(String topic, String clientID, String username);
+    void notifyTopicUnsubscribed(Subscription sub, String username);
 
     void notifyMessageAcknowledged(InterceptAcknowledgedMessage msg);
 
--- a/broker/src/main/java/io/moquette/interception/messages/InterceptUnsubscribeMessage.java
+++ b/broker/src/main/java/io/moquette/interception/messages/InterceptUnsubscribeMessage.java
@@ -42,11 +42,11 @@ public class InterceptUnsubscribeMessage implements InterceptMessage {
     }
 
     public String getTopicFilterClient() {
-        return subscription.getTopicFilter().toString();
+        return subscription.getTopicFilterClient().toString();
     }
 
-    public String getTopicFilterInternal() {
-        return subscription.getTopicFilter().toString();
+    public String getTopicFilterRewritten() {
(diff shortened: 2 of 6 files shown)


broker/src/main/java/io/moquette/interception/Interceptor.java
```diff
@@ -44,7 +44,7 @@ public interface Interceptor {
 
     void notifyTopicSubscribed(Subscription sub, String username);
 
-    void notifyTopicUnsubscribed(String topic, String clientID, String username);
+    void notifyTopicUnsubscribed(Subscription sub, String username);
```
andsel, line 47: This is an API change, I think if should be described in the Changelog to help others which encounter a compilation error when include the new Moquette versions.
We are in the pre-major releases so generally, we can break APIs on patch/minors, but we have ease the transition.

broker/src/main/java/io/moquette/interception/messages/InterceptUnsubscribeMessage.java
```diff
@@ -13,27 +13,40 @@
...
+        return subscription.getOption();
+    }
+
+    public String getTopicFilterClient() {
+        return subscription.getTopicFilter().toString();
+    }
+
+    public String getTopicFilterInternal() {
+        return subscription.getTopicFilter().toString();
     }
```
andsel, line 50: Are these related to #932, if so we have to merge that PR before.

andsel (changes requested):
Left a couple of comments.

broker/src/main/java/io/moquette/interception/messages/InterceptUnsubscribeMessage.java
hylkevds: Ah, I backported it from a branch that has 932, and I forgot to back-name that method. It should just be `getTopicFilter` and 932 will have to be updated to change the name to `getTopicFilterRedirected` and add `getTopicFilterClient`.
Fixed now

broker/src/main/java/io/moquette/interception/Interceptor.java
hylkevds: I forgot the changelog again :-D
Updated!

hylkevds:
I've updated the PR to reflect the topicRewriting changes.

andsel (approved):
LGTM

PR #957, opened 2026-07-06
Keep the session event loop alive when a command throws (remote DoS)

gigioneggiando:
## What does this PR do?
Makes `SessionEventLoop.run()` catch `Throwable` per command and continue, instead of letting the `RuntimeException` rethrown by `executeTask()` propagate out of `run()` and terminate the thread.

## Why is it important?
Each session loop is shared by all client ids that hash to its ordinal (`abs(clientId.hashCode()) % availableProcessors`). Today a single command that throws (e.g. an uncaught exception while handling a malformed packet) kills the whole loop and it is never restarted, so every other client bound to that loop can no longer have its PUBLISH/SUBSCRIBE/PUBACK/... commands processed. An attacker can spread client ids across all loops and take down session processing for the entire broker. The `FutureTask` already captures the failure for its submitter, so the shared loop should not die with it.

This is the structural containment for a whole class of per-command crashes (malformed `$share` filter, deeply nested topic, null-field NPEs), mirroring the containment `NewNettyMQTTHandler.channelRead` already applies on the Netty thread (close only the offending channel).

## Checklist
- [x] I have added tests that prove my fix is effective (`SessionEventLoopTest`)

`mvn -pl broker -Dtest=SessionEventLoopTest test` - the test submits a throwing command then a normal one and asserts the loop still executes the latter.

Reported privately via the repository Security Advisory; opening this DoS-hardening fix as a public PR per the maintainer. cc @andsel

### DIFF
--- a/broker/src/main/java/io/moquette/broker/SessionEventLoop.java
+++ b/broker/src/main/java/io/moquette/broker/SessionEventLoop.java
@@ -49,11 +49,8 @@ final class SessionEventLoop extends Thread {
                 LOG.info("SessionEventLoop {} interrupted", Thread.currentThread().getName());
                 Thread.currentThread().interrupt();
             } catch (Throwable th) {
-                // A single failing command must never terminate the shared session loop: the FutureTask
-                // has already captured the failure for its submitter. Log and keep serving the other
-                // sessions bound to this loop. Previously the RuntimeException rethrown by executeTask
-                // propagated out of run() and killed the thread, wedging every co-located client (a
-                // malformed packet on one session became a partitioned, broker-wide denial of service).
+                // Keep the shared loop alive on a failing command; the FutureTask already captured the
+                // failure for its submitter, so co-located sessions must not be taken down with it.
                 LOG.error("SessionEventLoop {} caught an unhandled error while processing a command; "
                         + "the loop continues", Thread.currentThread().getName(), th);
             }
--- a/broker/src/main/java/io/moquette/broker/SessionEventLoop.java
+++ b/broker/src/main/java/io/moquette/broker/SessionEventLoop.java
@@ -48,6 +48,14 @@ final class SessionEventLoop extends Thread {
             } catch (InterruptedException e) {
                 LOG.info("SessionEventLoop {} interrupted", Thread.currentThread().getName());
                 Thread.currentThread().interrupt();
+            } catch (Throwable th) {
+                // A single failing command must never terminate the shared session loop: the FutureTask
+                // has already captured the failure for its submitter. Log and keep serving the other
+                // sessions bound to this loop. Previously the RuntimeException rethrown by executeTask
+                // propagated out of run() and killed the thread, wedging every co-located client (a
+                // malformed packet on one session became a partitioned, broker-wide denial of service).
+                LOG.error("SessionEventLoop {} caught an unhandled error while processing a command; "
+                        + "the loop continues", Thread.currentThread().getName(), th);
             }
         }
         LOG.info("SessionEventLoop {} exit", Thread.currentThread().getName());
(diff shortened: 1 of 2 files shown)


broker/src/main/java/io/moquette/broker/SessionEventLoop.java
```diff
@@ -48,6 +48,14 @@ public void run() {
             } catch (InterruptedException e) {
                 LOG.info("SessionEventLoop {} interrupted", Thread.currentThread().getName());
                 Thread.currentThread().interrupt();
+            } catch (Throwable th) {
+                // A single failing command must never terminate the shared session loop: the FutureTask
+                // has already captured the failure for its submitter. Log and keep serving the other
+                // sessions bound to this loop. Previously the RuntimeException rethrown by executeTask
+                // propagated out of run() and killed the thread, wedging every co-located client (a
+                // malformed packet on one session became a partitioned, broker-wide denial of service).
```
andsel: Avoid such long comment to describe the code added.

broker/src/test/java/io/moquette/broker/SessionEventLoopTest.java
```diff
@@ -0,0 +1,58 @@
+/*
+ * Copyright (c) 2012-2018 The original author or authors
```
andsel:
```suggestion
 * Copyright (c) 2012-2026 The original author or authors
```

broker/src/test/java/io/moquette/broker/SessionEventLoopTest.java
```diff
@@ -0,0 +1,58 @@
...
+import java.util.concurrent.TimeUnit;
+
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+class SessionEventLoopTest {
+
+    @Test
+    public void aCommandThatThrowsMustNotTerminateTheSharedSessionLoop() throws InterruptedException {
+        final BlockingQueue<FutureTask<String>> queue = new LinkedBlockingQueue<>();
+        final SessionEventLoop loop = new SessionEventLoop(queue, 0, new MetricsProviderNull());
```
andsel:
```suggestion
        final SessionEventLoop sut = new SessionEventLoop(queue, 0, new MetricsProviderNull());
```

SUT is the system under test which the subject of the test, so naming it sut makes immediately clear which is the center of the test.

broker/src/test/java/io/moquette/broker/SessionEventLoopTest.java
```diff
@@ -0,0 +1,58 @@
...
+            }));
+
+            // A subsequent command bound to the SAME loop must still be executed.
+            final CountDownLatch executed = new CountDownLatch(1);
+            queue.add(new FutureTask<>(() -> {
+                executed.countDown();
+                return "ok";
+            }));
+
+            assertTrue(executed.await(5, TimeUnit.SECONDS),
```
andsel: Maybe we can use Awaitility assertion here.

broker/src/test/java/io/moquette/broker/SessionEventLoopTest.java
```diff
@@ -0,0 +1,58 @@
...
+import java.util.concurrent.FutureTask;
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.TimeUnit;
+
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+class SessionEventLoopTest {
+
+    @Test
+    public void aCommandThatThrowsMustNotTerminateTheSharedSessionLoop() throws InterruptedException {
```
andsel:
```suggestion
    public void givenAnErroringCommandWhenExecutedByTheLoopThenDoesntKillTheThreadExecutingIt() throws InterruptedException {
```

andsel (changes requested):
Left some comments and possible suggestions.

gigioneggiando:
Thanks! Addressed:
- Trimmed the inline comment on the `Throwable` catch to two lines.
- Renamed the test to the `given…when…then` pattern and the subject under test to `sut`.
- Swapped the `CountDownLatch` wait for an Awaitility condition (`untilTrue`).
- Bumped the copyright header to 2026.

andsel (approved):
LGTM

PR #959, opened 2026-07-06
Guard the periodic expiration task so one failing entity can't disable all expirations

gigioneggiando:
## What does this PR do?
Wraps `ScheduledExpirationService.checkExpiredEntities()` so an exception from the expiration action can never escape the periodic task, and iterates the drained entities one by one so a single failing action does not skip the rest of the batch. Adds a test.

## Why is it important?
The task is scheduled with `scheduleWithFixedDelay`. Per its contract, if a run throws, the executor stops rescheduling it - permanently. So a single throwing expiration action silently disables ALL future expirations: delayed Wills never fire, retained messages and sessions never expire, and that state accumulates without bound until the broker restarts. Three services share this pattern (Will/retained expiry, session expiry, and the check loop).

## Checklist
- [x] I have added tests that prove my fix is effective (`ScheduledExpirationServiceTest`)

`mvn -pl broker -Dtest=ScheduledExpirationServiceTest test` - tracks two already-expired entities whose action throws for one of them, runs the check, and asserts it does not throw and still processes the other.

Reported privately via the repository Security Advisory; opening this hardening fix as a public PR per the maintainer. cc @andsel

### DIFF
--- a/broker/src/main/java/io/moquette/broker/scheduler/ScheduledExpirationService.java
+++ b/broker/src/main/java/io/moquette/broker/scheduler/ScheduledExpirationService.java
@@ -38,14 +38,26 @@ public class ScheduledExpirationService<T extends Expirable> {
             TimeUnit.SECONDS);
     }
 
-    private void checkExpiredEntities() {
-        List<ExpirableTracker<T>> expiredEntities = new ArrayList<>();
-        int drainedEntities = expiringEntities.drainTo(expiredEntities);
-        LOG.debug("Retrieved {} expired entity on {}", drainedEntities, expiringEntities.size());
+    // Package-private for testing. Runs from scheduleWithFixedDelay: if an exception escapes, the
+    // executor silently stops rescheduling and NO further expirations ever fire again. Never let one
+    // throw out, and never let one failing entity skip the rest of the batch.
+    void checkExpiredEntities() {
+        try {
+            List<ExpirableTracker<T>> expiredEntities = new ArrayList<>();
+            int drainedEntities = expiringEntities.drainTo(expiredEntities);
+            LOG.debug("Retrieved {} expired entity on {}", drainedEntities, expiringEntities.size());
 
-        expiredEntities.stream()
-            .map(ExpirableTracker::expirable)
-            .forEach(action);
+            for (ExpirableTracker<T> tracker : expiredEntities) {
+                try {
+                    action.accept(tracker.expirable());
+                } catch (Throwable th) {
+                    LOG.warn("Expiration action failed for an entity; continuing with the remaining ones", th);
+                }
+            }
+        } catch (Throwable th) {
+            // Guarantee the periodic firer keeps running regardless.
+            LOG.error("Unexpected error while checking expired entities; the periodic task continues", th);
+        }
     }
 
     public void track(String entityId, T entity) {
--- a/broker/src/main/java/io/moquette/broker/scheduler/ScheduledExpirationService.java
+++ b/broker/src/main/java/io/moquette/broker/scheduler/ScheduledExpirationService.java
@@ -31,7 +31,7 @@ public class ScheduledExpirationService<T extends Expirable> {
 
     public ScheduledExpirationService(Clock clock, Consumer<T> action) {
         this.clock = clock;
-        this.action = guarded(action);
+        this.action = exceptionSafeWrapper(action);
         this.actionsExecutor = Executors.newSingleThreadScheduledExecutor();
         this.expiredEntityTask = actionsExecutor.scheduleWithFixedDelay(this::checkExpiredEntities,
             FIRER_TASK_INTERVAL.getSeconds(), FIRER_TASK_INTERVAL.getSeconds(),
@@ -41,7 +41,7 @@ public class ScheduledExpirationService<T extends Expirable> {
     // The action runs from a scheduleWithFixedDelay task: if it throws, the executor silently stops
     // rescheduling and no further expirations ever fire. Decorate it so a failing entity is logged and
     // the batch continues, instead of letting the exception escape into the scheduler.
-    private Consumer<T> guarded(Consumer<T> delegate) {
+    private Consumer<T> exceptionSafeWrapper(Consumer<T> delegate) {
         return entity -> {
             try {
                 delegate.accept(entity);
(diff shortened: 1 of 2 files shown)


broker/src/main/java/io/moquette/broker/scheduler/ScheduledExpirationService.java
```diff
@@ -38,14 +38,26 @@ public ScheduledExpirationService(Clock clock, Consumer<T> action) {
...
 
-    private void checkExpiredEntities() {
-        List<ExpirableTracker<T>> expiredEntities = new ArrayList<>();
-        int drainedEntities = expiringEntities.drainTo(expiredEntities);
-        LOG.debug("Retrieved {} expired entity on {}", drainedEntities, expiringEntities.size());
+    // Package-private for testing. Runs from scheduleWithFixedDelay: if an exception escapes, the
+    // executor silently stops rescheduling and NO further expirations ever fire again. Never let one
+    // throw out, and never let one failing entity skip the rest of the batch.
+    void checkExpiredEntities() {
+        try {
```
andsel: While I understand to make the `action` safe (we can use a private decorator method for that) I don't understand in which cases the `drainTo` or the iteration on `expiredEntities` could throw an error.

broker/src/test/java/io/moquette/broker/scheduler/ScheduledExpirationServiceTest.java
```diff
@@ -0,0 +1,76 @@
...
+        }
+
+        @Override
+        public Optional<Instant> expireAt() {
+            return Optional.of(expireAt);
+        }
+    }
+
+    @Test
+    public void aThrowingExpirationActionMustNotStopThePeriodicService() {
```
andsel:
```suggestion
    public void givenAnActionThatThrowsAnErrorWhenExecutedThenDoesntStopExecutingNextScheduledActions() {
```

broker/src/test/java/io/moquette/broker/scheduler/ScheduledExpirationServiceTest.java
```diff
@@ -0,0 +1,76 @@
...
+        final ScheduledExpirationService<Entity> service = new ScheduledExpirationService<>(clock, action);
+        try {
+            final Instant alreadyExpired = Instant.ofEpochMilli(9_000); // < clock -> ready to be drained
+            service.track("boom", new Entity("boom", alreadyExpired));
+            service.track("good", new Entity("good", alreadyExpired));
+
+            // Running the periodic check must not propagate the action's exception (which would make
+            // scheduleWithFixedDelay stop firing forever)...
+            assertDoesNotThrow(service::checkExpiredEntities);
+            // ...and the non-throwing entity must still have been processed in the same batch.
```
andsel:
```suggestion
```

The code should speak itself without too much commenting.

broker/src/test/java/io/moquette/broker/scheduler/ScheduledExpirationServiceTest.java
```diff
@@ -0,0 +1,76 @@
...
+            assertTrue(processed.contains("good"),
```
andsel:
```suggestion
            assertThat(processed).contains("good");
```

Using AssertJ provides better explanation then `assertTrue` when it fails.

andsel (changes requested):
Thanks @gigioneggiando for the PR. Left a couple of notes.

broker/src/main/java/io/moquette/broker/scheduler/ScheduledExpirationService.java
```diff
@@ -31,21 +31,35 @@ public class ScheduledExpirationService<T extends Expirable> {
...
         List<ExpirableTracker<T>> expiredEntities = new ArrayList<>();
         int drainedEntities = expiringEntities.drainTo(expiredEntities);
         LOG.debug("Retrieved {} expired entity on {}", drainedEntities, expiringEntities.size());
 
-        expiredEntities.stream()
-            .map(ExpirableTracker::expirable)
-            .forEach(action);
+        for (ExpirableTracker<T> tracker : expiredEntities) {
+            action.accept(tracker.expirable());
+        }
```
andsel:
```suggestion
        fexpiredEntities.stream()
            .map(ExpirableTracker::expirable)
            .forEach(action);
```

At this point given that original `action` is wrapped by `guarded` we can go back to the original stream form.

broker/src/main/java/io/moquette/broker/scheduler/ScheduledExpirationService.java
```diff
@@ -31,21 +31,35 @@ public class ScheduledExpirationService<T extends Expirable> {
...
         this.expiredEntityTask = actionsExecutor.scheduleWithFixedDelay(this::checkExpiredEntities,
             FIRER_TASK_INTERVAL.getSeconds(), FIRER_TASK_INTERVAL.getSeconds(),
             TimeUnit.SECONDS);
     }
 
-    private void checkExpiredEntities() {
+    // The action runs from a scheduleWithFixedDelay task: if it throws, the executor silently stops
+    // rescheduling and no further expirations ever fire. Decorate it so a failing entity is logged and
+    // the batch continues, instead of letting the exception escape into the scheduler.
+    private Consumer<T> guarded(Consumer<T> delegate) {
```
andsel:
```suggestion
    private Consumer<T> exceptionSafeWrapper(Consumer<T> delegate) {
```

andsel (changes requested):
Thanks @gigioneggiando , just left a couple of nitpicks before merging.

broker/src/test/java/io/moquette/broker/scheduler/ScheduledExpirationServiceTest.java
```diff
@@ -0,0 +1,72 @@
+/*
+ * Copyright (c) 2012-2023 The original author or authors
```
andsel:
```suggestion
 * Copyright (c) 2012-2026 The original author or authors
```

gigioneggiando:
Thanks @andsel! Both nitpicks are in `ed72f2e3`: renamed `guarded` → `exceptionSafeWrapper`, and — now that the action is wrapped — restored the original `stream().map(ExpirableTracker::expirable).forEach(action)` form in `checkExpiredEntities()`. Thanks as well for fixing the copyright date yourself 🙏 Should be ready for another look whenever you have a moment!

andsel (approved):
Thanks @gigioneggiando for you effort on this, well done :-)

LGTM

PR #963, opened 2026-07-16
Guard pattern ACL substitution against a null username (remote DoS)

gigioneggiando:
## What does this PR do?
Makes the pattern-ACL branch of `AuthorizationsCollector.canDoOperation()` treat a missing (`null`) client id / username as an empty string before substituting it into the rule, instead of passing `null` into `String.replace()`.

## Why is it important?
An anonymous client has a `null` username (`login()` returns `true` without calling `NettyUtils.userName`). When a `pattern` ACL rule is configured, the first authorization check substitutes the identity into the rule via `auth.topic.toString().replace("%c", client).replace("%u", username)`. `String.replace(CharSequence, CharSequence)` calls `replacement.toString()` eagerly, so a `null` username throws `NullPointerException` regardless of whether `%u` appears in the rule. That check runs on the session event loop, so the unchecked exception kills the worker thread and every co-located session stops being served (the same broker-wide DoS amplifier as #957). Substituting an empty string fails closed: a `%u` rule simply does not match an anonymous client, while `%c` rules keep working.

(The sibling wildcard-in-identity case is already handled by the `hasTopicWildcard` guard; this PR closes the remaining `null`-identity path.)

## Checklist
- [x] I have added tests that prove my fix is effective (`AuthorizationsCollectorTest#givenPatternACLWhenUsernameIsNullThenSubstitutesSafelyWithoutThrowing`)

`mvn -pl broker -Dtest=AuthorizationsCollectorTest test`

Reported privately via the repository Security Advisory; opening this DoS-hardening fix as a public PR per the maintainer. cc @andsel

### DIFF
--- a/ChangeLog.txt
+++ b/ChangeLog.txt
@@ -1,4 +1,5 @@
 Version 0.19-SNAPSHOT
+   [fix] Guarded pattern ACL substitution against a null username to prevent a remote DoS. (#963)
    [feature] InterceptUnsubscribeMessage now has the entire Subscription and the notification hook has changed
      from `notifyTopicUnsubscribed(String topic, String clientID, String username)`
      to `notifyTopicUnsubscribed(Subscription sub, String username)`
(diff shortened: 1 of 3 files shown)


andsel (changes requested):
@gigioneggiando it's almost good to me, please update the Changelog.txt file

andsel (dismissed):
LGTM

andsel (approved):
LGTM

PR #964, opened 2026-07-16
Namespace the H2 queue metadata map to prevent a client-id collision (durable corruption)

gigioneggiando:
## What does this PR do?
Gives the H2 persistent-queue **message** map and **metadata** map disjoint name spaces so no client id can make one client's map collide with another's. The metadata map moves from `queue_<id>_meta` to `queuemeta_<id>`; the message map name (`queue_<id>`) is unchanged, and a one-time migration copies the head/tail pointers out of any legacy `queue_<id>_meta` map and drops it.

## Why is it important?
`H2QueueRepository.getOrCreateQueue(clientId)` uses the raw client id as the queue name, and `H2PersistentQueue` built both map names by suffixing a shared `queue_` prefix:

```
messageMap  = "queue_" + queueName            // Long -> EnqueuedMessage
metadataMap = "queue_" + queueName + "_meta"   // String -> Long (head/tail)
```

Because the names are plain concatenations, `"queue_" + "sensor_meta"` is byte-identical to `"queue_" + "sensor" + "_meta"`. A client that connects with a durable (non-clean) session and client id `sensor_meta` therefore opens, as its **message** map, the exact MVStore map that client `sensor` uses as its **metadata** map — two sessions reading/writing the same map with incompatible value types. The result is durable cross-session corruption of client `sensor`'s queue (message loss, misdelivery, or a failed reload), and a crafted collision can expose the victim's queued content. Client ids are never validated, so only knowing/guessing a victim id is required.

`queuemeta_` differs from `queue_` at a fixed position (index 5: `'m'` vs `'_'`), so a message map name can never equal a metadata map name for any queue name — the collision is impossible by construction rather than by input validation.

**Upgrade / compatibility:** message maps keep their names, so durable messages are preserved. On first open of an existing queue, `migrateLegacyMetadata` copies `head`/`tail` from the old `queue_<id>_meta` map into the new one and removes the legacy map (removal is required so a later `<id>_meta` client can't reopen the leftover as its message map). Stores with no durable queues are unaffected.

## Checklist
- [x] I have added tests that prove my fix is effective (`H2PersistentQueueTest`: a `sensor` / `sensor_meta` non-collision test and a legacy-metadata migration test)

`mvn -pl broker -Dtest=H2PersistentQueueTest test`. (On Windows the unrelated `SegmentPersistentQueueTest` reports a pre-existing `@TempDir` cleanup error on the mmap-ed store; it is not touched by this change.)

Reported privately via the repository Security Advisory; opening this hardening fix as a public PR per the maintainer. cc @andsel

### DIFF
--- a/ChangeLog.txt
+++ b/ChangeLog.txt
@@ -1,4 +1,5 @@
 Version 0.19-SNAPSHOT
+   [fix] Namespaced the H2 queue metadata map to prevent a client-id collision that could corrupt a durable queue. (#964)
    [fix] Check Last Will topic for writability rights before publish.
    [fix] Reject a malformed $share subscription before parsing it (remote DoS). (#958)
    [fix] Keep the session event loop alive when a command throws (remote DoS). (#957)
--- a/broker/src/main/java/io/moquette/persistence/H2PersistentQueue.java
+++ b/broker/src/main/java/io/moquette/persistence/H2PersistentQueue.java
@@ -25,6 +25,15 @@ import java.util.concurrent.atomic.AtomicLong;
 
 class H2PersistentQueue extends AbstractSessionMessageQueue<SessionRegistry.EnqueuedMessage> {
 
+    // The message map and the metadata map must live in disjoint name spaces. The queue name is a raw
+    // client id, so deriving both names from a shared "queue_" prefix let a client id ending in "_meta"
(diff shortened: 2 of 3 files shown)


andsel:
@gigioneggiando please also update the Changelog.txt file.

andsel (approved):
LGTM

PR #967, opened 2026-07-18
Fail closed when a configured authorizator/authenticator class fails to load

gigioneggiando:
## What does this PR do?
Makes `Server.initializeAuthorizatorPolicy()` / `initializeAuthenticator()` abort startup with a clear `IllegalArgumentException` when the operator has configured `authorizator_class`/`authenticator_class` and that class fails to load, instead of silently falling through to the permissive default (`PermitAllAuthorizatorPolicy` / `AcceptAllAuthenticator`) reserved for the "nothing configured" case.

## Why is it important?
`loadClass(...)` catches reflection failures (typo in the class name, missing/renamed jar, a constructor that throws, a classpath problem) and returns `null`, logging only a WARN. Both `initializeAuthorizatorPolicy` and `initializeAuthenticator` then treat that `null` exactly the same as "no class was configured at all", falling through to the default branch and booting fully open: any client authenticates, every publish/subscribe is authorized. This is the dangerous case because it fires precisely when the operator *believes* auth is enforced (they configured a custom class), not the documented/intentional "no security configured" default - and it's easy to miss a single WARN line at startup.

This PR does not change behavior when nothing is configured: `initializeAuthorizatorPolicy`/`initializeAuthenticator` still fall back to the ACL-file/PermitAll and password-file/AcceptAll defaults exactly as before in that case (covered by a test below).

## Checklist
- [x] I have added tests that prove my fix is effective (`ServerAuthConfigFailClosedTest`: a configured-but-missing authorizator class and authenticator class each now throw `IllegalArgumentException`; a third test confirms the no-class-configured default `PermitAllAuthorizatorPolicy` path is unaffected)

`mvn -pl broker -Dtest=ServerAuthConfigFailClosedTest test`. Baseline-verified: reverting just the `Server.java` change locally makes the first two tests fail (no exception thrown), confirming the test actually exercises this path.

Reported privately via the repository Security Advisory; opening this fail-closed hardening fix as a public PR per the maintainer. cc @andsel

### DIFF
--- a/broker/src/main/java/io/moquette/broker/Server.java
+++ b/broker/src/main/java/io/moquette/broker/Server.java
@@ -498,11 +498,8 @@ public class Server {
         if (authorizatorPolicy == null && !authorizatorClassName.isEmpty()) {
             authorizatorPolicy = loadClass(authorizatorClassName, IAuthorizatorPolicy.class, IConfig.class, props);
             if (authorizatorPolicy == null) {
-                // The operator explicitly configured a custom authorizator and it failed to load (typo,
-                // missing jar, throwing constructor, ...). Fail startup instead of silently falling through
-                // to the permissive default below: that default is only correct when NO class was configured.
                 throw new IllegalArgumentException("Unable to load configured " + IConfig.AUTHORIZATOR_CLASS_NAME
-                    + " '" + authorizatorClassName + "'; refusing to start with a silently permissive fallback. "
+                    + " '" + authorizatorClassName + "'; rejecting to start. "
                     + "Check the class name and classpath (see the WARN log above for the cause).");
             }
         }
@@ -533,10 +530,8 @@ public class Server {
         if (authenticator == null && !authenticatorClassName.isEmpty()) {
             authenticator = loadClass(authenticatorClassName, IAuthenticator.class, IConfig.class, props);
             if (authenticator == null) {
-                // Same rationale as initializeAuthorizatorPolicy: a configured-but-failed-to-load
-                // authenticator must not silently degrade to AcceptAllAuthenticator below.
                 throw new IllegalArgumentException("Unable to load configured " + IConfig.AUTHENTICATOR_CLASS_NAME
-                    + " '" + authenticatorClassName + "'; refusing to start with a silently permissive fallback. "
+                    + " '" + authenticatorClassName + "'; rejecting to start. "
                     + "Check the class name and classpath (see the WARN log above for the cause).");
             }
         }
--- a/broker/src/main/java/io/moquette/broker/Server.java
+++ b/broker/src/main/java/io/moquette/broker/Server.java
@@ -497,6 +497,14 @@ public class Server {
         String authorizatorClassName = props.getProperty(IConfig.AUTHORIZATOR_CLASS_NAME, "");
         if (authorizatorPolicy == null && !authorizatorClassName.isEmpty()) {
             authorizatorPolicy = loadClass(authorizatorClassName, IAuthorizatorPolicy.class, IConfig.class, props);
+            if (authorizatorPolicy == null) {
+                // The operator explicitly configured a custom authorizator and it failed to load (typo,
+                // missing jar, throwing constructor, ...). Fail startup instead of silently falling through
+                // to the permissive default below: that default is only correct when NO class was configured.
+                throw new IllegalArgumentException("Unable to load configured " + IConfig.AUTHORIZATOR_CLASS_NAME
+                    + " '" + authorizatorClassName + "'; refusing to start with a silently permissive fallback. "
+                    + "Check the class name and classpath (see the WARN log above for the cause).");
+            }
         }
 
         if (authorizatorPolicy == null) {
@@ -524,6 +532,13 @@ public class Server {
 
         if (authenticator == null && !authenticatorClassName.isEmpty()) {
             authenticator = loadClass(authenticatorClassName, IAuthenticator.class, IConfig.class, props);
+            if (authenticator == null) {
+                // Same rationale as initializeAuthorizatorPolicy: a configured-but-failed-to-load
+                // authenticator must not silently degrade to AcceptAllAuthenticator below.
+                throw new IllegalArgumentException("Unable to load configured " + IConfig.AUTHENTICATOR_CLASS_NAME
+                    + " '" + authenticatorClassName + "'; refusing to start with a silently permissive fallback. "
+                    + "Check the class name and classpath (see the WARN log above for the cause).");
+            }
         }
 
         IResourceLoader resourceLoader = props.getResourceLoader();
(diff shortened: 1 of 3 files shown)


broker/src/main/java/io/moquette/broker/Server.java
```diff
@@ -497,6 +497,14 @@ private IAuthorizatorPolicy initializeAuthorizatorPolicy(IAuthorizatorPolicy aut
         String authorizatorClassName = props.getProperty(IConfig.AUTHORIZATOR_CLASS_NAME, "");
         if (authorizatorPolicy == null && !authorizatorClassName.isEmpty()) {
             authorizatorPolicy = loadClass(authorizatorClassName, IAuthorizatorPolicy.class, IConfig.class, props);
+            if (authorizatorPolicy == null) {
+                // The operator explicitly configured a custom authorizator and it failed to load (typo,
+                // missing jar, throwing constructor, ...). Fail startup instead of silently falling through
+                // to the permissive default below: that default is only correct when NO class was configured.
```
```suggestion
            if (authorizatorPolicy == null) {
```

No need for such comment, the exception message already reveal the intention.

broker/src/main/java/io/moquette/broker/Server.java
```diff
@@ -497,6 +497,14 @@ private IAuthorizatorPolicy initializeAuthorizatorPolicy(IAuthorizatorPolicy aut
...
+                throw new IllegalArgumentException("Unable to load configured " + IConfig.AUTHORIZATOR_CLASS_NAME
+                    + " '" + authorizatorClassName + "'; refusing to start with a silently permissive fallback. "
```
```suggestion
                    + " '" + authorizatorClassName + "'; rejecting to start. "
```

broker/src/main/java/io/moquette/broker/Server.java
```diff
@@ -524,6 +532,13 @@ private IAuthenticator initializeAuthenticator(IAuthenticator authenticator, ICo
 
         if (authenticator == null && !authenticatorClassName.isEmpty()) {
             authenticator = loadClass(authenticatorClassName, IAuthenticator.class, IConfig.class, props);
+            if (authenticator == null) {
+                // Same rationale as initializeAuthorizatorPolicy: a configured-but-failed-to-load
+                // authenticator must not silently degrade to AcceptAllAuthenticator below.
```
```suggestion
```

broker/src/main/java/io/moquette/broker/Server.java
```diff
@@ -524,6 +532,13 @@ private IAuthenticator initializeAuthenticator(IAuthenticator authenticator, ICo
...
+                throw new IllegalArgumentException("Unable to load configured " + IConfig.AUTHENTICATOR_CLASS_NAME
+                    + " '" + authenticatorClassName + "'; refusing to start with a silently permissive fallback. "
```
```suggestion
                    + " '" + authenticatorClassName + "'; rejecting to start."
```

broker/src/test/java/io/moquette/broker/ServerAuthConfigFailClosedTest.java
```diff
@@ -0,0 +1,87 @@
+/*
+ * Copyright (c) 2012-2018 The original author or authors
```
andsel:
```suggestion
 * Copyright (c) 2012-2026 The original author or authors
```

broker/src/test/java/io/moquette/broker/ServerAuthConfigFailClosedTest.java
```diff
@@ -0,0 +1,87 @@
...
+import java.lang.reflect.Method;
+import java.util.Properties;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+/**
+ * A configured-but-failing-to-load authorizator/authenticator class must abort startup, not
+ * silently fall through to the permissive PermitAll/AcceptAll default reserved for the
+ * nothing-configured case. See MOQ-2026-0036.
```
andsel: What's this `See MOQ-2026-0036.` ?

broker/src/test/java/io/moquette/broker/ServerAuthConfigFailClosedTest.java
```diff
@@ -0,0 +1,87 @@
...
+
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.util.Properties;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+/**
+ * A configured-but-failing-to-load authorizator/authenticator class must abort startup, not
```
andsel:
```suggestion
 * A configured-but-failing-to-load authorizator/authenticator class must reject startup, not
```

broker/src/test/java/io/moquette/broker/ServerAuthConfigFailClosedTest.java
```diff
@@ -0,0 +1,87 @@
...
+ * silently fall through to the permissive PermitAll/AcceptAll default reserved for the
+ * nothing-configured case. See MOQ-2026-0036.
+ */
+public class ServerAuthConfigFailClosedTest {
+
+    private static final String MISSING_CLASS_NAME = "io.moquette.broker.NoSuchAuthorizatorPolicy";
+
+    private Object invokePrivate(String methodName, Class<?> paramType, Object first, IConfig config)
```
andsel: I don't like to use introspection tricks to invoke private methods. I think it's better to be explicit.

If a ba class name is defined in `IConfig.AUTHORIZATOR_CLASS_NAME` property, when `Server.startServer` throws that exception, so better to take inspiration from integration tests like `ServerIntegrationOpenSSLTest`.

andsel (changes requested):
Hi @gigioneggiando thank's for this PR.
I've left a doubt on the test, I would prefer to avoid introspection.

broker/src/test/java/io/moquette/broker/ServerAuthConfigFailClosedTest.java
gigioneggiando: Just an internal tracking reference of mine (this project doesn't use that ID scheme) - not meaningful outside my own notes, so I removed it in the rewrite. Thanks for catching it!

gigioneggiando:
Hi @andsel — heads up, this PR shows as conflicting now, but it's trivial: just one line in `ChangeLog.txt` clashing with the entries from #963/#964 (both since merged). `Server.java` and the new test file merge clean on their own.

Happy to rebase and push whenever's convenient, or I can leave it as-is until you're back to re-review — whichever you prefer.

andsel (dismissed):
Hi @gigioneggiando thanks for updating the PR!
LGTM, you can rebase and fix the conflicts, so that we can merge.

gigioneggiando:
Re-ran the failing job twice — different, unrelated test failed each time (`FlowControlTest` first run, `MetricsProviderPrometheusTest` second run), neither touching the files this PR changes. Looks like pre-existing flakiness in the suite, not a regression from this change. Rebased and conflict-free, ready whenever.

andsel (approved):
LGTM

PR #948, opened 2026-06-18
Added method for message interceptors to get the original message

hylkevds:
- enhancement

## What does this PR do?

MessageInterceptors can be used by applications that embed Moquette to receive the messages being sent through Moquette. The `InterceptPublishMessage` class is used to pass publishes to the intercepting application. It contains three methods for getting the Topic, the Payload and the ClientID associated with the message. The class does contain the original message, but does not offer a way to access the message.

MQTT 5 adds much more data to a message, such as a ContentType, user properties and correlation data. Interceptors also need to be able to access those fields. To give the interceptor access to these fields, we can either add methods for all those fields to the `InterceptPublishMessage` class, or give access to the contained message directly. The PR implements the direct access option.

- [X] My code follows the style guidelines of this project
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] ~~I have made corresponding changes to the documentation~~
- [X] ~~I have made corresponding change to the default configuration files (and/or docker env variables)~~
- [X] ~~I have added tests that prove my fix is effective or that my feature works~~
- [X] I have updated the Changelog if it's a feature or a fix that has to be reported

PR #970, opened 2026-08-18
Expose MQTT properties on InterceptPublishMessage

### LINKED ISSUE #918: Feature Request: Ability to query user properties when intercepting a PUBLISH message
Can the user properties be queried when intercepting messages? It looks like InterceptPublishMessage doesn't have the implementation yet to do that.

Thanks for looking!

torfinnberset:
## Release notes

InterceptPublishMessage exposes the MQTT properties of the intercepted PUBLISH message.

## What does this PR do?

Adds `getProperties()` to `InterceptPublishMessage`, returning the `MqttProperties` of the
intercepted PUBLISH. The properties survive the `retainedDuplicate()` that
`BrokerInterceptor.notifyTopicPublished` makes for each handler, because the duplicate shares the
variable header.

An intercept handler can read only the topic, payload, client id and username. Everything else the
publisher attached to an MQTT 5 PUBLISH is lost on the way to the handler: user properties, content
type, response topic, correlation data, payload format indicator and message expiry interval. The
broker forwards all of them to subscribers, so the interception API is the only place where this
metadata disappears.

With the accessor an embedder can:

- read the content type the publisher declared and parse the payload accordingly, instead of
  guessing or agreeing on a format out of band
- read the user properties the publisher set, and use them for routing, filtering, tracing or
  tenant tagging without adding its own header inside the payload
- take part in request and response flows, because the response topic and the correlation data are
  now visible to the handler
- honour the message expiry interval when it stores or forwards the message

Today the only way to reach these values is to stop using the interception API and add a Netty
handler or a custom pipeline, which means giving up the hook the broker offers for exactly this
purpose.

## Checklist

- [ ] ~~I have made corresponding changes to the documentation~~
- [ ] ~~I have made corresponding change to the default configuration files (and/or docker env variables)~~

    ./mvnw -pl broker test -Dtest=InterceptPublishMessageTest

The two tests read the properties back from a PUBLISH built with them, once directly and once
through the retained duplicate that `BrokerInterceptor` makes for each handler.

## Related issues

- Closes #918

## Use cases

- An embedder bridges MQTT traffic to another system, for example Kafka or an HTTP endpoint, and
  carries the publisher user properties over as message headers instead of dropping them.
- An audit or archive handler stores the content type and the correlation data next to the payload,
  so a later reader can decode the payload and match a reply to its request.
- A handler that observes a request and response flow reads the response topic to know where the
  answer is expected, without parsing the payload.
- A handler that reads no properties keeps working unchanged, because the accessor only adds a
  method.

### DIFF
--- a/ChangeLog.txt
+++ b/ChangeLog.txt
@@ -1,4 +1,5 @@
 Version 0.19-SNAPSHOT
+   [feature] InterceptPublishMessage exposes the MQTT properties of the intercepted PUBLISH message. (#918)
    [feature] Pass remote address of a client when notify to interceptor handlers, adding it into InterceptConnectMessage. (#946)
    [fix] A second connect message sent on an already connected client must close the connection as a protocol error. (#969)
    [fix] Add a length max to the session queues, this avoids that a slow ACK-ing subscriber consumes too much resources if a fast publisher is sending to same topic.
--- a/broker/src/main/java/io/moquette/interception/messages/InterceptPublishMessage.java
+++ b/broker/src/main/java/io/moquette/interception/messages/InterceptPublishMessage.java
@@ -17,6 +17,7 @@
 package io.moquette.interception.messages;
 
 import io.netty.buffer.ByteBuf;
+import io.netty.handler.codec.mqtt.MqttProperties;
 import io.netty.handler.codec.mqtt.MqttPublishMessage;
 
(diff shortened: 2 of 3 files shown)


torfinnberset:
Closing, #948 now exposes the original message, so the properties are reachable through getMessage().variableHeader().properties(). #918 can be closed with it.

